Issue #1, opened 2020-09-01
No targets error while make project after make


SuhoSon:
I clone xNVMe repository and since I did not install meson, make is failing.
After install meson, I do re-make the project, but error is occurred like below.
I think it caused by failing to make third-party, but I'm not sure.

``` bash
CMake Error at CMakeLists.txt:255 (message):
  Failed building SPDK/DPDK


-- Configuring incomplete, errors occurred!
See also "/home/oslab4/xNVMe/build/CMakeFiles/CMakeOutput.log".
See also "/home/oslab4/xNVMe/build/CMakeFiles/CMakeError.log".


Configuration failed


Makefile:53: recipe for target 'config' failed
make[1]: *** [config] Error 1
make[1]: Leaving directory '/home/oslab4/xNVMe'
Makefile:47: recipe for target 'default' failed
make: *** [default] Error 2
root@oslab4 ~/xNVMe (master)$
root@oslab4 ~/xNVMe (master)$ make                                                  <-- my command
## xNVMe: make info
OSTYPE:
PLATFORM: Linux
CC: cc
CXX: g++
MAKE: make
CTAGS: ctags
NPROC: 32
## xNVMe: make tags
/bin/sh: 1: ctags: not found
## xNVMe: make default
$( case $( uname -s ) in ( Linux ) echo "make" ;; ( FreeBSD | OpenBSD | NetBSD ) echo "gmake" ;; ( * ) echo Unrecognized ;; esac) build
make[1]: Entering directory '/home/oslab4/xNVMe'
## xNVMe: make info
OSTYPE:
PLATFORM: Linux
CC: cc
CXX: g++
MAKE: make
CTAGS: ctags
NPROC: 32
## xNVMe: make build
cd build && $( case $( uname -s ) in ( Linux ) echo "make" ;; ( FreeBSD | OpenBSD | NetBSD ) echo "gmake" ;; ( * ) echo Unrecognized ;; esac)
make[2]: Entering directory '/home/oslab4/xNVMe/build'
make[2]: *** No targets specified and no makefile found.  Stop.
make[2]: Leaving directory '/home/oslab4/xNVMe/build'
Makefile:74: recipe for target 'build' failed
make[1]: *** [build] Error 2
make[1]: Leaving directory '/home/oslab4/xNVMe'
Makefile:47: recipe for target 'default' failed
make: *** [default] Error 2
```

safl:
Hi @SuhoSon,

[...]

PR #2, opened 2020-09-01
Change Makefile and CMakeLists.txt on third-party

SuhoSon:
Related to #1 .

When doing make, if it was terminated during the previous configure, there was a problem with a rebuild and I need to do "make clean" and then "make" again.
In this commit, I checked that the CMakeLists.txt file exists under the build directory so that I can do "make" it again without "make clean".

Additionally, "make clean" did not clean the third-party projects, so I added a command line for cleaning third-party projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ clean:
 	@echo "## xNVMe: make clean"
 	rm -fr $(BUILD_DIR) || true
 	rm -fr cmake-build-debug || true
-	$(MAKE) third-party-clean
+	$(MAKE) third-party-clobber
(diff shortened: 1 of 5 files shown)


safl:
The CI system is there to help make testing easier, feel free to use it by creating a pull-request.

Unfortunately when you created your your PRs earlier, then the pip-installed meson needed for building the third-party dependency had gone incompatible with the Python version shipping with Ubuntu so the build was failing. If you had a fix related to #1, then please reopen or create a new PR.

Cheers,
Simon

PR #9, opened 2021-02-19
Add xnvme_nvm_write_zeroes command with a test

halfzebra:
This PR contains the implementation of `xnvme_nvm_write_zeroes` and a test for it.
The test is not implemented to reuse the helpers, but rather to learn the process in-depth as Michael has advised.

Please [:point_right: skip to the actual test :point_left:](https://github.com/OpenMPDK/xNVMe/pull/9/files#r579041100) to not have to scroll through changes I didn't make.


tests/lblk.c
```diff
@@ -526,6 +443,129 @@ test_write_uncorrectable(struct xnvmec *cli)
 	return err;
 }
```
halfzebra, line 444: This is the only file that contains changes :point_down:

PR #12, opened 2021-08-03
GU Project: add support for xnvme repo in lb build system

ofir-lb:
* Makefile.lb, lb.yaml and required cmake file for integrating xnvme into duroslight build


ofir-lb:
@an

PR #16, opened 2021-10-28
xnvme_fioe_reset_wp() was resetting one too many zone

pierrelabat:
Signed-off-by: Pierre Labat <plabat@micron.com>

The problem is exposed when running fio random write on a zns device via xnvme.
Fio, after a while, fills up a number of zones. Then it starts to reset some zones, one at a time.
When fio reset a zone xnvme_fioe_reset_wp() is called. 
xnvme_fioe_reset_wp() resets 2 zones instead of only one as it should. The next one is also reset.
Later, when fio tries to write in the zone reset by mistake (and without fio being aware of it), the zns device returns a write error because fio thinks the WP is A  (with A > zone start) while on the device it the WP is now zone start. The FIO tests fails.

I tested the fix, and with it, the fio test completes without error.

### DIFF
--- a/tools/fio-engine/xnvme_fioe.c
+++ b/tools/fio-engine/xnvme_fioe.c
@@ -968,7 +968,7 @@ xnvme_fioe_reset_wp(struct thread_data *td, struct fio_file *f, uint64_t offset,
 	last = (((offset + length) >> ssw) / geo->nsect) * geo->nsect;
 	XNVME_DEBUG("INFO: first: 0x%lx, last: 0x%lx", first, last);
 
-	for (uint64_t zslba = first; zslba <= last; zslba += geo->nsect) {
+	for (uint64_t zslba = first; zslba < last; zslba += geo->nsect) {
 		struct xnvme_cmd_ctx ctx = xnvme_cmd_ctx_from_dev(dev);
 
 		if (zslba >= (geo->nsect * geo->nzone)) {


pierrelabat:
To hit the problem I used:
fio    --filename=/dev/nvme0n1 \
       --ioengine=libxnvme-fio-engine.so --xnvme_async=libaio \
       --direct=1 --zonemode=zbd --offset=0 --rw=randwrite \
       --bs=32k \
       --ignore_zone_limits=1 --max_open_zones=4096 \
       --runtime=1800 --thread=1 --numjobs=1 --iodepth=1 --group_reporting \
       --name=pierre_test

Note that I updated the xnvme fio from 3.27 to 3.28 to have the option --ignore_zone_limits and be able to have a large number of active zones (--max_open_zones=4096).

I think you can also hit the problem with 3.27 and less active zones if the test runs long enough to have fio reset zones.

micvbang:
Thanks for this @pierrelabat.
Just to confirm, which version of xNVMe are you running? v0.0.25?

pierrelabat:
> Thanks for this @pierrelabat. Just to confirm, which version of xNVMe are you running? v0.0.25?

Yes.

micvbang:
Sorry for the turnaround time :)

Good news is that I was able to recreate the problem and verify that your fix mitigates it. Awesome!
I'll merge once the tests have completed.

Thank you @pierrelabat!

javigon:
Merging this on behalf of Michael

PR #22, opened 2022-02-21
Fix xnvme_3p.py and update xnvme_3p_ver.c

### LINKED ISSUE #20: 3p-information is stale
The third-party information, e.g. which version of fio, SPDK, and liburing which xNVMe is built against has gone stale.
The reason being that the `xnvme_3p.py` is broken, this came with the to meson where the git-submodules were also removed.
Thus, the script ``xnvme_3p.py`` needs fixing, to again produce relevant information about the 'meson-subprojects', as well as OS version that xNVMe is built against.

karlowich:
Closes #20 


karlowich:
This is made obsolete by #24

PR #23, opened 2022-02-22
Make 3p function use the array it takes

karlowich:


karlowich:
This is made obsolete by #24

PR #26, opened 2022-03-03
Updated git pre-commit hook to only apply to staged changes

Baekalfen:

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,6 +1,22 @@
 #!/usr/bin/env bash
 # Call the style-check script
 exec 1>&2
+
+function cleanup {
+  echo "Git: Unstashing files"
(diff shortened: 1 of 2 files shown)


safl:
This looks interesting, however, it seems like it breaks when running ``./style.sh`` manually, since it has to be run inside ``scripts/``, which does not match with ``git diff`` outputting e.g. ``./lib/xnvme_be.c``.

Also, this is a bit nit-picky, but I do prefer e.g. "Update git pre-commit..." instead of "Updated git pre-commit...", sort of describing what that commit will change, rather than what the author has done.

Baekalfen:
Closing this in favor of making a `pre-commit` branch instead.

Baekalfen:
Closing this in favor of making a `pre-commit` branch instead.

PR #27, opened 2022-03-03
Added support for building on Darwin/macOS

Baekalfen:


.gitignore
```diff
@@ -32,3 +29,9 @@ out/
 *.exe
 *.lib
 *.vcxproj.*
+subprojects/*
```
safl, line 32: Can you provide a brief description, in the commit-message, of why this is needed?

Makefile
```diff
@@ -16,47 +16,47 @@ PLATFORM_ID = $$( uname -s )
...
 	case $(PLATFORM_ID) in \
-		( Linux | Windows) echo "git" ;; \
+		( Linux | Windows | Darwin ) echo "git" ;; \
 		( FreeBSD | OpenBSD | NetBSD ) echo "git" ;; \
 		( * ) echo Unrecognized ;; \
 	esac)
 
 PROJECT_VER = $(( python3 scripts/xnvme_ver.py --path meson.build ))
-BUILD_DIR?=builddir
+BUILD_DIR?=builddir-$(shell uname | tr '[:upper:]' '[:lower:]')-$(shell arch)
```
safl, line 59: A brief description in the commit-message describing this would be great.
This would create a builddir named ``builddir-linux-x86_64`` on Linux and something fitting on MacOS.
Why is that required?

lib/xnvme_buf.c
```diff
@@ -18,6 +18,8 @@ xnvme_buf_virt_alloc(size_t alignment, size_t nbytes)
 #ifdef WIN32
 	return _aligned_malloc(nbytes, alignment);
 #else
+	// nbytes has to be a multiple of alignment. Therefore, we round up to the nearest multiple.
```
safl, line 21: This would be great as commit in itself with a commit-message like:
```
buf: defensively meet alignment constraints in xnvme_buf_virt_alloc()

The virtual memory allocator uses aligned_alloc()/_aligned_mmalloc().
These require that nbytes is a positive multiple of alignment, this change ensures that.
```

And then move the re-computation of ``nbytes`` up in between 17-18, such that it also applies to Windows.

safl: Then for the meson-changes, something like:
```
build:meson: enable build of be:posix on Darwin/macOS
```

safl (changes requested):
Exciting! This is looking pretty good.
Added some comments, pleaee have a look.
Thanks,
Simon

Makefile
safl: Or revert simply use the "builddir" name...

Baekalfen: I'll withdraw it for now, and maybe do a separate PR with it, because the functionality is not complete (subprojects are not acting nice). But it allows you to compile for multiple platforms/archs in the same repo.

.gitignore
Baekalfen: Done

safl (changes requested):
This looks good, one tiny thing, looks like an unintentional line-break on: 477774e422284fd04817dcc9e2021509e420d577

Perhaps use something like:
```
git:ignore: exclude subprojects

The current git-ignore specifies specific subprojects to ignore. That behavior
in insufficient as subprojects can fetch other subprojects. Thus, this change
ignores everything in subprojects, except for those things we actually care
about; the ``packagefiles`` and ``.wrap``.
```

Then the reader also get the story behind the change.

Makefile
safl: I see that would be useful, but yeah, let's do that in another set of changes which has more of that functional completeness :)

Baekalfen:
> 

The line-break is automatic. It's part of the "standard" for Git to break the line of the header at 80 columns.

Maybe we should consider whether the this information should be in the code/files, or in Git. But I've added the text you wrote.

Makefile
```diff
@@ -16,41 +16,41 @@ PLATFORM_ID = $$( uname -s )
...
-		( Linux ) echo "ctags" ;; \
+		( Linux | Darwin ) echo "ctags" ;; \
 		( FreeBSD | OpenBSD | NetBSD | Windows) echo "exctags" ;; \
 		( * ) echo Unrecognized ;; \
 	esac)
 
 MAKE = $$( \
 	case $(PLATFORM_ID) in \
-		( Linux | Windows) echo "make" ;; \
+		( Linux | Windows | Darwin ) echo "make" ;; \
```
safl, line 32: Should this be added to the next line instead to invoke ``gmake``?

Makefile
```diff
@@ -16,41 +16,41 @@ PLATFORM_ID = $$( uname -s )
...
-		( Linux | FreeBSD | OpenBSD | NetBSD | Windows) echo $(PLATFORM_ID) ;; \
+		( Linux | FreeBSD | OpenBSD | NetBSD | Windows | Darwin ) echo $(PLATFORM_ID) ;; \
 		( * ) echo Unrecognized ;; \
 	esac)
 
 CTAGS = $$( \
 	case $(PLATFORM_ID) in \
-		( Linux ) echo "ctags" ;; \
+		( Linux | Darwin ) echo "ctags" ;; \
 		( FreeBSD | OpenBSD | NetBSD | Windows) echo "exctags" ;; \
```
safl, line 26: Are ``ctags`` available are they called ``exctags`` on MacOS like they are on the BSDs?

safl (changes requested):
Looks good, a couple of questions inline.
Other than that there just seems to be an source-format that needs to be applied.

Makefile
Baekalfen: It's correct as it is. `brew install ctags` installs Exuberant Ctags as `ctags`

Makefile
Baekalfen: It's just `make`. I don't remember if I installed it through `brew`, but there's no `gmake` on my system.

PR #29, opened 2022-03-04
mk: change format of 'make help'

safl:
@karlowich your addition of 'make help' is great, after having used it a bit, I felt like it would be nice if it was slightly more compact. Do you have any thoughts on format changes provided here?
@micvbang you poked about this on Discord, do you have any thoughts on this? :)

Cheers,
Simon


karlowich:
This looks very nice!
I think the verbose version could use a little more love.
Something to highlight the commands would be nice, maybe they could be **bold**?
That way it would be easier to see the difference between the commands and the verbose descriptions.

image

safl:
Good idea, it looks better and more easily readable. Respun the patch.

micvbang:
This looks great!
I completely agree with @karlowich that making the commands/targets bold is great.
The only thing I can think of to add (which I guess is super obvious anyway) is a 
```
Usage:
  make [command (or target?)]
```

micvbang:
Hmm, trying `verbose` out now, it seems there are too many newlines before the verbose descriptions. This is what I'm seeing:

```
[1mbuild               [0m : Build xNVMe with Meson
[1mbuild-deb           [0m : Build a binary DEB package


 NOTE: The binary DEB packages generated here are not meant to be used for anything
 but easy install/uninstall during test and development

[1mclean               [0m : Remove Meson builddir
[1mclean-subprojects   [0m : Remove Meson builddir as well as running 'make clean' in all subprojects 
[1mclobber             [0m : Clean third-party builds, drop any third-party changes and clean any untracked stuff lying around
[1mconfig              [0m : Configure Meson
[1mconfig-debug        [0m : Configure Meson to compile xNVMe with '--buildtype=debug'
[1mconfig-slim         [0m : Configure Meson to compile xNVMe without backends
[1mconfig-uring        [0m : Configure Meson to compile xNVMe with liburing as the only backend
[1mdefault             [0m : Run info, tags, git-setup and configure targets

 This is called when running 'make'

[1mgen-bash-completions[0m : Helper-target to produce Bash-completions for tools (tools, examples, tests)


 NOTE: This target requires a bunch of things: binaries must be built and
 residing in 'builddir/tools' etc. AND installed on the system. Also, the find
 command has only been tried with GNU find

[1mgen-libconf         [0m : Helper-target generating third-party/subproject/os/library-configuration string
[1mgen-man-pages       [0m : Helper-target to produce man pages for tools (tools, examples, tests)

 
 NOTE: This target requires a bunch of things: binaries must be built and
 residing in 'builddir/tools' etc. AND installed on the system. Also, the find
 command has only been tried with GNU find

[1mgen-src-archive     [0m : Helper-target to produce full-source archive
[1mgit-setup           [0m : Setup .githooks
[1mhelp                [0m : Print the description of every target
[1mhelp-verbose        [0m : Print the verbose description of every target
[1minfo                [0m : Print information relevant to xNVMe

 Such as:
 - OS
 - Project version
 - Versions of tools

[1minstall             [0m : Install xNVMe with Meson
[1minstall-deb         [0m : Install the binary DEB package created with: make build-deb

 This will install it instead of copying it directly
 into the system paths. This is convenient as it is easier to purge it by
 running eg.
 	make uninstall-deb

[1mtags                [0m : Helper-target to produce tags
[1mtags-xnvme-public   [0m : Make tags for public APIs
[1muninstall           [0m : Uninstall xNVMe with Meson
[1muninstall-deb       [0m : Uninstall xNVMe with apt-get
```

micvbang:
Other than the above, I think this is a great addition to xNVMe!

safl:
Ahh, thanks for the input, fixed!
It now looks like the following.

**brief**
brief
**verbose**
verbose

safl:
Final rendition:
done

micvbang:
Looks great!

PR #33, opened 2022-03-05
mk,ci,scripts: switch code-formater from astyle to clang-format

safl:
The occasional nondeterministic behavior of astyle motivates a switch of tooling for code-formatting.
Also, currently, a lot of in-consistent code-format exists as astyle does not catch all.
This PR attempts to fix that by switching from ``astyle`` to clang-format.
Also, the git pre-commit gets in the way when rebasing, this PR also provide a way to disable hooks when rebasing
by setting the environment variable ``GIT_HOOKS_OFF=1``.
The PR is marked as DRAFT since, there is still some tasks to do related to this, the biggest being, actually applying the new style as it affects all source and header files. Hopefully, after this switch we won't need to do a major rework like this again.

TODO:

- [x] Fix formatting on sub-commands
- [x] Apply the new style-formatting


safl:
Another note: the source-format.sh should probably exit with a better message in case ``clang-format-14`` is not available.

PR #42, opened 2022-03-14
macOS docs and CI build

Baekalfen:
Add package and documentation for building xnvme on macOS. Additionally,
adding macOS as part of GitHub Actions build scripts on pull-requests.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -330,6 +330,51 @@ jobs:
       run: xnvme library-info
 
   #
+  # Build on macOS
+  #
+  build-macos:
+    needs: source-archive
+    runs-on: ${{ matrix.runner.os }}-${{ matrix.runner.ver }}
+    if: (! contains(github.ref, 'ci')) || contains(github.ref, 'ci-build-mac')
+
+    strategy:
+      matrix:
+        runner:
+          - { os: 'macos', ver: '11' }
+
+    steps:
+    - name: Retrieve the source-archive
+      uses: actions/download-artifact@v2
+      with:
+        name: archive-src
+    - name: Extract the full-source-archive
+      run: |
+        tar xzf xnvme*.tar.gz --strip 1
+        rm xnvme*.tar.gz
+
+    - name: Install build-requirements
+      run: |
+        source scripts/pkgs/${{ matrix.runner.os }}-${{ matrix.runner.ver }}.sh || true
+
+    - name: Configure, Build, and Install
+      env:
+        BSCRIPT_DEF: scripts/pkgs/default-build.sh
+        BSCRIPT: scripts/pkgs/${{ matrix.runner.os }}-${{ matrix.runner.ver }}-build.sh
+      run: |
+        if [[ -f "${BSCRIPT}" ]]; then source ${BSCRIPT}; else source ${BSCRIPT_DEF}; fi
+
+    - name: dump meson-log on failure
+      if: failure()
+      run: |
+        cat builddir/meson-logs/meson-log.txt
+
+    - name: Execute 'xnvme enum'
+      run: xnvme enum
+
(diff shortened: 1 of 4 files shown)


.github/workflows/verify.yml
```diff
@@ -329,6 +329,55 @@ jobs:
...
+    runs-on: ${{ matrix.runner.os }}-${{ matrix.runner.ver }}
+    if: (! contains(github.ref, 'ci')) || contains(github.ref, 'ci-build-mac')
+
+    strategy:
+      matrix:
+        runner:
+          - { os: 'macos', ver: '11' }
+
+    steps:
+    - name: Container-preparation, openSUSE does not have tar and gzip...
```
safl, line 346: You can remove this step

scripts/pkgs/macos-11.txt
```diff
@@ -0,0 +1,4 @@
+astyle
```
safl, line 1: Surprisingly few deps, once SPDK is not taken into consideration.
What about ``clang``? I guess that comes with the user-land? No need to specify installation of it?
I guess ``astyle`` can be removed at this point?

safl (changes requested):
Looks great!
For the git-short-message, since the commit affects multiple "areas", then prefix it like:
```ci,scripts,docs: add build of macOS``

File: .github/workflows/verify.yml (hunk 1 above)
Comment (Baekalfen) [reply .github/workflows/verify.yml]: Done

File: scripts/pkgs/macos-11.txt (hunk 2 above)
Comment (Baekalfen) [reply scripts/pkgs/macos-11.txt]: A working installation of `brew` assumes Xcode and its command line tools are installed. This is what supplies `clang`, and likely a lot of other smaller dependencies.

I've removed `astyle`.

**Review by safl — CHANGES_REQUESTED** [review]
Syntax error in ``verify.yml`` otherwise: LGTM.

File: .github/workflows/verify.yml [hunk 3]
```diff
@@ -302,6 +302,7 @@ jobs:
       uses: actions/download-artifact@v2
       with:
         name: archive-src
+
```
safl, line 305: Please don't make unrelated stylistic changes.

The reason why there is no empty line is that those two steps go "hand-in-hand" download + extract the source.
One can argue whether or not an empty line should be here, however, that would be a separate commit changing that.
If you feel strongly about the it, then the best way forward would be the addition of some tool forcing certain formatting for the ``.yaml`` or at least a separate commit that one can add to git-blame-ignore.

.github/workflows/verify.yml
```diff
@@ -570,6 +618,7 @@ jobs:
...
```
safl, line 621: Please don't make unrelated stylistic changes.

The reason why there is no empty line is that those two steps go "hand-in-hand" download + extract the source.
One can argue whether or not an empty line should be here, however, that would be a separate commit changing that.
If you feel strongly about the it, then the best way forward would be the addition of some tool forcing certain formatting for the ``.yaml`` or at least a separate commit that one can add to git-blame-ignore.

.github/workflows/verify.yml
```diff
@@ -348,6 +395,7 @@ jobs:
...
```
safl, line 398: Please don't make unrelated stylistic changes.

The reason why there is no empty line is that those two steps go "hand-in-hand" download + extract the source.
One can argue whether or not an empty line should be here, however, that would be a separate commit changing that.
If you feel strongly about the it, then the best way forward would be the addition of some tool forcing certain formatting for the ``.yaml`` or at least a separate commit that one can add to git-blame-ignore.

safl (changes requested):
Added some nit-picking comments :)

safl (approved):
Thanks @Baekalfen ! This is looking great.

PR #43, opened 2022-03-15
xNVMe io_uring passthrough command support with bigsqe entries

krish-reddy:
Signed-off-by: Krishna Kanth Reddy <krish.reddy@samsung.com>

### DIFF
--- a/include/libxnvme_spec.h
+++ b/include/libxnvme_spec.h
@@ -1193,6 +1193,47 @@ struct xnvme_spec_cmd_common {
 XNVME_STATIC_ASSERT(sizeof(struct xnvme_spec_cmd_common) == 64, "Incorrect size")
 
 /**
+ * NVMe Command Accessor for Passthru 80byte NVMe command
+ *
+ * @struct xnvme_spec_cmd_passthru
+ */
+struct xnvme_spec_cmd_passthru {
+	/* cdw 00 */
+	uint8_t opcode;			///< OPC: Command Opcode
+	uint8_t flags;
+	uint16_t rsvd;
+
+	/* cdw 01 */
+	uint32_t nsid;			///< NSID: Namespace identifier
+
+	uint32_t cdw02;
+	uint32_t cdw03;
+
+	/* cdw 04-05 */
+	uint64_t mptr;			///< MPTR -- metadata pointer
+
+	/* cdw 06-09: */		///< DPTR -- data pointer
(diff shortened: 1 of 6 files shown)


include/libxnvme_spec.h
```diff
@@ -1192,6 +1192,47 @@ struct xnvme_spec_cmd_common {
 };
 XNVME_STATIC_ASSERT(sizeof(struct xnvme_spec_cmd_common) == 64, "Incorrect size")
 
+/**
```
safl, line 1195: The ``libxnvme_spec.h`` is intended for spec-definitions only, this struct should go into the Linux-backend implementation: ``xnvme_be_linux.h``.

safl (changes requested):
Thanks for this!
From a quick look, then a change-request for placing the cmd/ioctl definition in the internal backend header instead of in the public API.
Will continue later today.

include/libxnvme_spec.h
krish-reddy: Okay Simon, sure. Will move this structure to xnvme_be_linux.h

safl: To fix the source-format caught by CI, then run: ``make source-format``.
For the build-error on Debian Buster, then you can temporarily disable that build by editing ``.github/workflows/verify.yml`` and commenting out line 280 (buster) and 281 (stretch).

safl (commented):
Upon closer inspection, since ``xnvme_be_linux_async_liburing.c`` is providing use of the upstream supported interface, then let's move all the ``bigsqe`` changes onto ``xnvme_be_linux_async_ucmd.c`` as it contains the non-stable/experimental interface.
To support flipping the ``big-sqe`` ring/queue-initialization and encapsulating the bigsqe/ioctl then those things should go into another header like ``xnvme_be_linux_async_ucmd.h``.
I'll take a stab at this based on your PR and the one from @ankit-sam
https://github.com/OpenMPDK/xNVMe/pull/49
I'll then put up a PR of that for you guys to review.

safl:
Hi @krish-reddy, have a look at: https://github.com/OpenMPDK/xNVMe/pull/51
It is missing the ``big-sqe`` portion of the intended changes as well as NVMe-ioctl-vec in ``ucmd``.
for some reason, GitHUB would not let me add you are reviewer, but I would appreciate if you can take a look.
Thanks,
Simon

PR #60, opened 2022-03-25
Easing into the 'pre-commit' framework (pcf)

safl:
This is a re-work of @Baekalfen initiative to introduce ``pre-commit`` at https://github.com/OpenMPDK/xNVMe/pull/48
This picks up the commits and does the following:
* Remove a handful of scripts and tools that are out-dated
* Do the clang-format using the existing clang-format definitions and adding a script using them
* Add helpers to ``Makefile`` invoking ``pre-commit`` manually
* Add ``pre-commit`` to ``.githooks/pre-commit`` with
* Add the different pre-commit hooks/plugins
* Apply them and manually fix-up lint issues in a bulk-commit
* Add the commit to git-blame-ignore
* Add a 'pre-commit' badge in README


scripts/pcf_clang_format.py
```diff
@@ -0,0 +1,83 @@
...
+    args.style_file = expand_path(args.style_file)
+
+    return args
+
+
+def clang_format(args):
+    """Invoke clang-format"""
+
+    # Form arguments to clang-format parameter "-style"
+    style = []
```
Baekalfen, line 51: Unnecessary definition.

scripts/pcf_clang_format.py
```diff
@@ -0,0 +1,83 @@
...
+    with open(args.style_file) as sfd:
```
Baekalfen, line 52: Unpythonic variable naming: `sfd` -> `style_file`, `f` or like that

Baekalfen (commented):
Tiny comments, will work regardless

Baekalfen (commented):
In commit message:

> With the option of running it on everything:
  pre-commit run clang-format --all

`--all` refers to all hooks, `--all-files` will make it run on unstaged files as well.

Makefile
```diff
@@ -122,6 +122,24 @@ git-setup:
...
+	@pre-commit run
+	@echo "## xNVME: format [DONE]"
+
+define format-all-help
+# run code format (style, code-conventions and language-integrity) on staged and committed changes
+endef
+.PHONY: format-all
+format-all:
+	@echo "## xNVMe: format-all"
+	@pre-commit run --all
```
Baekalfen, line 140: `--all` refers to all hooks, `--all-files` will make it run on unstaged files as well.

.githooks/pre-commit
```diff
@@ -1,25 +1,45 @@
...
 ==============================================================================
- Error: invalid source-format
+Error: the 'pre-commit' framework https://pre-commit.com/ is not available
 ==============================================================================
 
-* You can inspect the current format-issues here: 'scripts/source-format.log'
+You can install it via the Python Package Index:
 
-* Run the source-formatter / fix the source-format issues
+  pip3 install --user pre-commit
```
Baekalfen, line 16: Additional step for commit message check:
```
pre-commit install --hook-type commit-msg
```

.pre-commit-config.yaml
```diff
@@ -59,3 +59,69 @@ repos:
...
+    - '--extend-ignore=E203' # ignore "whitespace before ':'"
+- repo: https://github.com/jendrikseipp/vulture
+  rev: 'v2.3'
+  hooks:
+  - id: vulture
+    name: Python-lint-vulture
+    types: [python]
+    pass_filenames: true
+    # excluding sphinx-conf as it would break to change it
+    exclude: docs/autogen/conf.py$
```
Baekalfen, line 96: I think we need a bit on config to make it actually trigger. But it doesn't break anything to have it in.

.github/workflows/verify.yml
```diff
@@ -56,49 +56,24 @@ jobs:
...
-        apt -qy install wget gnupg2 lsb-release software-properties-common
-        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-        add-apt-repository -y "deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-14 main"
-        apt -qy update
-        apt -qy install clang-format-14
-        update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-14 100
-
-    - name: xNVMe, run the source-format script
+        apt-get -qy update && apt-get -qy install shellcheck
+        pip3 install virtualenv==20.0.33
```
Baekalfen, line 73: Shouldn't this be in `refenv/xnvme-devtools-ci:latest`?

.github/workflows/verify.yml
```diff
@@ -56,49 +56,24 @@ jobs:
...
-      if: failure()
-      run: |
-        cat scripts/source-format.log
-
-    - name: Upload source-format log
-      uses: actions/upload-artifact@v2
-      with:
-        name: source-format-issues
-        path: scripts/source-format.log
+        pre-commit run --all
```
Baekalfen, line 76: I think you'll need `--all-files` https://github.com/pre-commit/action/blob/main/action.yml

Baekalfen:
Maybe I'm wrong about the `--all` vs `--all-files` arg. I can't actually find `--all` in the docs.

.github/workflows/verify.yml
safl: Nice, catch, yes. Just added it here until it "worked". Fixing...

.github/workflows/verify.yml
safl: Nice catch. Curiously enough it "works" with ``--all`` but that must be a fluke-accident from how the parser works... fixing...

scripts/pcf_clang_format.py
safl: Yeah, I see your point. My C abbreviation habits leaking in here. Fixing it.

safl:
> 

I think you are right, however, for some reason ``--all`` behaves the same as ``--all-files``. Changing it to ``--all-files`` since that is what the ``--help`` says..

.githooks/pre-commit
```diff
@@ -1,25 +1,45 @@
...
+when using Bash:
+
+  export GIT_HOOKS_OFF=1
+
+==============================================================================
+EOF
+	exit 1
+}
+
+if ! pre-commit run; then
```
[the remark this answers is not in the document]
safl: Hmm, yeah... this overlaps a bit with the whether or not ``GIT_HOOKS_OFF=1`` is actually needed at all.
I just kind of like that the script tells the user what to do and how to easily work around it.
But dropping the custom-script with ``GIT_HOOKS_OFF`` and relying on ``pre-commit install`` would ensure a proper invocation of ``pre-commit``. Hmm...

.githooks/pre-commit
safl: Doesn't ``pre-commit`` auto-install these on first-run? Along with all the others?

Makefile
safl: As discussed elsewhere, it is curious... but it is not changed to ``--all-files``.

scripts/pcf_clang_format.py
safl: Fixed.

Makefile
safl: Fixed.

safl:
Thanks for the input @Baekalfen ! I have addressed the comments, except for one/two things:

- [ ] Question: Is the addition of ``pre-commit install --hook-type commit-msg`` required? Won't pre-commit take care of that along with along the other plugin/hook-installations?
- [ ] Whether or not ``pre-commit install`` should take of ``.githooks/pre-commit`` or whether the current approach is good enough. There is a clear argument that using ``pre-commit install`` would be preferred as it sets up ``pre-commit`` according the local environment, and that one should just use ``--no-verify`` when committing/rebasing in case one wants to ignore ``pre-commit``. Let me just mull on this one...

.pre-commit-config.yaml
```diff
@@ -0,0 +1,135 @@
...
+    types: [yaml]
+
+# Python: for all the helpers in scripts/*
+- repo: https://github.com/psf/black
+  rev: '22.1.0'
+  hooks:
+  - id: black
+    name: Python-format-black
+    args:
+    - '--line-length=79'
```
Baekalfen, line 70: Can't say I like the 79 character limit. That's going to be in the way

.pre-commit-config.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - '--mapping=2'
+    - '--sequence=2'
+    - '--offset=0'
+    - '--width=120'
+    - '--preserve-quotes'
+    types: [yaml]
+
+# Python: for all the helpers in scripts/*
+- repo: https://github.com/psf/black
+  rev: '22.1.0'
```
Baekalfen, line 65: Should be updated to `'22.3.0'`. This fixes a bug which is blocking CI right now

Baekalfen: https://github.com/psf/black/issues/2964

.githooks/pre-commit
Baekalfen: Could be it was only explicitly needed, as I was adding the commit-msg hook after the regular install.

But if we have our own pre-commit script, we need a hook in Git to trigger on commit-msg.

.pre-commit-config.yaml
Baekalfen: Passing filenames will presumably make `vulture` only check for dead-code within each file. You need to supply directories/packages to check inter-dependability.

.githooks/pre-commit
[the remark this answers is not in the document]
safl: Adjusted according to offline discussion.

.pre-commit-config.yaml
safl: Fixed

.pre-commit-config.yaml
safl: Bumped up to use the default provided by black.

docs/autogen/apigen.py
```diff
@@ -3,44 +3,48 @@
...
+
 import jinja2
 
 DECLARATIONS = {
     "func": [],
     "macro": [],
     "struct": [],
     "enum": [],
-}
+}  # type: dict
```
Baekalfen, line 21: Is it `mypy` that requires this? Why not:
```python
DECLARATIONS : dict = {
     "func": [],
     "macro": [],
     "struct": [],
     "enum": [],
 }
```

docs/autogen/apigen.py
```diff
@@ -157,24 +160,25 @@ def main(args):
...
+        "xnvme",
     ]
 
     syms = symbols(args, namespaces)
 
     pps = {}
-    for ns in syms:
-        pps[ns] = find_pp(args, syms[ns])
+    for nsprefix, val in syms.items():
+        pps[nsprefix] = find_pp(val)
```
Baekalfen, line 170: Don't fix this unless you want to, but it's perfect for a dict-comprehension:
```python
pps = {k:find_pp(val) for k,v in syms.items()}
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,134 @@
...
+# pre-commit runs pylint from an isolated virtualenv. Many of pylint's checks perform dynamic
+# analysis which will fail there. You may find configuring pylint as a local hook more useful.
+- repo: https://github.com/pre-commit/mirrors-pylint
+  rev: 'v3.0.0a4'
+  hooks:
+  - id: pylint
+    name: Python-lint-pylint
+    args:
+    - '--jobs=4' # Run with 4 threads
+    - '--good-names=i,j,k,n,f,_,Run,ex' # Add 'f' for with open()
```
Baekalfen, line 116: `v` is also common in `for k,v in some_dict.items()`

Baekalfen (approved):
Looks good. Nothing critical

docs/autogen/apigen.py
safl: Yup, mypy complains when the type-annotation is not there

.pre-commit-config.yaml
safl: Fixing...

safl: Fixed.

safl:
Thanks for the feedback. When checks pass I'll move it forward.

PR #66, opened 2022-04-01
Misc fixes picked off from feature additions

safl:
Picked these of to go into ``next`` sooner than the PR branches they lived on would be merged.
@Baekalfen Regarding [build:meson: exclude SPDK from building on aarch64 systems](https://github.com/OpenMPDK/xNVMe/commit/01fbe7c9e09f3a7546f07cd3efd0212bc9cf2f3d), is this MacOS specific or a general aarch64 issue?

PR #55, opened 2022-03-22
be:psync,macos: unify pread/pwrite across Unix-like platforms

Baekalfen:
Merge the identical code for sync backends for FreeBSD, Linux and macOS.
By unifying the code, we eliminate redundancy, and make the code more
concise.

With the unified code, a trivial macOS backend is added. This can be
used as the foundation for more platform-specific features.

Signed-off-by: Mads Ynddal <m.ynddal@samsung.com>

PR #68, opened 2022-04-04
be:macos: Add simple backend for macOS

Baekalfen:
New backend which supports macOS mostly using existing POSIX functionality. Although with the addition to  use the NVMeSMARTLibExternal.h provided by Xcode to query some information from the target disk. This does not enable arbitrary NVMe commands to be sent to a disk, but rather support for Get Log Page and Identify commands.

### DIFF
--- a/toolbox/cijoe-pkg-xnvme/testcases/examples-xnvme_hello.sh
+++ b/toolbox/cijoe-pkg-xnvme/testcases/examples-xnvme_hello.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Verify that `xnvme_hello` runs without error
 #
--- a/toolbox/cijoe-pkg-xnvme/testplans/xnvme-base-macos.plan
+++ b/toolbox/cijoe-pkg-xnvme/testplans/xnvme-base-macos.plan
@@ -1,0 +1,31 @@
+---
+descr: Verify base xNVMe functionality using the macOS backend
+descr_long: |
+  Exercises 
+  * library-information
+  * device enumeration
+  * admin-commands idfy-ns, idfy-ctrlr, get-log, get/set feature
+  * queue-init/teardown of io_uring/libaio/posix
+  Requires
+  * XNVME_URI points to an NVMe namespace e.g. /dev/disk4
+evars:
+  NVME_NSTYPE: "lblk"
+  DEV_TYPE: "nvme"
+  XNVME_BE: "macos"
+  XNVME_ADMIN: "macos"
+
+
+testsuites:
+- name: xnvme_core
+  alias: "xNVMe: verify base functionality using be:macos"
+
+- name: xnvme_admin_idfy_log
+  alias: "xNVMe: verify base functionality using be:macos"
+
+- name: xnvme_queue
+  alias: "xNVMe: verify base functionality using be:macos/async:posix"
+  evars: {XNVME_ASYNC: "posix"}
+
+- name: xnvme_queue
+  alias: "xNVMe: verify base functionality using be:macos/async:thrpool"
+  evars: {XNVME_ASYNC: "thrpool"}
--- a/Makefile
+++ b/Makefile
@@ -250,7 +250,7 @@ endef
 .PHONY: clean-subprojects
 clean-subprojects:
 	@echo "## xNVMe: make clean-subprojects"
-	@if [ -d "subprojects/spdk" ]; then cd subprojects/spdk &&$(MAKE) clean; fi;
+	@if [ -d "subprojects/spdk" ] && [ ${PLATFORM_ID} != 'Darwin' ]; then cd subprojects/spdk &&$(MAKE) clean; fi;
 	@if [ -d "subprojects/fio" ]; then cd subprojects/fio && $(MAKE) clean; fi;
 	@if [ -d "subprojects/liburing" ]; then cd subprojects/liburing &&$(MAKE) clean; fi;
 	rm -fr $(BUILD_DIR) || true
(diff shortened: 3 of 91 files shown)


Baekalfen:
Blocked by #55

safl:
Regarding the first commit. The issue you are experiencing is not a general ``aarch64`` issue.
Disabling the SPDK backend on all ``aarch`` is therefore not a good solution.
See here: https://github.com/OpenMPDK/xNVMe/pull/66
So, either:
1) Find the root cause of the issue and make a patch for SPDK, we can carry this patch:
  - ``subprojects/packagefiles/spdk/patches/`` until we get the patch upstreamed in SPDK
2) Alter the commit to do a work-around specific to where you are seeing the issue.

safl:
Regarding the second commit ``be:psync,macos: unify pread/pwrite across Unix-like platforms``. Thanks for taking a stab at this, it is an area that would be really great to improve for re-use. However, I am not certain this commit is the way to go. There are some things that need a bit more investigation when it comes the platform-specific behavior and error-handling on even the simplest things like ``pread()``/``pwrite()``.
Also, on MacOS, it looks like some special handling is needed for fsync():
* https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html#//apple_ref/doc/man/2/fsync
* https://twitter.com/marcan42/status/1494213862970707969

So, instead of the refactor, how about just doing this:
* export the POSIX ``cmd_io()`` implementation and re-use it in ``lib/xnvme_be_macos_sync.c``
* add cmd_iov() in ``lib/xnvme_be_macos_sync.c``

Right now, I would prefer not to alter the behavior of all of POSIX, Linux, and FreeBSD for the sake of the MacOS implementation. I would much rather like that the MacOS backend is added and then we can investigate a proper re-usable ``common``/``base`` implementation. Taking other things into account such as the ``common/base`` backend state and other things that are commonly needed/re-used.

**TLDR**; to unblock the addition of the MacOS addition, then just add the MacOS implementation and re-use the ``cmd_io()`` from the POSIX backend.

Baekalfen:
> Regarding the first commit `build:meson: exclude SPDK from building on aarch64 systems`. The issue you are experiencing is not a general `aarch64` issue. Disabling the SPDK backend on all `aarch` is therefore not a good solution. See here: #66 So, either:
> 
> 1. Find the root cause of the issue and make a patch for SPDK, we can carry this patch:
> 
> * `subprojects/packagefiles/spdk/patches/` until we get the patch upstreamed in SPDK
> 
> 2. Alter the commit to do a work-around specific to where you are seeing the issue.

I'll drop it from here. It was just part of the branch before you made the comments in #66

Baekalfen:
> Regarding the second commit `be:psync,macos: unify pread/pwrite across Unix-like platforms`. Thanks for taking a stab at this, it is an area that would be really great to improve for re-use. However, I am not certain this commit is the way to go. There are some things that need a bit more investigation when it comes the platform-specific behavior and error-handling on even the simplest things like `pread()`/`pwrite()`. Also, on MacOS, it looks like some special handling is needed for fsync():
> 
> * https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html#//apple_ref/doc/man/2/fsync
> * https://twitter.com/marcan42/status/1494213862970707969
> 
> So, instead of the refactor, how about just doing this:
> 
> * export the POSIX `cmd_io()` implementation and re-use it in `lib/xnvme_be_macos_sync.c`
> * add cmd_iov() in `lib/xnvme_be_macos_sync.c`
> 
> Right now, I would prefer not to alter the behavior of all of POSIX, Linux, and FreeBSD for the sake of the MacOS implementation. I would much rather like that the MacOS backend is added and then we can investigate a proper re-usable `common`/`base` implementation. Taking other things into account such as the `common/base` backend state and other things that are commonly needed/re-used.
> 
> **TLDR**; to unblock the addition of the MacOS addition, then just add the MacOS implementation and re-use the `cmd_io()` from the POSIX backend.

Maybe I should have pointed it out more clearly. That commit is actually from #55

Baekalfen:
@safl If you have no more comments, then I consider this PR ready for merge.

toolbox/cijoe-pkg-xnvme/testplans/xnvme-base-macos.plan
```diff
@@ -0,0 +1,32 @@
...
+  * admin-commands idfy-ns, idfy-ctrlr, get-log, get/set feature
+  * queue-init/teardown of io_uring/libaio/posix
+  Requires
+  * XNVME_URI points to an NVMe namespace e.g. /dev/nvme0n1
+evars:
+  NVME_NSTYPE: "lblk"
+  DEV_TYPE: "nvme"
+  XNVME_BE: "macos"
+  XNVME_ADMIN: "macos"
+  XNVME_URI: "/dev/disk4"
```
safl, line 16: The ``XNVME_URI`` be removed from here and instead adjust the construction of ``XNVME_URI`` in ``toolsbox/cijoe-pkg-xnvme/envs/refenv-xnvme.sh`` and ``toolsbox/cijoe-pkg-xnvme/envs/refenv-xnvme.sh`.

toolbox/cijoe-pkg-xnvme/testcases/examples-xnvme_hello.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
```
safl, line 1: The commit title should be something along the lines of: ``toolbox:cijoe: change shebang to '/usr/bin/env bash'``

And the body should describe why this is needed.

safl (changes requested):
See in-line comments, adding to those:
* Could the introduction of ``additional_link_args`` / ``link_args`` in meson be handled in a manner similar to ``dependency('threads')``?

safl (approved):

safl:
Thanks!

PR #72, opened 2022-04-12
dsm: Implement DSM support

Baekalfen:
Implement tool and library support for the Dataset Management command
(DSM) found the in NVMe specifications.

### DIFF
--- a/include/libxnvme_nvm.h
+++ b/include/libxnvme_nvm.h
@@ -111,6 +111,23 @@ xnvme_nvm_scopy(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t sdlba,
 		struct xnvme_spec_nvm_scopy_fmt_zero *ranges, uint8_t nr,
 		enum xnvme_nvm_scopy_fmt copy_fmt);
 
+/**
+ * Deallocate or hint at read/write usage of a range.
+ *
+ * @param ctx Pointer to command context (::xnvme_cmd_ctx)
+ * @param nsid Namespace Identifier
+ * @param dsm_range A buffer of ranges, each with context attributes, slba and nlb
+ * @param nr Number of ranges in the given ranges-buffer, zero-based value
+ * @param ad The NVM subsystem may deallocate all provided ranges
+ * @param idw The dataset should be optimized for write access as an integral unit
+ * @param idr The dataset should be optimized for read access as an integral unit
+ *
+ * @return On success, 0 is returned. On error, -1 is returned.
+ */
+int
+xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, struct xnvme_spec_dsm_range *dsm_range,
+	      uint8_t nr, bool ad, bool idw, bool idr);
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/xnvme_be_ramdisk_sync.c
+++ b/lib/xnvme_be_ramdisk_sync.c
@@ -46,6 +46,10 @@ xnvme_be_ramdisk_sync_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_
 	case XNVME_SPEC_FS_OPC_FLUSH:
 		break;
 
+	case XNVME_SPEC_NVM_OPC_DATASET_MANAGEMENT:
+		// Just pass on the command, as it is just a hint to the controller.
+		break;
+
 	default:
 		XNVME_DEBUG("FAILED: nosys opcode: %d", ctx->cmd.common.opcode);
 		return -ENOSYS;
--- a/lib/xnvme_nvm.c
+++ b/lib/xnvme_nvm.c
@@ -113,6 +113,20 @@ xnvme_prep_nvm(struct xnvme_cmd_ctx *ctx, uint8_t opcode, uint32_t nsid, uint64_
 }
 
 int
+xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, struct xnvme_spec_dsm_range *dsm_range,
+	      uint8_t nr, bool ad, bool idw, bool idr)
+{
+	ctx->cmd.common.opcode = XNVME_SPEC_NVM_OPC_DATASET_MANAGEMENT;
+	ctx->cmd.common.nsid = nsid;
+	ctx->cmd.dsm.nr = nr;
+	ctx->cmd.dsm.ad = ad;
+	ctx->cmd.dsm.idw = idw;
+	ctx->cmd.dsm.idr = idr;
+
+	return xnvme_cmd_pass(ctx, (void *)dsm_range, sizeof(*dsm_range) * nr, NULL, 0);
+}
+
+int
 xnvme_nvm_scopy(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t sdlba,
 		struct xnvme_spec_nvm_scopy_fmt_zero *ranges, uint8_t nr,
 		enum xnvme_nvm_scopy_fmt copy_fmt)
--- a/lib/xnvmec.c
+++ b/lib/xnvmec.c
@@ -939,6 +939,26 @@ static struct xnvmec_opt_attr xnvmec_opts[] = {
 		.name = "\nWith <args> for backend:\n",
 		.descr = "",
 	},
+
+	{
+		.opt = XNVMEC_OPT_AD,
+		.vtype = XNVMEC_OPT_VTYPE_NUM,
+		.name = "ad",
+		.descr = "If set, deallocate ranges",
+	},
+	{
+		.opt = XNVMEC_OPT_IDW,
+		.vtype = XNVMEC_OPT_VTYPE_NUM,
+		.name = "idw",
+		.descr = "If set, hint to use range as an integral unit when writing.",
+	},
+	{
+		.opt = XNVMEC_OPT_IDR,
+		.vtype = XNVMEC_OPT_VTYPE_NUM,
+		.name = "idr",
+		.descr = "If set, hint to use range as an integral unit when reading.",
+	},
+
 	{
 		.opt = XNVMEC_OPT_END,
 		.vtype = XNVMEC_OPT_VTYPE_NUM,
@@ -1531,7 +1551,15 @@ xnvmec_assign_arg(struct xnvmec *cli, struct xnvmec_opt_attr *opt_attr, char *ar
 	case XNVMEC_OPT_NSR:
 		args->nsr = num;
 		break;
-
+	case XNVMEC_OPT_AD:
+		args->ad = arg ? num : 0;
+		break;
+	case XNVMEC_OPT_IDW:
+		args->idw = arg ? num : 0;
+		break;
+	case XNVMEC_OPT_IDR:
+		args->idr = arg ? num : 0;
+		break;
 	case XNVMEC_OPT_POSA_TITLE:
 	case XNVMEC_OPT_NON_POSA_TITLE:
 	case XNVMEC_OPT_ORCH_TITLE:
--- a/tools/xnvme.c
+++ b/tools/xnvme.c
@@ -623,6 +623,50 @@ exit:
 	return err;
 }
 
+static int
+sub_dsm(struct xnvmec *cli)
+{
+	// We can only define one range in CLI
+	uint32_t nr = 1;
+	struct xnvme_spec_dsm_range *dsm_range;
+
+	struct xnvme_dev *dev = cli->args.dev;
+	struct xnvme_cmd_ctx ctx = xnvme_cmd_ctx_from_dev(dev);
+	uint32_t nsid = cli->args.nsid;
+	uint64_t slba = cli->args.slba;
+	uint32_t nlb = cli->args.nlb;
+	bool ad = cli->args.ad;
+	bool idw = cli->args.idw;
+	bool idr = cli->args.idr;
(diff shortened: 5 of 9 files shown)


include/libxnvme_nvm.h
```diff
@@ -96,6 +96,21 @@ xnvme_nvm_scopy(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t sdlba,
...
+ * @param nsid Namespace Identifier
+ * @param nr Number of ranges in the given ranges-buffer, zero-based value
+ * @param ad The NVM subsystem may deallocate all provided ranges
+ * @param idw The dataset should be optimized for write access as an integral unit
+ * @param idr The dataset should be optimized for read access as an integral unit
+ *
+ * @return On success, 0 is returned. On error, -1 is returned.
+ */
+int
+xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint8_t nr, bool ad, bool idw, bool idr);
```
safl, line 112: This needs a pointer to the range which the the command is supported to operate on.

safl: As an array of ``struct xnvme_spec_dsm_range``.

tools/xnvme.c
```diff
@@ -623,6 +623,34 @@ sub_library_info(struct xnvmec *XNVME_UNUSED(cli))
...
 }
 
+static int
+sub_dsm(struct xnvmec *cli)
+{
+	struct xnvme_dev *dev = cli->args.dev;
+	struct xnvme_cmd_ctx ctx = xnvme_cmd_ctx_from_dev(dev);
+	uint32_t nsid = cli->args.nsid;
+	uint32_t nr = cli->args.nr;
+	uint32_t ad = cli->args.ad;
```
safl, line 633: The ``ad``, ``idw``, and ``idr`` should probably just be bools.

tools/xnvme.c
```diff
@@ -623,6 +623,34 @@ sub_library_info(struct xnvmec *XNVME_UNUSED(cli))
...
+	uint32_t idw = cli->args.idw;
+	uint32_t idr = cli->args.idr;
+	int err;
+
+	xnvmec_pinf("xNVMe DSM");
+	if (!cli->given[XNVMEC_OPT_NSID]) {
+		nsid = xnvme_dev_get_nsid(cli->args.dev);
+	}
+
```
safl, line 642: It should be possible to create at least one range when calling this.

tools/xnvme.c
```diff
@@ -897,6 +925,22 @@ static struct xnvmec_sub g_subs[] = {
...
+		sub_dsm,
+		{
+			{XNVMEC_OPT_URI, XNVMEC_POSA},
+			{XNVMEC_OPT_NSID, XNVMEC_LREQ},
+			{XNVMEC_OPT_NR, XNVMEC_LREQ},
+			{XNVMEC_OPT_AD, XNVMEC_LREQ},
+			{XNVMEC_OPT_IDW, XNVMEC_LREQ},
+			{XNVMEC_OPT_IDR, XNVMEC_LREQ},
+			{XNVMEC_OPT_DEV_NSID, XNVMEC_LOPT},
+			{XNVMEC_OPT_BE, XNVMEC_LOPT},
```
safl, line 941: For the user to supply the range, it could use SLBA and NLB here.
Should be possible to supply context-attributes as well.

safl (changes requested):
See comments in-line, adding to those:
* A test in ``tests/`` should be added with a couple of sub-commands
  - Checking for support of the command
  - Print the idfy-related fields
  - Logical verification of what can be verified, to start with just a basic verification that a DSM command completes successfully when DSM is supported.
* Several ``cijoe-pkg-xnvme`` testcases
  - Add a testcase invoking the before-mentioned tests
  - Add a testcase invoking the ``xnvme dsm``
  - Add these testcases to the ``testplans/xnvme-nvm-*.plan`` testplans

include/libxnvme_nvm.h
Baekalfen: Done

tools/xnvme.c
Baekalfen: Done

tools/xnvme.c
Baekalfen: The CLI can now handle one range. The API can handle more

tools/xnvme.c
Baekalfen: Done

lib/xnvme_be_ramdisk_sync.c
```diff
@@ -46,6 +46,10 @@ xnvme_be_ramdisk_sync_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_
 	case XNVME_SPEC_FS_OPC_FLUSH:
 		break;
 
+	case XNVME_SPEC_NVM_OPC_DATASET_MANAGEMENT:
```
karlowich, line 49: I think this can be moved up to the to `XNVME_SPEC_NVM_OPC_FLUSH` case, the comment is implied and I don't think there's any reason to have a separate case just for that

lib/xnvmec.c
```diff
@@ -939,6 +939,33 @@ static struct xnvmec_opt_attr xnvmec_opts[] = {
 		.name = "\nWith <args> for backend:\n",
 		.descr = "",
 	},
+
```
karlowich, line 942: The empty lines here should be removed to follow the convention of the previous blocks

lib/xnvmec.c
```diff
@@ -939,6 +939,33 @@ static struct xnvmec_opt_attr xnvmec_opts[] = {
...
+	{
+		.opt = XNVMEC_OPT_AD,
+		.vtype = XNVMEC_OPT_VTYPE_NUM,
+		.name = "ad",
+		.descr = "If set, deallocate ranges when calling Dataset "
```
karlowich, line 947: I don't think there's any need for specifying that this (and the 2 other opts below) is for DSM.
This will be clear from the context when they're listed in `xnvme dsm --help`

karlowich (changes requested):
This is great!
I have left a few nitpicky comments

lib/xnvme_be_ramdisk_sync.c
Baekalfen: I think it's best to keep them separate, as it could also imply that DSM is related or should be handled the same as a flush. Maybe the comment is trivial if you know about DSM, but it can't hurt to tell the reader why we are skipping on it.

lib/xnvmec.c
Baekalfen: I'm not sure I follow. We make new-lines in this list at line 887 and 880.

lib/xnvmec.c
Baekalfen: So you want to remove the "... when calling Dataset Management command (DSM)" part? I agree

lib/xnvme_be_ramdisk_sync.c
karlowich: That's a good point.

lib/xnvmec.c
karlowich: Exactly

Baekalfen: Done

karlowich (approved):

include/libxnvme_nvm.h
```diff
@@ -111,6 +111,22 @@ xnvme_nvm_scopy(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t sdlba,
...
 
+/**
+ * Deallocate or hint at read/write usage of a range.
+ *
+ * @param ctx Pointer to command context (::xnvme_cmd_ctx)
+ * @param nsid Namespace Identifier
+ * @param nr Number of ranges in the given ranges-buffer, zero-based value
+ * @param ad The NVM subsystem may deallocate all provided ranges
+ * @param idw The dataset should be optimized for write access as an integral unit
+ * @param idr The dataset should be optimized for read access as an integral unit
```
safl, line 122: This seems to be missing a comment of ``dsm_range``

lib/xnvme_nvm.c
```diff
@@ -112,6 +112,21 @@ xnvme_prep_nvm(struct xnvme_cmd_ctx *ctx, uint8_t opcode, uint32_t nsid, uint64_
...
+{
+	ctx->cmd.common.opcode = XNVME_SPEC_NVM_OPC_DATASET_MANAGEMENT;
+	ctx->cmd.common.nsid = nsid;
+	ctx->cmd.dsm.nr = nr;
+	ctx->cmd.dsm.ad = ad;
+	ctx->cmd.dsm.idw = idw;
+	ctx->cmd.dsm.idr = idr;
+
+	return xnvme_cmd_pass(ctx, NULL, 0, (void *)dsm_range,
+			      sizeof(struct xnvme_spec_dsm_range) * nr);
```
safl, line 127: This should should be passed to the ``dbuf_nbytes`` argument of ``xnvme_cmd_pass()`` not ``mbuf_nbytes``.

safl: Additionally, then:
``sizeof(struct xnvme_spec_dsm_range) * nr``
Could be expressed as:
``sizeof(*dsm_range) * nr``
This is convenient as established as relation between the argument and its type.

include/libxnvme_nvm.h
```diff
@@ -111,6 +111,22 @@ xnvme_nvm_scopy(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t sdlba,
...
+ *
+ * @return On success, 0 is returned. On error, -1 is returned.
+ */
+int
+xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint8_t nr, bool ad, bool idw, bool idr,
```
safl, line 127: The addition of ``dsm_range`` fits better before the ``nr``, e.g.:
```
int
xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, struct xnvme_spec_dsm_range *dsm_range, uint8_t nr, bool ad, bool idw, bool idr);
```

tools/xnvme.c
```diff
@@ -623,6 +623,45 @@ sub_library_info(struct xnvmec *XNVME_UNUSED(cli))
 	return err;
 }
 
+static int
+sub_dsm(struct xnvmec *cli)
+{
+	// We can only define one range in CLI
+	uint32_t nr = 1;
+	struct xnvme_spec_dsm_range dsm_range;
```
safl, line 631: As a data-payload, then the ``dsm_range`` should be allocated by ``xnvme_buf_alloc()`` not on the stack.

tools/xnvme.c
```diff
@@ -623,6 +623,45 @@ sub_library_info(struct xnvmec *XNVME_UNUSED(cli))
...
+
+	dsm_range = (struct xnvme_spec_dsm_range){
+		.cattr = 0,
+		.nlb = nlb,
+		.slba = slba,
+	};
+
+	err = xnvme_nvm_dsm(&ctx, nsid, nr, ad, idw, idr, &dsm_range);
+	if (err || xnvme_cmd_ctx_cpl_status(&ctx)) {
+		xnvmec_perr("xnvme_adm_dsm()", err);
```
safl, line 656: Seems like a typo here? Should ``xnvme_adm_dsm`` --> ``xnvme_nvm_dsm``?

include/libxnvme_nvm.h
safl: That is, such that the array and the count of elements in the array are grouped together, with the array before the member-count.

include/libxnvme_nvm.h
karlowich: Done

tools/xnvme.c
karlowich: Fixed!

tools/xnvme.c
karlowich: Fixed!

include/libxnvme_nvm.h
karlowich: Fixed!

lib/xnvme_nvm.c
karlowich: Fixed!

lib/xnvme_nvm.c
```diff
@@ -112,6 +112,21 @@ xnvme_prep_nvm(struct xnvme_cmd_ctx *ctx, uint8_t opcode, uint32_t nsid, uint64_
...
+	      struct xnvme_spec_dsm_range *dsm_range)
+{
+	ctx->cmd.common.opcode = XNVME_SPEC_NVM_OPC_DATASET_MANAGEMENT;
+	ctx->cmd.common.nsid = nsid;
+	ctx->cmd.dsm.nr = nr;
+	ctx->cmd.dsm.ad = ad;
+	ctx->cmd.dsm.idw = idw;
+	ctx->cmd.dsm.idr = idr;
+
+	return xnvme_cmd_pass(ctx, NULL, 0, (void *)dsm_range,
```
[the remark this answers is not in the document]
karlowich: Done

safl (approved):

safl (commented):
This is looking mighty fine, thanks! I little nitpick, how about changing the brief commit-message ``dsm: Implement DSM support`` to either:

* ``add support for the DSM command``

or spelling out all the components it touches:

* ``api,tools,cli,ramdisk,toolbox: add support for the DSM command``

I lean more towards the latter...

safl (approved):
Awesome, thanks!

PR #78, opened 2022-04-25
Revive the xNVMe Python language bindings

safl:
xNVMe previously had Python bindings, however, these where manually written wrappers which were not properly tested, thus they became stale and were removed from the repository.

However, recently when providing examples of how to dynamically load xNVMe from C and Python, then it seemed worth-while to explore just how much is possible via ``ctypes``. With a couple of non-intrusive changes to the public C API (adding a couple of includes), then it is possible to auto-generate a complete ctypes-wrapping for the C API. With the essentials auto-generated, then using ``ctypes`` is a breeze.

With this PR a full xNVMe C API mapping is available along with automatic dynamic loading. This module can serve for those wanting to experiment with the API from the comforts of Python. Additionally, as a reference for adding Cython bindings as well, for those going beyond poking at the API.

- [x] Add testcases and testplan for CIJOE invoking and verifying the library
- [x] Add testing to CI
- [x] Add description of the Python module in the docs

Have a look at the ``toolbox/py_xnvme_pkg/README.rst`` for more information.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -323,6 +323,15 @@ jobs:
     - name: Execute 'xnvme library-info'
       run: xnvme library-info
 
+    - name: Check pkg-config
+      run: |
+        pkg-config xnvme --libs
+        pkg-config xnvme --variable=datadir
+        pkg-config xnvme --variable=includedir
+        pkg-config xnvme --variable=libdir
+        pkg-config xnvme --variable=pcfiledir
+        pkg-config xnvme --variable=prefix
+        ls -lh $(pkg-config xnvme --variable=libdir) | grep xnvme
 
     - name: Check Python, pip, platform, and sysconfig
       run: |
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -323,8 +323,13 @@ jobs:
     - name: Execute 'xnvme library-info'
       run: xnvme library-info
 
-    - name: Check pip
-      run: python3 -m pip --version
+
+    - name: Check Python, pip, platform, and sysconfig
+      run: |
+        python3 --version
+        python3 -m pip --version
+        python3 -c "import platform; print(platform.system()); print(platform.uname())"
+        python3 -c 'import sysconfig; print(sysconfig.get_config_var("SHLIB_SUFFIX"))'
 
   #
   # Build on macOS
@@ -381,8 +386,12 @@ jobs:
         pkg-config xnvme --variable=prefix
         ls -lh $(pkg-config xnvme --variable=libdir) | grep xnvme
 
-    - name: Check pip
-      run: python3 -m pip --version
+    - name: Check Python, pip, platform, and sysconfig
+      run: |
+        python3 --version
+        python3 -m pip --version
+        python3 -c "import platform; print(platform.system()); print(platform.uname())"
+        python3 -c 'import sysconfig; print(sysconfig.get_config_var("SHLIB_SUFFIX"))'
 
   #
   # Build on Windows
@@ -437,8 +446,12 @@ jobs:
         xnvme.exe enum
         xnvme.exe library-info
 
-    - name: Check pip
-      run: python3 -m pip --version
+    - name: Check Python, pip, platform, and sysconfig
+      run: |
+        python3 --version
+        python3 -m pip --version
+        python3 -c "import platform; print(platform.system()); print(platform.uname())"
+        python3 -c 'import sysconfig; print(sysconfig.get_config_var("SHLIB_SUFFIX"))'
 
   #
   # Build on run CIJOE testplans
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -250,10 +250,55 @@ jobs:
       run: docker push refenv/xnvme
 
   #
+  # Build Python-bindings
+  #
+  build-python:
+    needs: source-archive
+    runs-on: ubuntu-latest
+    if: ((github.event_name == 'pull_request') || ((github.event_name == 'workflow_dispatch') && ((github.event.inputs.job
+      == 'all') || contains('build-linux', github.event.inputs.job))))
+    container: refenv/alpine-bash
+
+    steps:
+    - name: Retrieve the source-archive
+      uses: actions/download-artifact@v2
+      with:
+        name: archive-src
+    - name: Extract the full-source-archive
+      run: |
+        tar xzf xnvme*.tar.gz --strip 1
+        rm xnvme*.tar.gz
+
+    - name: Install build-requirements
+      run: |
+        ./toolbox/pkgs/alpine-latest.sh
+
+    - name: Python bindings, system packages
+      run: |
+        apk add clang-libs
+
+    - name: Python bindings, Python packages
+      run: |
+        cd python
+        python3 -m pip install -r requirements.txt
+
+    - name: Python bindings, generate, patch, and build-sdist
+      run: |
+        cd python
+        make build
+        for file in dist/*.tar.gz; do mv $file ${file//xnvme/xnvme-py}; done
+
+    - name: Upload Python package
+      uses: actions/upload-artifact@v2
+      with:
+        name: python-pkg
+        path: python/dist/xnvme-py-*.tar.gz
+
+  #
   # Build on Linux using different Linux distributions
   #
   build-linux:
-    needs: [source-archive, source-format-check]
+    needs: [source-archive, source-format-check, build-python]
     runs-on: ubuntu-latest
     if: ((github.event_name == 'pull_request') || ((github.event_name == 'workflow_dispatch') && ((github.event.inputs.job
       == 'all') || contains('build-linux', github.event.inputs.job))))
@@ -340,11 +385,22 @@ jobs:
         python3 -c "import platform; print(platform.system()); print(platform.uname())"
         python3 -c 'import sysconfig; print(sysconfig.get_config_var("SHLIB_SUFFIX"))'
 
+    - name: Retrieve the xNVMe Python bindings (prebuilt sdist)
+      uses: actions/download-artifact@v2
+      with:
+        name: python-pkg
+
+    - name: Install and test xNVMe Python bindings
+      if: (!contains(matrix.container.ver, 'bookworm'))
+      run: |
+        python3 -m pip install xnvme-py*.tar.gz --user
+        python3 -m pytest --pyargs xnvme
+
   #
   # Build on macOS
   #
   build-macos:
-    needs: [source-archive, source-format-check]
+    needs: [source-archive, source-format-check, build-python]
     runs-on: ${{ matrix.runner.os }}-${{ matrix.runner.ver }}
     if: ((github.event_name == 'pull_request') || ((github.event_name == 'workflow_dispatch') && ((github.event.inputs.job
       == 'all') || contains('build-macos', github.event.inputs.job))))
@@ -401,6 +457,16 @@ jobs:
         python3 -m pip --version
         python3 -c "import platform; print(platform.system()); print(platform.uname())"
         python3 -c 'import sysconfig; print(sysconfig.get_config_var("SHLIB_SUFFIX"))'
+
+    - name: Retrieve the xNVMe Python bindings (prebuilt sdist)
+      uses: actions/download-artifact@v2
+      with:
+        name: python-pkg
+
+    - name: Install and test xNVMe Python bindings
+      run: |
+        python3 -m pip install xnvme-py*.tar.gz --user
+        python3 -m pytest --pyargs xnvme
 
   #
   # Build on Windows
--- a/docs/python/index.rst
+++ b/docs/python/index.rst
@@ -1,0 +1,188 @@
+.. _sec-python:
+
+========
+ Python
+========
+
+**xNVMe** language bindings for Python provide access to the **xNVMe** C API
+via :ref:`sec-python-ctypes` and :ref:`sec-python-Cython`. Two language binding
+interfaces are provided as they serve different use-cases.
+
+* The :ref:`sec-python-ctypes` bindings are intended for interactive usage when
+  experimenting with the **xNVMe** API, as well as integration into
+  adminstrative tools such as Web-UI for managing fabrics
+  :ref:`sec-tutorials-fabrics`.
+
+* The :ref:`sec-python-cython` bindings are intended for Cython applications,
+  that is, to provide Cython bindings readily available such that you don't
+  have to make them in order to use **xNVMe** in your Cython code.
+
+The library bindings are a work-in-progress, provided as a preview of what can
+be done when integrating with Python. See, section :ref:`sec-python-todo` for a
+list of known short-comings and areas to improve. Your input on these are most
+welcome along with pull-requests to improve the interfaces.
+
+Install
+=======
+
+Before installing, then ensure that the following system dependencies are met:
+
+* Python 3.6+
+* **xNVMe** installed on your system
+* ``pkg-config`` installed and able to locate **xNVMe** on your system
+
+See, the :ref:`sec-building` on how to build and install xNVMe. With the above
+then you can go ahead and install the bindings using ``pip``::
+
+  # install: for the current user
+  python3 -m pip install xnvme --user
+
+For a description of building and installing the package from source, then see
+the :ref:`sec-python-development` section.
+
+.. note:: The **xNVMe** Python package comes with pytests, thus, you can verify
+   your installation by running: ``python3 -m pytest --pyargs xnvme``.
+
+.. _sec-python-ctypes:
+
+ctypes
+======
+
+Python ctypes_ comes with CPython_, that is, with the reference implementation
+of the Python intepreter. When using the ctypes_ bindings, the only other
+runtime dependency is pytest_.
+
+Usage examples
+--------------
+
+See the files in ``bin/*``. At the time of writing this, there were three
+examples, two of which require no arguments, these are run like so::
+
+  # Enumerate devices on the system
+  ./bin/xpy_enumerate
+
+  # Print library-information
+  ./bin/xpy_libconf
+
+The third opens and queries a device for information, and thus requires a
+device as well as permissions, on a Linux system, running it could look
+something like this::
+
+  sudo -E ./bin/xpy_dev_open --uri /dev/nvme0n1
+
+.. note:: In the above example ``sudo -E`` is used. The ``sudo`` is there to
+   ensure permission to ``/dev/nvme0n1``, then ``-E`` is there such that the
+   **xNVMe** package installed as ``--user`` is available.
+
+.. _sec-python-cython:
+
+Cython
+======
+
+Python Cython_ ...
+
+.. _sec-python-development:
+
+Development
+===========
+
+The bindings are generated using the ctypeslib2_ tool, patched (injecting an
+extension to the library-loader), and then formatted with black_.
+
+The build-requirements are installable via ``requirements.txt``::
+
+  python3 -m pip install -r requirements.txt --user
+
+Furthermore, ``clang`` is needed on the system::
+
+  # Debian
+  sudo apt-get install libclang-dev
+
+  # Alpine
+  sudo apk add clang-libs
+
+A Makefile is available for common tasks, run::
+
+  make help
+
+To see what it provides / common-tasks during development. For example::
+
+  make build uninstall install test
+
+The above ``make`` invocation will generate the ctypes-mapping via
+``clang2py``, then patch the mappings using the auxilary scripts
+``aux/patch_ctypes_mapping.py``, adjust the style according to the conventions
+of ``black``, create a Python sdist package, install the package, and finally
+run the pytests.
+
+**CAVEAT:** the mappings produced by ``clang2py`` aren't stable. That is, the
+ordering in which classes are emitted can change from each invocation of the
+tool.
+
+.. _sec-python-todo:
+
+TODO
+====
+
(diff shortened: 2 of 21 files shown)


.github/workflows/verify.yml
```diff
@@ -370,6 +370,13 @@ jobs:
     - name: Execute 'xnvme library-info'
       run: xnvme library-info
 
+    - name: Build and install Python bindings
+      run: |
+        pip3 install pytest
+        cd toolbox/py_xnvme_pkg
+        make
+        make test
+
```
Baekalfen, line 379: Just a heads-up. You've only added it to `build-macos`. Linux is at line 326.

toolbox/py_xnvme_pkg/setup.py
```diff
@@ -0,0 +1,43 @@
...
+    here = os.path.abspath(os.path.dirname(__file__))
+
+    # intentionally *not* adding an encoding option to open, See:
+    #   https://github.com/pypa/virtualenv/issues/201#issuecomment-3145690
+    with codecs.open(os.path.join(here, *parts), "r") as pfp:
+        return pfp.read()
+
+
+setuptools.setup(
+    name="py-xNVMe",
```
Baekalfen, line 20: This is the name for `pip uninstall [name]`

toolbox/py_xnvme_pkg/Makefile
```diff
@@ -0,0 +1,123 @@
...
+	@echo "## py: make test [DONE]"
+
+define generate-ctypes-help
+# Generate definitions (xnvme_ctypes.header) from xNVMe public C API headers
+endef
+.PHONY: generate-ctypes
+generate-ctypes:
+	@echo "## py: gen"
+	@clang2py ${XNVME_CAPI_HEADERS} > xnvme/ctypes_mapping.py
+	@echo "## Remember to also call 'make format-all' to fix format"
```
Baekalfen, line 83: `format-all ` should probably be `patch-ctypes`

toolbox/py_xnvme_pkg/Makefile
```diff
@@ -0,0 +1,123 @@
+#
+# This Makefile serves as convenient command-line auto-completion when working on xNVMe-ctypes
+#
+#
+TOOLBOX_DIR?=../
+PROJECT_NAME=xnvme
+XNVME_CAPI_HEADERS=$(shell ls ../../include/libxnvme* | grep -v util | grep -v libxnvmec)
+XNVME_SO_PATH=$(shell pkg-config xnvme --variable=libdir)/libxnvme-shared.so
```
Baekalfen, line 8: This is Linux specific. I would also recommend, that we look for the lib in the local repo. For example on Mac: `$GIT_ROOT/builddir/lib/libxnvme-shared.dylib`

toolbox/py_xnvme_pkg/Makefile
```diff
@@ -0,0 +1,123 @@
...
+
+define uninstall-help
+# uninstall
+#
+# Prefix with 'sudo' when uninstalling a system-wide installation
+endef
+.PHONY: uninstall
+uninstall:
+	@echo "## py: make uninstall"
+	@pip3 uninstall ${PROJECT_NAME} --yes || echo "Cannot uninstall => That is OK"
```
Baekalfen, line 53: `setup.py` dictates the name to provide here. It's currently `py-xNVMe`, and not `xnvme` as `${PROJECT_NAME}` resolves to.

toolbox/py_xnvme_pkg/Makefile
```diff
@@ -0,0 +1,123 @@
...
+	@pytest tests
+	@echo "## py: make test [DONE]"
+
+define generate-ctypes-help
+# Generate definitions (xnvme_ctypes.header) from xNVMe public C API headers
+endef
+.PHONY: generate-ctypes
+generate-ctypes:
+	@echo "## py: gen"
+	@clang2py ${XNVME_CAPI_HEADERS} > xnvme/ctypes_mapping.py
```
Baekalfen, line 82: I had to specify additional arguments for macOS: `@clang2py --clang-args="-I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/ -I../../include/" ${XNVME_CAPI_HEADERS} > xnvme/ctypes_mapping.py`

toolbox/py_xnvme_pkg/xnvme/ctypes_mapping.py
```diff
@@ -0,0 +1,4272 @@
+# -*- coding: utf-8 -*-
```
Baekalfen, line 1: I would recommend not having the auto-generated mapping in the repo. As @safl noted offline, the ordering of the output is not stable, so git-diffs are impossible to check. I would recommend just adding it as part of the build process, and publish it as part of a package on PyPi.

toolbox/py_xnvme_pkg/xnvme/library_loader.py
```diff
@@ -0,0 +1,50 @@
...
+            check=True,
+            capture_output=True,
+        )
+        if not proc.returncode:
+            yield os.path.join(
+                proc.stdout.decode("utf-8").strip(), "libxnvme-shared.so"
+            )
+    except subprocess.CalledProcessError:
+        pass
+
```
Baekalfen, line 37: I would recommend adding a third-priority path with the local build: `yield f"{git_root}/builddir/lib/libxnvme-shared.dylib"`

toolbox/py_xnvme_pkg/Makefile
safl: Right! Thanks!

toolbox/py_xnvme_pkg/setup.py
safl: Right, fixing that.

toolbox/py_xnvme_pkg/Makefile
safl: Fixed.

.github/workflows/verify.yml
safl: Right. Fixed!
Windows and FreeBSD are still untested.
I have made a note of that in the TODO/Known Issues in ``docs/python/index.rst``.

toolbox/py_xnvme_pkg/Makefile
safl: Good spot. Usage of this is completely removed, it now relies on the library being installed on the system as that is where ``clang2py`` looked for headers anyway, so the library and the headers are needed there regardless.

toolbox/py_xnvme_pkg/Makefile
safl: Hmm, curious. This is probably something to explore further. To get this moving along then the build-testing is restricted to a single platform, the package generated and then tested on all the usual suspects.

toolbox/py_xnvme_pkg/xnvme/ctypes_mapping.py
safl: I agree. They are have been removed now.
The facilitate the build then a ci-job now builds the the sdist package and added as an artifact.
By doing so, then we get to do two things: 1) Automate the build of the package. 2) Test that the runtime-dependencies are met on all the usual targets.

toolbox/py_xnvme_pkg/xnvme/library_loader.py
safl: I don't like having a dependency on the source-repository in the library loader. Libraries should be installed such that ``pkg-config`` and regular tooling can locate them.
What I have done instead is add a platform specific extension, such that it loads using ``.so``, ``.dll`` and ``.dylib`` depending the platform it is running on.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+Furthermore, ``clang`` is needed on the system::
+
+  # Debian
+  sudo apt-get install libclang-dev
+
+  # Alpine
+  sudo apk add clang-libs
+
+  # macOS
+  sudo brew ??
```
Baekalfen, line 107: Installed with Xcode. You wouldn't get as far as building xNVMe without it.

python/Makefile
```diff
@@ -0,0 +1,122 @@
...
+#
+TOOLBOX_DIR?=../toolbox
+PROJECT_NAME=xnvme
+XNVME_CAPI_HEADERS=$(shell ls ../include/libxnvme* | grep -v util | grep -v libxnvmec)
+
+define default-help
+# invoke: 'make uninstall', 'make install'
+endef
+.PHONY: default
+default: uninstall install
```
Baekalfen, line 12: I think we should make the default `generate-ctypes patch-ctypes`, as I understand you'd idiomatically write `make && make install`.

python/Makefile
```diff
@@ -0,0 +1,122 @@
...
+	@python3 -m pip install .
+	@echo "## py: make install-system [DONE]"
+
+define test-help
+# Run pytest on tests/
+endef
+.PHONY: test
+test:
+	@echo "## py: make test"
+	@python3 -m pytest --pyargs xnvme
```
Baekalfen, line 71: Omit `--pyargs`

python/bin/xpy_dev_open
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+import argparse
+import sys
+
+import xnvme.ctypes_mapping as capi
+
+capi.guard_unloadable()
```
Baekalfen, line 7: I think it's better to move this to the library.

python/bin/xpy_enumerate
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+import sys
+from ctypes import POINTER, c_uint32, cast, pointer
+
+import xnvme.ctypes_mapping as capi
+
+capi.guard_unloadable()
```
Baekalfen, line 7: Same here

python/setup.py
```diff
@@ -0,0 +1,43 @@
...
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
+    ],
+    data_files=[
+        ("bin", glob.glob("bin/*")),
+    ],
+    package_dir={"": "./"},
+    packages=setuptools.find_packages(where="./"),
+    python_requires=">=3.6",
+    install_requires="pytest",
```
Baekalfen, line 42: Install wouldn't require `pytest`. Either use `tests_require` or now that you have a `python/pyproject.toml`, put it there instead.

python/requirements.txt
```diff
@@ -0,0 +1,5 @@
+# For ctypes-bindings
+black
+ctypeslib2
+# All
+pytest
```
Baekalfen, line 5: See comment in setup.py

python/xnvme/library_loader.py
```diff
@@ -0,0 +1,60 @@
...
+import ctypes.util
+import os
+import platform
+import subprocess
+
+SHARED_EXT = {
+    "linux": "so",
+    "windows": "dll",
+    "darwin": "dylib",
+}
```
Baekalfen, line 21: You can probably get Python's `sysconfig` to report this

.github/workflows/verify.yml
```diff
@@ -249,11 +249,62 @@ jobs:
...
+    - name: Install build-requirements
+      run: |
+        ./toolbox/pkgs/alpine-latest.sh
+
+    - name: Configure (slim), build, and install
+      run: |
+        make config-slim
+        make
+        make install
```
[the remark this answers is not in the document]
safl: Right. removed. This also removes the testing done at this stage which is also done in the later ``build-linux`` for Alpine,

.github/workflows/verify.yml
```diff
@@ -249,11 +249,62 @@ jobs:
...
+      run: |
+        cd python
+        make generate-ctypes patch-ctypes install build-sdist test
+        for file in dist/*.tar.gz; do mv $file ${file//xnvme/xnvme-py}; done
+
+    - name: Upload Python package
+      uses: actions/upload-artifact@v2
+      with:
+        name: python-pkg
+        path: python/dist/xnvme-py-*.tar.gz
```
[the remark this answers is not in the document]
safl: This is just the creation of a GitHUB artifact, such that the package is available for the other jobs.

.github/workflows/verify.yml
```diff
@@ -381,8 +456,21 @@ jobs:
...
+
+    - name: Retrieve the xNVMe Python bindings (prebuilt sdist)
+      uses: actions/download-artifact@v2
+      with:
+        name: python-pkg
+
+    - name: Install and test xNVMe Python bindings
+      run: |
+        python3 -m pip install xnvme-py*.tar.gz --user
+        python3 -m pytest --pyargs xnvme
```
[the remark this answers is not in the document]
safl: The ``-pyargs`` tells it to search the xnvme module installed on the system. By doing this, then we do not need to known where it is located.

python/setup.py
safl: I have gone with the approach of installing the testcases, this allows the ease-of-use of running the pytest via the package-discovery. This is why pytest is added as a runtime requirement.

python/bin/xpy_enumerate
safl: How about this would go on the list of things to refine in ``docs/python/index.rst``?

python/xnvme/library_loader.py
safl: Ohh, neat, ``SHLIB_SUFFIX`` looks like something that is useful for that purpose. Changing it to use that instead.

safl: ``sysconfig.get_config_var("SHLIB_SUFFIX")`` seems like the thing to use?

python/bin/xpy_dev_open
safl: How about this would go on the list of things to refine in ``docs/python/index.rst``? Have done so now.

safl:
@Baekalfen I have addressed your feedback, some with changes, some with clarification, and one thing post-poned as "nice-to-have" / added to ``docs/python/index.rst`` for future fixing.

python/xnvme/library_loader.py
Baekalfen: I haven't investigated it, but it (incorrectly?) provides `.so` on macOS
```
>>> import sysconfig
>>> sysconfig.get_config_var("SHLIB_SUFFIX")
'.so'
```

safl: No very useful then. Reverting back.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+then you can go ahead and install the bindings using ``pip``::
+
+  # install: for the current user
+  python3 -m pip install xnvme --user
+
+For a description of building and installing the package from source, then see
+the :ref:`sec-python-development` section.
+
+.. note:: The **xNVMe** Python package comes with pytests, thus, you can verify
+   your installation by running: ``python3 -m pytest --pyargs xnvme``.
```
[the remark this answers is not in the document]
safl: Yes, tests are shipped with the package.
``--pyargs xnvme`` is there to collect tests from the package.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. _sec-python-ctypes:
+
+ctypes
+======
+
+Python ctypes_ comes with Python, that is, with the reference implementation of
```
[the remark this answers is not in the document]
safl: That is Sphinx-doc notation for producing a link.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+the Python intepreter. When using the ``ctypes`` bindings, not other runtime
```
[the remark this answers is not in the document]
safl: fixed.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+dependencies are introduced.
+
+Usage examples
+--------------
+
+See the files in ``bin/*``, at the time of writing this, there were three
+examples, two of which require no arguments, these are run like so::
```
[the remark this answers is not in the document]
safl: Fixed.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Python Cython_ ...
+
+.. _sec-python-development:
+
+Development
+===========
+
+When generating the ctypes_ bindings, then **xNVMe** must be installed on the
+system, otherwise ``clang`` fails to locate the header-files. The bindings are
```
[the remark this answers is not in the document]
safl: Fixed.

python/Makefile
safl: As stated elsewhere, it is there for a reason.

python/setup.py
Baekalfen: I still think we should not have it here, and put it in either `tests_require` or `pyproject.toml`, as it's not part of normal runtime usage.

Also, isn't it supposed to be a list? `["pytest"]`?

.github/workflows/verify.yml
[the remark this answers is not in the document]
Baekalfen: You're right. I misunderstood the docs. Cool trick

python/xnvme/library_loader.py
safl: Yup. reverted.

python/setup.py
Baekalfen: The reason not to do it, would be because it's not a runtime dependency. I think it would be exceedingly rare somebody else is running our test suite. And as soon as we merge my Cython PR, the tests would no longer succeed without proper configuration.

But why not put it under the test requirements, if it's for testing?

safl: Adding to that, this is why ``python3 -m pytest --pyargs xnvme`` is added as a bold note in the docs right after installation.

docs/python/index.rst
```diff
@@ -0,0 +1,167 @@
...
+be done when integrating with Python. See, section :ref:`sec-python-todo`` for
+a list of known short-comings and areas to improve. Your input on these are
+most welcome along with pull-requests to improve the interfaces.
+
+Install
+=======
+
+Before installing, then ensure that the following runtime dependencies are met:
+
+* Python 3.6+
```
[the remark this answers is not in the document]
safl: Yeah, but that is because it only builds on Alpine which has some Python 3.7+ version, however, there are several instances of testing it on Python 3.6 (Leap / Ubuntu), also the ``setup.py`` has ``python_requires=">=3.6"``.

python/requirements.txt
safl: done.

safl:
@Baekalfen found some of your comments on the commits. Addressed those as well and pushed. Thanks for all the input! Your review improved this PR in a lot of ways:

* Reduced the ``build-python`` from ~50 seconds to  ~30 seconds
* Removed the generator-requirement needing the xNVMe libraries installed on the system
* Removed generated bindings from the repos
* Improved the ``Makefile`` helper, it now behaves idiomatically
* Several fixes on the docs and additions to todo

Thank you!

python/setup.py
safl: Also, this does not dictate that all tests are installed with the package, just the ones useful for the use-case described above.

Baekalfen: How about a different approach? Instead of implementing these as a pytest, why not just a simple function?

```py
def self_test():
     # Self-test for the dynamically loaded library

     # The 'capi' is loaded upon module access, this checks that it succeeded
     assert capi.is_loaded, "C API isn't loaded"

     # This verifies that the loaded library actually is the xNVMe library, by checking for
     # the existance of a xNVMe specific funtion.
     assert capi.xnvme_enumerate, "xnvme_enumerate is not present in library. Library isn't loaded correctly"

     # Verifies that at at least one function is callable
     assert capi.xnvme_libconf_pr(capi.XNVME_PR_DEF), "Failed to call function 'xnvme_libconf_pr in xNVMe. Library isn't loaded correctly"

     print('xNVMe ctypes bindings loaded correctly!')
```

Then in the docs: `python3 -c 'import xnvme.ctypes_bindings as xnvme; xnvme.self_test()'`

python/bin/xpy_dev_open
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+import argparse
+import sys
+
+import xnvme.ctypes_mapping as capi
```
Baekalfen, line 5: Not a hard requirement, but maybe consider `import xnvme.ctypes_mapping as xnvme`. Doesn't make much difference here, but in a larger program, it's maybe a bit ambiguous to call it `capi`.

safl: Sure. Changed to ``xnvme``.

python/setup.py
safl: That is also an approach, the down-side is that then the tests need to added in multiple places and not behaving as nicely as ``pytest``.

I do like the messages on the asserts, so I have added that to the tests.

I don't see any reason not to use ``pytest`` for this, the only counter-argument is that it is "unusual" but it is no more unusual than ``pytest`` describing the use-case in their documentation and providing test-collection for it. So, its a well-documented and supported feature of ``pytest``.

Should the addition of ``pytest`` cause us grief, then it is no worse than we can remove it and replace it with something like the approach above. However, for now I see no reason to, it only makes things simpler/easier to use.

Baekalfen: > the only counter-argument is that it is "unusual"

The other argument is that this solution is short-lived. As soon as we merge my Cython-to-Python bindings, it won't work anymore.

> Should the addition of pytest cause us grief, then it is no worse than we can remove it and replace it with something like the approach above.

If you want to, then keep and let's see.

Baekalfen (approved):

safl:
Thanks @Baekalfen !

PR #83, opened 2022-05-04
Cython xnvme bindings

Baekalfen:


toolbox/cy_xnvme_pkg/cython_loader.py
```diff
@@ -0,0 +1,68 @@
+from functools import partial
```
Baekalfen, line 1: I want to break this into a separate package (a pytest plugin)  before merging.

Baekalfen:
@safl This PR is ready to merge once the CIJoe tests get the right env vars.

python/xnvme/tests/test_cython.pyx
```diff
@@ -0,0 +1,79 @@
+from cython.operator import dereference
```
Baekalfen, line 1: Optimally, we build these at the CI. But then we are already now forced to do binary wheels. I propose we let it stay like this, and solve it with the Cython wrapper

safl (changes requested):
First off, thanks for looking into this!

Before looking at the actual introduction of Cython then a lot of stuff is provided here, this will require a couple of iterations to go over. 

I haven't looked at the ci commits, since that seems to be broken.

General form
============

I've tried to provide a description of the form at:

https://xnvme.io/docs/latest/contributing/index.html#commit-messages

Let me know if they aren't clear.

Commit Messages
---------------

For the commits in the PR, then it would be nice if the titles were changed as follows:

* "git: flake8 ignores line-lengths" => "pcf: change flake8-hook to ignore line-lengths"
  - Due to "imperative mood"

* "py: Cython bindings for xNVMe" => "py: add Cython bindings for xNVMe"
  - Due to "imperative mood" / describe that the commit "adds" something.

* "py: added Pytests for Cython bindings" => "py: add pytests for Cython bindings"
  - Due to "imperative mood"

* "ci: added pytest to GHA and CIJoe" => "ci: add pytests to GHA and CIJOE"
  - Due to "imperative mood"

* "git: update gitignore for python package"
  - Nothing, this is great.

Commit descriptions and contents
--------------------------------

The commit titled "py: Cython bindings for xNVMe" has a lot going on.
There is the addition of Cython and then there are bunch of changes, that could be commits titled as:

py:mk: fix helper 'generate-ctypes' on macOS
py:mk: improve helper 'clean'
py:mk: add helper installing build requirements (build-requirements)
py:mk: add parameterization of Python interpreter
meson: add build of Python bindings
mk: add helper cleaning build of Python bindings (clean-python)
py:mk: add dynamic versioning of Python bindings

Having them as separate commits, and possibly the fixes/additions as a PR, the addition to meson as another, and then drop the dynamic versioning of the Python bindings since it is brittle; looking at the CI then it is installing a package with version v0.0.0.

With the above a preparation patches, then it would make it much clearer to see what is involved for Cython and what are fixes, improvements and additions to the build-system.

* "py: added Pytests for Cython bindings"

It would must easier to follow what this does if the commit description described it.

E.g. why are setup.py and requirements.txt needed inside the pytest-package? It seems unusual? It might also be that I need to read up on Cython / pytest, but some pointers here would be nice.

safl:
As discussed offline, then it might make sense to separate the ctypes and cython bindings into separate distribution packages using namespace packaging. I threw an example together here: https://github.com/safl/nspkgs

safl:
Took a stab at adjusting it here: https://github.com/OpenMPDK/xNVMe/pull/105
However, seems like something is missing.

safl:
Packaging Python bindings using namespace packages looks like a path forward, see the PR: https://github.com/OpenMPDK/xNVMe/pull/105
It provides a structure for the Cython to live in, along-side the current Python package, and thus a way to separate build and distribution requirements.

I would like for the Cython bindings to go down this path. For the C-libraries, the way to get and use xNVMe is by getting the source-archive or the repository, build, and install it from source. For Python, we have ``pip``/``PyPI`` as a distribution channel.

I would like for the revival/introduction of Python to have minimal impact on the build of the C-libraries. Since, for Python we can provide the bindings via ``pip/PyPI``, then it should not introduce new dependencies and hick-ups when the user grabs and installs xNVMe, since the task of building from source should be as simple as possible, this is why the getting started guide has toolchain documentation for so many different distros, such that potential issues/hick-ups can be caught via CI and described in the documentation. Some general changes:

* Either change the default of building the Python bindings via meson to 'false' or remove the meson-integration.
* Remove to the dynamic versioning of the Python package
* Adopt the namespace packaging with package-name ``xnvme-cython``, based on https://github.com/OpenMPDK/xNVMe/pull/105 after it is reviewed and merged

Thanks!

safl:
Closing this as work has continue on: https://github.com/OpenMPDK/xNVMe/pull/121

PR #85, opened 2022-05-11
Cleanup of the xNVMe fio io-engine

safl:


vincentkfu (commented):
A comment (now moved to index.rst) mentions that touching `io_ops` in `_init()` prevents a segfault in `_queue()`. Is that problem gone now?

safl:
> 

The comment / docs are dated and need a bit more attention. However, at some point touching the pointer was needed to work-around the segfault. However, it was been quite some time since I have seen that error and currently the use of the engine on the CI without it no longer segfaults. I think I will remove the comment from the docs when going over them.

safl:
> 

Good point. Will remove that.

Issue #86, opened 2022-05-16
Unable to open device with fab URI


vkalef:
I'm working on version 0.0.29, over CentOS 7
Trying to open a device using the format fab:174.60.77.139:4420?nsid=1, but all the examples fails to get it as a valid format.

Help will be appreciated.

safl:
Hi @vkalef,
A lot has happened since v0.0.29, currently the latest integration of this is on this PR: https://github.com/OpenMPDK/xNVMe/pull/126
Can you try it, to see whether it fixes the issue?
Thanks!
Simon

karlowich:
Hi @vkalef, I'm closing this due to inactivity.
If you're still having trouble feel free to reopen, and I will be happy to help :)

vkalef:
We've managed to solve it, communication was done via discord, it just took
me a while to find out there is one.
I think you should emphasize it more on your site. It's very helpful.

karlowich:
That's great feedback!
Thanks @vkalef

PR #102, opened 2022-05-31
docs:tutorial:fabrics: adjust for multiple devices

safl:
The NVMe-over-Fabrics tutorial in the documentation only exports a single device.
This is fine when showing the various components to configure in Linux and SPDK, however, not very useful for setting up a more realistic setup consisting of an array of devices. The latter would be useful to provide an easy guide to following for troubleshooting issues such as the one reported here:  https://github.com/OpenMPDK/xNVMe/issues/86

This PR is a WIP to expand the guide for multiple devices. Here is what is missing:

- [ ] The text needs an update; it is currently describing the old scripts
- [ ] Fix the ``fabrics_target_spdk.sh`` it is currently not constructing the subsystem correctly, it is listening but when using it, either via ``nvme-cli`` or ``xnvme`` then the second device is not showing up.

### DIFF
--- a/docs/tutorial/fabrics/fabrics_env.sh
+++ b/docs/tutorial/fabrics/fabrics_env.sh
@@ -2,16 +2,18 @@
 #
 # Variables defining the fabrics setup, adjust these to match your setup
 #
-export NVMET_SUBSYS_NQN="nqn.2022-05.io.xnvme:ctrlnode1"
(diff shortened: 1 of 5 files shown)


safl:
Closing this, as it has been picked up in: https://github.com/OpenMPDK/xNVMe/pull/126

PR #105, opened 2022-06-03
py: change packaging to namespace packages

safl:
This provides multiple Python distribution packages to live in the same
Python namespace. This is done to separate the requirements for
building and distributing the xNVMe Python bindings, specifically in
preparation for the following extensions to the bindings:

* Cython API mapping
* Python bindings using the Cython API mapping
* Distribution of the entire xNVMe source archive via pip for
  build-and-install of the xNVMe libraries using Python infrastructure

The ``python/README.rst`` contains and should be updated on these goals.

This change affects alot of files, however, these are just because they
are moving into a different location. Consequently, then the
pre-commit-hook-config needs updating as well as the CI workflow
definition.

One functional change are in the Python package itself, that is changing
'setup.py' and adjusting to the namespace-usage:

        import xnvme.ctypes_mapping as xnvme

to

        import xnvme.ctypes_mapping.api as xnvme

Since the 'ctypes_mapping' is now a sub-package instead of a module.
This usage-change is why the tests and bin-examples content are also
part of this, and not just moved.

**note:** namespace-packages aren't zip-safe, that means, they can be
installed and distributed as a zip-file, no issue, however, when
installed on the system it is extracted from the archive.

**note:** pytest test-package discovery looks for ``__init__.py``, these are
not there in the top-level namespace-package, just one has to well
pytest which where to look. Thus the change:

        python3 -m pytest --pyargs xnvme

to

        python3 -m pytest --pyargs xnvme.ctypes_mapping

That sums it up.


safl:
In addition to the change introducing namespace packages, then this contains two of the minor changes from https://github.com/OpenMPDK/xNVMe/pull/83 the flake8 line-length-ignore and disabling fail-fast for build-testing.

Baekalfen:
I've made a fix so that the requirements to `setuptools` is correctly described. [py: add setuptools>=40.1 to install_requires](https://github.com/OpenMPDK/xNVMe/pull/83/commits/f1d75d04fbfdb7083c53ff6c8fa690ec39d3b5b3)

Baekalfen:
I've made a fix so that the requirements to `setuptools` is correctly described. [py: add setuptools>=40.1 to install_requires](https://github.com/OpenMPDK/xNVMe/pull/83/commits/f1d75d04fbfdb7083c53ff6c8fa690ec39d3b5b3)

Baekalfen (approved):
Looks good

python/xnvme-core/bin/xpy_libconf
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import sys
 
-import xnvme.ctypes_mapping as xnvme
+import xnvme.ctypes_mapping.api as xnvme
```
Baekalfen, line 4: I think we should get rid of the `.api`. It can easily be done through a star-import in `__init__.py` or use the code generator for it.

Baekalfen (changes requested):

safl:
> I've made a fix so that the requirements to `setuptools` is correctly described. [py: add setuptools>=40.1 to install_requires](https://github.com/OpenMPDK/xNVMe/pull/83/commits/f1d75d04fbfdb7083c53ff6c8fa690ec39d3b5b3)

Awesome, thanks! Cherry-picked it up.

python/xnvme-core/bin/xpy_libconf
safl: Yeah, I agree it would be nicer. Hmm, I did a quick star-import but it seems for some reason to have wider implications...

safl: Thanks for the offline input, will squash in the star-import.

safl: done.

PR #111, opened 2022-06-09
New libvfn-based backend

birkelund:
Add a new user-space VFIO-based backend, implemented using libvfn. This backend is largely similar to be:spdk, but the core dependency (libvfn) is, dependency-wise, easier to handle than SPDK.
    
The backend is feature complete as far as it can be, considering the current limitations of libvfn. This includes no xnvme_buf_realloc() or iovec support. xnvme_buf_vtophys() is also unimplemented (physical address are invalid/irrelevant in this backend).

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -745,10 +745,13 @@ jobs:
           "${CIJ_TESTPLANS}/xnvme-zns-zrwa-linux-spdk.plan" \
           "${CIJ_TESTPLANS}/xnvme-base-linux.plan" \
           "${CIJ_TESTPLANS}/xnvme-base-spdk.plan" \
+          "${CIJ_TESTPLANS}/xnvme-base-vfio.plan" \
           "${CIJ_TESTPLANS}/xnvme-nvm-spdk.plan" \
+          "${CIJ_TESTPLANS}/xnvme-nvm-vfio.plan" \
           "${CIJ_TESTPLANS}/xnvme-nvm-linux-nvme.plan" \
           "${CIJ_TESTPLANS}/xnvme-nvm-linux-null.plan" \
           "${CIJ_TESTPLANS}/xnvme-zns-spdk.plan" \
+          "${CIJ_TESTPLANS}/xnvme-zns-vfio.plan" \
           "${CIJ_TESTPLANS}/xnvme-zns-linux-nvme.plan" \
           "${CIJ_TESTPLANS}/xnvme-zns-linux-null.plan" \
           "${CIJ_TESTPLANS}/example_01.plan"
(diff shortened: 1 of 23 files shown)


Baekalfen (approved):
I've added the label to run CIJoe tests in GHA. If that all passes, then it looks good to me.

safl:
This is looking great, thank you! Quick question, what are the deal-breakers for libvfn on:

* Alpine
* CentOS 7
* Debian Buster
* OpenSuse leap 15.3
* Ubuntu Bionic

I ask since, on the older distro's a user-space driver is usually really nice in case one is unable to bump up the kernel.
On Alpine I am curious, since the user-land is recent but muslc based.

birkelund:
> This is looking great, thank you! Quick question, what are the deal-breakers for libvfn on:
> 
>     * Alpine

libvfn require glibc (for now).

> 
>     * CentOS 7
> 
>     * Debian Buster
> 
>     * OpenSuse leap 15.3
> 
>     * Ubuntu Bionic
> 

If i remember correctly I use features of /usr/include/vfio.h that are not present. I removed some to support more of the distros, but I'm still lacking support on these systems. Also, I think, these all complain that meson wont work "soon" due to python 3.6 going EOL ;)


> 
> I ask since, on the older distro's a user-space driver is usually really nice in case one is unable to bump up the kernel. On Alpine I am curious, since the user-land is recent but muslc based.

See above.

eli-schwartz:
> Also, I think, these all complain that meson wont work "soon" due to python 3.6 going EOL ;)

This isn't even a problem, because the currently specified minimum version of Meson for this project is a version of Meson that supports being installed on python 3.6 :) it would only be a problem for them if you start using Meson 0.62.0 as your minimum.

birkelund:
> > Also, I think, these all complain that meson wont work "soon" due to python 3.6 going EOL ;)
> 
> This isn't even a problem, because the currently specified minimum version of Meson for this project is a version of Meson that supports being installed on python 3.6 :) it would only be a problem for them if you start using Meson 0.62.0 as your minimum.

Right, thanks! I actually also managed to build on all the platforms now ;)

birkelund:
Await merge of #124.

safl:
> Await merge of #124.

#124 is in
This is awesome btw. thank you!
Why is **be:spdk: add spdk prefix to cref functions** needed? The functions should be local to the translation unit and not leak out any symbol conflicts, or perhaps they do? If it is not strictly needed, could we drop it yo avoid clashing with https://github.com/OpenMPDK/xNVMe/pull/120

birkelund:
> Why is **be:spdk: add spdk prefix to cref functions** needed? The functions should be local to the translation unit and not leak out any symbol conflicts, or perhaps they do? If it is not strictly needed, could we drop it yo avoid clashing with #120

This was just a "cleanup" patch (I think) that I carried over from Michael's original be-vfio poc. I think it can be safely dropped. 

Pushed without.

birkelund:
> > This is looking great, thank you! Quick question, what are the deal-breakers for libvfn on:
> > ```
> > * Alpine
> > ```
> 
> libvfn require glibc (for now).
> 
> > ```
> > * CentOS 7
> > 
> > * Debian Buster
> > 
> > * OpenSuse leap 15.3
> > 
> > * Ubuntu Bionic
> > ```
> 

By the way, it builds on all of the above now.

safl:
Thank you!

PR #112, opened 2022-06-10
api: drop sgls from API

safl:
The level in which xNVMe encapsulates payloads are contigous buffers and
iovecs. As these are "layouts" commonly used for storage in user-facing
libraries, that is, where the user is another library or an application
running in user space.

The SGL representation and helper functions to create sgl-lists are
currently only usable with the SPDK/NVMe driver backend. This breaks the
goal of a unified API, ideally xNVMe should be providing contig. buffers
and iovecs and leave the PRP and SGL construction either to the backend,
or to the driver. Thus, this change to remove them.

Should it be requested, then SGLs could be re-introduced, along with
proper testing of their functionality and examples of their use.
Ideally, re-introduced in a manner compatible with other user-space NVMe
drivers, as the current implementation is specific to the SPDK/NVMe
driver.


karlowich (approved):
Looks good!

python/xnvme-cy-header/aux/generate_cython_header.py
```diff
@@ -55,7 +55,6 @@ def generate_pxd(workspace_root):
         "libxnvme_znd.h",
         "libxnvme_pp.h",
         "libxnvme_be.h",
-        "libxnvme_sgl.h",
```
ankit-sam, line 58: There are traces of xnvme_sgl here (Line 9, 10). they can be removed

ankit-sam (commented):
Changes look good.

Small nit - few spelling mistakes in commit message

- librariy
- contigous

safl:
Thanks @ankit-sam , fixed!

python/xnvme-cy-header/aux/generate_cython_header.py
safl: Overlooked this, fixed now. Also, found another mention of the SGLs in the cython header. Removed that as well.

Baekalfen: Nice, to get rid of some sgl-specific fixes as well.

ankit-sam (approved):

karlowich (approved):

PR #113, opened 2022-06-10
Debian stretch EOL

Baekalfen:
Debian 9 "Stretch" will no longer be maintained after June 30, 2022,
therefore we will also no longer support it either.


birkelund (approved):

safl (approved):
LGTM!

Issue #109, opened 2022-06-09
Opening new namespace when device already open fails with SPDK backend


LiadOz:
We have started a discussion on this issue on Discord.
The issue happens on version 0.3.0 and I've confirmed it happens on centos 7 and ubuntu 20.

Reproduction steps:
setup nvmf app, create 2 devices and add only one to the subsystem
```
./build/bin/nvmf_tgt &
./scripts/rpc.py nvmf_create_transport -t TCP -u 16384 -m 8 -c 8192
./scripts/rpc.py nvmf_create_subsystem nqn.2016-06.io.spdk:cnode1 -a -s SPDK00000000000001 -d SPDK_Controller1
./scripts/rpc.py nvmf_subsystem_add_listener nqn.2016-06.io.spdk:cnode1 -t tcp -a  127.0.0.1 -s 4420
./scripts/rpc.py bdev_null_create Null0 8589934592 4096
./scripts/rpc.py bdev_null_create Null1 8589934592 4096
./scripts/rpc.py nvmf_subsystem_add_ns nqn.2016-06.io.spdk:cnode1 Null0
```

Run the following code (note the sleep)
```
#include <libxnvme.h>
#include <libxnvme_pp.h>
#include <unistd.h>

int
main(int argc, char *argv[])
{
        struct xnvme_opts opts1 = xnvme_opts_default();
        struct xnvme_opts opts2 = xnvme_opts_default();
        struct xnvme_dev *dev1;
        struct xnvme_dev *dev2;

        opts1.nsid = 1;
        opts2.nsid = 2;
        dev1 = xnvme_dev_open("127.0.0.1:4420", &opts1);
        sleep(20);
        dev2 = xnvme_dev_open("127.0.0.1:4420", &opts2);
        xnvme_dev_close(dev1);
        xnvme_dev_close(dev2);

        return 0;
}
```

When it starts sleeping add the other namespace
```
sudo ./scripts/rpc.py nvmf_subsystem_add_ns nqn.2016-06.io.spdk:cnode1 Null1
```

Here are the logs:
```
# DBG:xnvme_be.c:xnvme_be_factory-667: INFO: obtained backend instance be: 'spdk'
# DBG:xnvme_be_spdk_dev.c:_spdk_env_init-230: INFO: SPDK NVMe PCIe Driver registration -- BEGIN
# DBG:xnvme_be_spdk_dev.c:_spdk_env_init-235: INFO: skipping, already registered.
[...]

Issue #107, opened 2022-06-09
SPDK backend uses detached controller


LiadOz:
I have found the following issue when working with version 0.3.0 and using ip:port for opening devices, here is the flow:
1. Create 5 namespaces (should happen with at least 2)
2. Open namespaces and issue commands
3. Close namespaces
4. Create namespace, open and issue commands
In step 4 I crash sporadically, when debugging the issue I noticed that the controller has garbage values at this stage.

This is what I see when debugging:
1. _cref_insert is called 5 times with the same controller, creating 5 entries
2. _cref_lookup is called 5 times
3. when closing the device _cref_deref is called 5 times removing the first entry of the controller out of 5 and detaching the controller
4. _cref_lookup is called and raises the refcount of the next entry (same controller) to 2

It seems to me that the root cause of the issue is creating the 5 entries instead of only 1. assuming _cref_deref would be called enough times to clear every entry in the table it will detach the controller 5 times which should not happen.

With these changes the I stop crashing
```
diff --git a/lib/xnvme_be_spdk_dev.c b/lib/xnvme_be_spdk_dev.c
index 09f2ffc..91bd247 100644
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -69,6 +69,15 @@ _cref_insert(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
                XNVME_DEBUG("FAILED: !ctrlr");
                return -EINVAL;
        }
+       for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) { // check if entry already exists
+                if (!g_cref[i].refcount) {
+                        continue;
+                }
+
+                if (strncmp(g_cref[i].uri, ident->uri, XNVME_IDENT_URI_LEN - 1)) {
+                        return 0;
+                }
+        }

        for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
                if (g_cref[i].refcount) {
```

safl:
Hi @LiadOz,
Thanks for reporting this and diving into the issue. For the five namespaces, the creation of those, is that similar to the scripts here:
* spdk nvmf app: https://github.com/OpenMPDK/xNVMe/pull/102/commits/28cb2296de4e7234feb98aedceb61dace0926e6e#diff-91c733ec4807dc30e5216883e2d73362bce00475cc9b06a6879d15a82636dfc6
[...]

PR #114, opened 2022-06-12
be:spdk: process spdk events before reusing controller

LiadOz:
Signed-off-by: Liad Oz <liadozil@gmail.com>

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -762,6 +762,16 @@ xnvme_be_spdk_state_init(struct xnvme_dev *dev)
 
 		XNVME_DEBUG("INFO: found dev->ident.uri: '%s' via cref_lookup()", dev->ident.uri);
 
+		err = spdk_nvme_ctrlr_process_admin_completions(state->ctrlr);
+		if (err < 0) {
+			XNVME_DEBUG("FAILED: spdk_nvme_ctrlr_process_admin_completions, err: %d",
+				    err);
+			if (_cref_deref(state->ctrlr, true)) {
+				XNVME_DEBUG("FAILED: _cref_deref");
+			}
+			goto probe;
+		}
+
 		ns = spdk_nvme_ctrlr_get_ns(state->ctrlr, dev->opts.nsid);
 		if (!ns) {
 			XNVME_DEBUG("FAILED: spdk_nvme_ctrlr_get_ns(0x%x)", dev->opts.nsid);


LiadOz:
resolves #109

safl:
Hi @LiadOz,
LGTM, CI is spinning up to check for any regressions.
Thanks,
Simon

safl:
Looks like the test ``xnvme_tests_enum open --count 4`` fails with this change.
The test is here ``tests/enum.c``. In the failing scenario it attempts to open the same PCIe address multiple times, this is the use-case the ``cref`` was made for, as we cannot have multiple driver instances, we have to share and re-use the one we have previously attached to. I am currently traveling, so can't dive into the code, however @birkelund and @Baekalfen did you see a similar thing / issues when references counting in https://github.com/OpenMPDK/xNVMe/pull/111 ? Something that might be applicable here?

birkelund:
> Looks like the test `xnvme_tests_enum open --count 4` fails with this change. The test is here `tests/enum.c`. In the failing scenario it attempts to open the same PCIe address multiple times, this is the use-case the `cref` was made for, as we cannot have multiple driver instances, we have to share and re-use the one we have previously attached to. I am currently traveling, so can't dive into the code, however @birkelund and @Baekalfen did you see a similar thing / issues when references counting in #111 ? Something that might be applicable here?

Test is OK for the be-vfio backend.

LiadOz:
Is there a way to see the logs for this test? I suspect the controller might be detached when using it similar to https://github.com/OpenMPDK/xNVMe/issues/107, before this change assuming the controller was detached it gets to spdk_nvme_ctrlr_get_ns which returns 0 and then creates a new controller.

lib/xnvme_be_spdk_dev.c
```diff
@@ -763,6 +763,16 @@ xnvme_be_spdk_state_init(struct xnvme_dev *dev)
 
 		XNVME_DEBUG("INFO: found dev->ident.uri: '%s' via cref_lookup()", dev->ident.uri);
 
+		err = spdk_nvme_ctrlr_process_admin_completions(state->ctrlr);
+		if (err < 0) {
+			XNVME_DEBUG("FAILED: spdk_nvme_ctrlr_process_admin_completions, err: %d",
+				    err);
+			if (_cref_deref(state->ctrlr)) {
```
karlowich, line 770: @safl @LiadOz This line needs to be changed to:
`if (_cref_deref(state->ctrlr, true)) {`

karlowich (approved):
LGTM!

safl (approved):
Great, thanks!

PR #119, opened 2022-06-22
Changes to subprojects/liburing build

safl:
When taking a stab at RPM-packaging (https://github.com/OpenMPDK/xNVMe/tree/packaging/rpm) xNVMe it seemed better to change the way the liburing subproject is build by:

* Applying patches to enable -fPIC rather than ``make CFLAGS=...``  override
* Adding a patch to change the definition of ``memfd_create`` when it is missing

The latter patch should be reconsidered as it might be an issue with the mock-env used when doing ``fedpkg ..`` rather than a general issue, adding to that, in case of a general issue, then it should be submitted as a patch to upstream liburing.

### DIFF
--- a/subprojects/packagefiles/liburing/liburing_build.sh
+++ b/subprojects/packagefiles/liburing/liburing_build.sh
@@ -1,2 +1,0 @@
-#!/usr/bin/env sh
-make -j CFLAGS="${CFLAGS} -fPIC"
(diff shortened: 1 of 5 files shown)


safl:
This has been replaced by linking with liburing instead of bundling it, see:

- https://github.com/OpenMPDK/xNVMe/pull/162

PR #120, opened 2022-06-23
be:spdk: fix cref logic

### LINKED ISSUE #107: SPDK backend uses detached controller

LiadOz:
fixes #107

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -85,7 +85,7 @@ _cref_insert(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
 }
 
 int
-_cref_deref(struct spdk_nvme_ctrlr *ctrlr)
+_cref_deref(struct spdk_nvme_ctrlr *ctrlr, bool detach_on_no_refcount)
 {
 	if (!ctrlr) {
 		XNVME_DEBUG("FAILED: !ctrlr");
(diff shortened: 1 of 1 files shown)


LiadOz:
it failed with the error, meson: command not found
@safl is that an issue with the host?

safl:
Yes, re-running it now.

safl:
HI @LiadOz ,
This has been integrated on https://github.com/OpenMPDK/xNVMe/pull/110 and merged into ``next``.
Thanks!
Simon

PR #121, opened 2022-06-23
Continuation of Cython header and bindings

safl:
TODO
=====

- [x] Fix up the different README.rst files to reflect their package and refer to the others.
- [x] Check if there is a way to reduce the ``build-python`` CI job
- [x] Refactor to reflect the different interface-offerings
      * ``xnvme.ctypes_mapping`` to ``xnvme.ctypes_bindings``
      * ``xnvme.cython_mapping`` to ``xnvme.cython_header``
      * ``xnvme.bindings`` to ``xnvme.cython_bindings``

Toolbox packages changes
========================

* Dropped the SSL-patching of the Python build on CentOS since it is not needed
* Dropped the addition of "toolbox/pkgs/debian-bookworm-build.sh" since it is identical to the "default-build.sh"
* Dropped the changes to Windows since they are not used for anything
* Dropped the duplicate additions of 'pkg-config' and 'openssl-devel'
* Sorted the contents of toolbox/pkgs/*.txt, might make it easier to catch duplicate entries..

Additionally:

* Remove astyle as it is deprecated

* Changed the commit-message to describe what the package-additions are for
  - python-venv, needed as a means to improve the xNVMe Python packages
  - python3-devel, this is needed for building the Cython bindings

  - why is git needed on Gentoo?

Adding the cython-header to its own namespace package
=====================================================

The addition of the cython-header to the core package does not make it easier
to understand the separation of namespace-packages, nor does it reap the
benefits. It could be added to the bindings-package, however, then we would
loose the benefits of a simple sdist without binary depends. For those who just
want to consume xNVMe from their Cython code, all they would need is that
header. Thus, adding the cython-header to its own namespace-package.

Undocumented
============

* The 'xnvme-core' namespace package must be named "xnvme-core..." not
  "xnvme-py", the "xnvme-core..." .tar.gz name is used in the CI.

Misc fixes
==========

* Removed the "clean-python" from main Makefile.

* The changes to the CI are really hard to follow, re-organizing them

* The xnvme-core/ctypes_mapping/tests/__init__.py were missing, losing the
ability to run the pytests.

* The TOOLBOX_DIR was wrong in Makefiles, breaking 'make help'

Commit: "py: add support for FILE* in Python bindings"

  * Was missing sign-off

Commit: "py: work around spdk enumerate bug"

  * Was missing sign-off
  * Does not say anything else, like why, what?

Commit: "py: update test_write_zeroes pytest"

  * Was missing sign-off and had no description

Commit: "py: prepare xnvme-core makefile for xnvme-cython"

  * Was missing sign-off and had no description


python/xnvme-cy-bindings/setup.py
```diff
@@ -0,0 +1,178 @@
...
+try:
+    os.remove(
+        xnvme_init_py_path
+    )  # Don't misinterpret the local directory as the xnvme package
+except FileNotFoundError:
+    pass
+import xnvme.cython_header  # noqa: E402
+
+PACKAGE_NAME = "xnvme"
+cy_xnvme_pkg_path = os.path.dirname(os.path.realpath(__file__))
```
jwdevantier, line 29: Would prefer `Path(__file__).parent` to using `os` package

python/xnvme-cy-bindings/setup.py
```diff
@@ -0,0 +1,178 @@
...
+WORKSPACE_ROOT = cy_xnvme_pkg_path + "/../.."
```
jwdevantier, line 30: Given above, this becomes `cy_xnvme_pkg_path.parent.parent`  -- avoids os-specific path-mangling.

python/README.rst
```diff
@@ -11,8 +11,14 @@ the **xNVMe** Python bindings then they are packaged as `namespace packages`_.
   - Also, if needed then this package can ship plain-Python utilities
   - Intented to be distributed as a Python source-package
 
-* xnvme-cython
-  - distribution package name: ``xnvme-cython``
+* xnvme-cy-header
+  - distribution package name: ``xnvme-cy-header``
+  - Provides a Cython (``libxnvme.pxd``) header, making the xNVMe C API
+    available Cython
```
jwdevantier, line 17: available to* Cython

python/README.rst
```diff
@@ -11,8 +11,14 @@ the **xNVMe** Python bindings then they are packaged as `namespace packages`_.
...
+  - Intented to be distributed as a Python source-package
```
jwdevantier, line 18: Intended*

safl:
Thanks for the input @jwdevantier , fixed the typos and added notes on things to improve going forward here: https://github.com/OpenMPDK/xNVMe/issues/123

PR #125, opened 2022-07-04
Add subnqn to device-identifier

safl:
When enumerating e.g. ``xnvme enum --uri <fabrics>``, without a subnqn then the list of devices have duplicated entries as a remote endpoint can have multiple suibsystems, there is no telling them apart. Thus, adding the ``subnqn`` to identifiers.
This is related to: https://github.com/OpenMPDK/xNVMe/pull/110


safl:
Closing this, as it has been picked up in https://github.com/OpenMPDK/xNVMe/pull/126

PR #132, opened 2022-07-20
tests: add ioworker test

karlowich:
Add ioworker read write test to relevant backends.

This is in preparation for the ramdisk enhancements, where the ioworker test is likely to be expanded.


safl (changes requested):
This looks good. I little clean/squash of the history would be nice, but really nice to see the expansion of the ioworker and testcases using it. Great!

safl (approved):

PR #133, opened 2022-07-20
tests: add flag for be to enum

### LINKED ISSUE #129: xnvme_tests_enum is missing flag for backend
The test `xnvme_tests_enum` is missing a flag to specify the desired backend.
Without this it will only test the SPDK backend. 
Since it is also part of the vfio testplan, it should be expanded to actually test the correct backend.

karlowich:
Make it possible to run `xnvme_tests_enum` with different backends
Closes #129


safl (approved):
LGTM, thanks!

PR #135, opened 2022-08-01
be:spdk: Add timeout to commands

LiadOz:


karlowich:
Hi @LiadOz, 
This is looking good.
SPDK v22.09 is now used in `next`, so you can remove the out-of-tree patches.
Then if the tests are passing, I think this should be merged. 
Thanks!

LiadOz:
I have pushed the latest patch we use on our fork, we put the timeout time to be 10s to correspond to the kernel timeout. note that on any command timeout this commit fails the controller, please verify thoroughly that this behaviour is appropriate.

LiadOz:
the timeout in the kernel for IO commands is actually 30 seconds and the admin timeout is 60 seconds as can be seen [here](https://elixir.bootlin.com/linux/v5.19.12/source/drivers/nvme/host/core.c#L38:~:text=unsigned%20int%20admin_timeout%20%3D%2060%3B%0Amodule_param(admin_timeout%2C%20uint%2C%200644)%3B%0AMODULE_PARM_DESC(admin_timeout%2C%20%22timeout%20in%20seconds%20for%20admin%20commands%22)%3B%0AEXPORT_SYMBOL_GPL(admin_timeout)%3B%0A%0Aunsigned%20int%20nvme_io_timeout%20%3D%2030%3B)

karlowich:
Hi @LiadOz, I'm working on expanding the xNVMe api to allow registering callback functions.
That way the user can decide what to do on a timeout etc. and we don't need to make assumptions.

LiadOz:
@karlowich sounds great

karlowich (changes requested):
@LiadOz Since #216 doesn't cover every case, let's go with this solution instead.
I think it is a good idea to modify it to use the timeout values from the kernel as you mentioned.

karlowich (approved):

karlowich:
@safl In my opinion, this is ready to be merged.
Thanks @LiadOz!

safl (approved):
Thanks @LiadOz and @karlowich !

PR #152, opened 2022-09-13
opts: change default nsid to 1

### LINKED ISSUE #131: xnvme info should indicate which arguments are required
When running `xnvme info` in a fabrics setup the required arguments are uri, dev-nsid and subnqn (if there are more subsystems).
However, this is not stated anywhere and the error from a missing argument is "ERR: 'xnvme_dev_open()': {errno: -2, msg: 'No such file or directory'}", which is not very helpful.

Ideally both the error message and the `xnvme info --h` should indicate which arguments are required.

### LINKED ISSUE #154: Clean up flow of xnvme_opts when using xnvmec
`xnvme_opts_default` gets overwritten when calling commands from cli. 
That is, when calling a xnvme command from the cli, fx. `xnvme info`, the `opts` get overwritten by `xnvmec_cli_to_opts` in `xnvmec`.
Looking at the code in `xnvmec.c` (line ~1840):
```
// opts are created
// opts are overwritten
if (xnvmec_cli_to_opts(cli, &opts)) {
	xnvmec_perr("xnvmec_cli_to_opts()", errno);
	return -1;
}
```
Furthermore, when xnvme_dev_open is called the `opts->rdwr` is reset to the default that was overwritten.
Overall it is a bit messy and not working as intended.

A solution could be to make `xnvme_opts_default` take `opts` as an argument and have it set the unset values to the default.
Then it could be called after the `xnvmec_cli_to_opts`.

karlowich:
Currently, spdk fails if nsid is not provided.
`--dev-nsid` is optional for `xnvme info`, however if unset it will fail with the userspace driver.
This results in a poor user experience, since the error message is "Input/output error"

closes #131 
(This solution seemed more elegant than making `--dev-nsid` a mandatory argument to `xnvme info`)

This includes #155 and closes #154

### DIFF
--- a/include/libxnvmec.h
+++ b/include/libxnvmec.h
@@ -524,8 +524,7 @@ xnvmec(struct xnvmec *cli, int argc, char **argv, int opts);
  * @param cli The command-line-interface to parse arguments for
  * @param opts The device-options to fill
  *
- * @return On success, the 0 is returned. On error, -1 is returned and `errno` set to indicate the
- * error.
+ * @return On success, 0 is returned.
  */
 int
 xnvmec_cli_to_opts(const struct xnvmec *cli, struct xnvme_opts *opts);
--- a/lib/xnvme_opts.c
+++ b/lib/xnvme_opts.c
@@ -11,6 +11,7 @@ xnvme_opts_default(void)
 	struct xnvme_opts opts = {0};
 
 	opts.rdwr = 1;
+	opts.nsid = 1;
 
 	// Value is only applicable if the user also sets opts.create = 1
 	opts.create_mode = S_IRUSR | S_IWUSR;
--- a/lib/xnvmec.c
+++ b/lib/xnvmec.c
@@ -1872,42 +1872,44 @@ xnvmec(struct xnvmec *cli, int argc, char **argv, int opts)
 int
 xnvmec_cli_to_opts(const struct xnvmec *cli, struct xnvme_opts *opts)
 {
-	opts->be = cli->args.be;
-	// opts->dev = cli->args.dev;
-	opts->mem = cli->args.mem;
-	opts->sync = cli->args.sync;
-	opts->async = cli->args.async;
-	opts->admin = cli->args.admin;
+	opts->be = cli->given[XNVMEC_OPT_BE] ? cli->args.be : opts->be;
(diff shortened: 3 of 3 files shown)


safl:
The ``libvfn`` backend could benefit for this as well. So, how about adding this to ``xnvme_opts_default()`` instead?

karlowich:
The reason I wanted to keep it "local" to spdk was because of [this](https://github.com/OpenMPDK/xNVMe/blob/66a0458c5481e1180ad9938647500bd837988545/lib/xnvme_be_macos_admin.c#L63)

It seems that maybe nsid has a different meaning depending on the backend.
However, `nsid == 1` seems like a sensible default

safl:
> The reason I wanted to keep it "local" to spdk was because of [this](https://github.com/OpenMPDK/xNVMe/blob/66a0458c5481e1180ad9938647500bd837988545/lib/xnvme_be_macos_admin.c#L63)
> 
> It seems that maybe nsid has a different meaning depending on the backend. However, `nsid == 1` seems like a sensible default

That is just a macos-specific handling of the interface provided to do what it wants to do. ``nsid`` should in the public API mean what is defined in the spec. so that is fine.

karlowich:
@safl This does not work as of right now due to the issue described here: #154

safl:
This is looking great, two minor things. For the docs, since the ``--dev-nsid=1`` is removed from the examples then a small ``note:: ...`` should be added to the docs describing how to control which namespace to use, as this is useful for user-space drivers. The other thing, ``xnvmec:`` is the label usually used as prefix in the commit messages, so it would great to re-word the first commit and use a "()" on the function name, and be slightly more specific. ``cli: clean up xnvmec_cli_to_opts`` --> ``xnvmec: fix xnvmec_cli_to_opts() by only overwriting given arguments``. Or some better wording...

safl (changes requested):
This is looking great, two minor things. For the docs, since the ``--dev-nsid=1`` is removed from the examples then a small ``note:: ...`` should be added to the docs describing how to control which namespace to use, as this is useful for user-space drivers. The other thing, ``xnvmec:`` is the label usually used as prefix in the commit messages, so it would great to re-word the first commit and use a "()" on the function name, and be slightly more specific. ``cli: clean up xnvmec_cli_to_opts`` --> ``xnvmec: fix xnvmec_cli_to_opts() by only overwriting given arguments``. Or some better wording...

safl (approved):

PR #155, opened 2022-09-15
cli: clean up xnvmec_cli_to_opts

### LINKED ISSUE #154: Clean up flow of xnvme_opts when using xnvmec

karlowich:
This fixes an issue where xnvme_opts_default gets overwritten when calling commands from cli (closes #154)

### DIFF
--- a/lib/xnvmec.c
+++ b/lib/xnvmec.c
@@ -1872,42 +1872,44 @@ xnvmec(struct xnvmec *cli, int argc, char **argv, int opts)
 int
 xnvmec_cli_to_opts(const struct xnvmec *cli, struct xnvme_opts *opts)
 {
-	opts->be = cli->args.be;
-	// opts->dev = cli->args.dev;
-	opts->mem = cli->args.mem;
-	opts->sync = cli->args.sync;
-	opts->async = cli->args.async;
-	opts->admin = cli->args.admin;
+	opts->be = cli->given[XNVMEC_OPT_BE] ? cli->args.be : opts->be;
+	opts->mem = cli->given[XNVMEC_OPT_MEM] ? cli->args.mem : opts->mem;
+	opts->sync = cli->given[XNVMEC_OPT_SYNC] ? cli->args.sync : opts->sync;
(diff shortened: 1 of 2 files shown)


safl (commented):
When the user provides opts, then those are the opts to use. This PR seems to always set opts to default?

lib/xnvmec.c
```diff
@@ -1837,7 +1837,7 @@ xnvmec(struct xnvmec *cli, int argc, char **argv, int opts)
 	}
 
 	if ((opts & XNVMEC_INIT_DEV_OPEN) && cli->args.uri) {
-		struct xnvme_opts opts = xnvme_opts_default();
```
safl, line 1840: Why change this? It provided defaults, which would be overwritten by cli-arguments.

karlowich: The problem is that they are overwritten even if the user provides no arguments. All the int values are set to 0

safl: I see. ``xnvmec_cli_to_opts`` needs a bit a attention, something like changing:

``opts->be = cli->args.be;``
to
``opts->be = cli->args.given[XNVMEC_OPT_BE] ? cli->args.be : opts->be;``

karlowich:
Included in #152

PR #158, opened 2022-09-19
CI: add scan-build to workflow

karlowich:
This is an experiment.
`scan-build` detects errors and warnings that I think is reasonable and should be addressed to improve code quality.
This PR will include fixes for the errors found and based on that we can make a decision of whether to make it a part of our CI.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -94,6 +94,41 @@ jobs:
       run: pre-commit run --all-files
 
   #
+  # Code-analysis using scan-build
+  #
+  analyze-scan-build:
+    needs: source-archive
+    runs-on: ubuntu-latest
+    container: refenv/xnvme-devtools-ci:latest
+
+    steps:
+    - name: Retrieve the source-archive
+      uses: actions/download-artifact@v2
+      with:
+        name: archive-src
+    - name: Extract the full-source-archive
+      run: |
+        tar xzf xnvme*.tar.gz --strip 1
+        rm xnvme*.tar.gz
+
+    - name: install scan-build (clang-tools)
+      run: apt-get install -qy clang-tools
+
+    - name: Run config-debug
+      run: scan-build make config-debug
+
+    - name: Run scan-build
+      run: scan-build --exclude subprojects -o /tmp/scan-build-report make
+
+    - name: Upload scan-build report
+      uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        name: scan-build-report
+        path: /tmp/scan-build-report/*
+        if-no-files-found: error
+
+  #
   # Code-analysis using GitHUB CodeQL
   #
   analyze-codeql:
--- a/lib/xnvme_be_linux_block.c
+++ b/lib/xnvme_be_linux_block.c
@@ -74,7 +74,7 @@ int
 xnvme_be_linux_sysfs_dev_attr_to_num(struct xnvme_dev *dev, const char *attr, uint64_t *num)
 {
 	const int buf_len = 0x1000;
-	char buf[buf_len];
+	char buf[0x1000] = {'\0'};
 	int base = 10;
 	int err;
 
--- a/lib/xnvme_be_posix_async_thrpool.c
+++ b/lib/xnvme_be_posix_async_thrpool.c
@@ -203,7 +203,7 @@ _posix_async_thrpool_term(struct xnvme_queue *q)
 		XNVME_DEBUG("FAILED: pthread_mutex_unlock(), err_lock: %d", err_lock);
 	}
 
-	for (int i = 0; i < queue->nthreads; i++) {
+	for (int i = 0; queue->threads && i < queue->nthreads; i++) {
 		pthread_join(queue->threads[i], NULL);
 	}
 	free(queue->threads);
--- a/tests/lblk.c
+++ b/tests/lblk.c
@@ -59,7 +59,8 @@ boilerplate(struct xnvmec *cli, uint8_t **wbuf, uint8_t **rbuf, size_t *buf_nbyt
 	if (!*rbuf) {
 		err = -ENOMEM;
 		xnvmec_perr("xnvme_buf_alloc()", err);
-		xnvme_buf_free(dev, *rbuf);
+		xnvme_buf_free(dev, *wbuf);
+		*wbuf = NULL;
 		return err;
 	}
 
--- a/tests/lblk.c
+++ b/tests/lblk.c
@@ -50,14 +50,14 @@ boilerplate(struct xnvmec *cli, uint8_t **wbuf, uint8_t **rbuf, size_t *buf_nbyt
 
 	*wbuf = xnvme_buf_alloc(dev, *buf_nbytes);
 	if (!*wbuf) {
-		err = -errno;
+		err = -ENOMEM;
 		xnvmec_perr("xnvme_buf_alloc()", err);
 		return err;
 	}
 
 	*rbuf = xnvme_buf_alloc(dev, *buf_nbytes);
 	if (!*rbuf) {
-		err = -errno;
+		err = -ENOMEM;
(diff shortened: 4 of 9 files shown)


safl:
This is an great initiative @karlowich , thanks!

tests/lblk.c
```diff
@@ -59,7 +59,8 @@ boilerplate(struct xnvmec *cli, uint8_t **wbuf, uint8_t **rbuf, size_t *buf_nbyt
 	if (!*rbuf) {
 		err = -ENOMEM;
 		xnvmec_perr("xnvme_buf_alloc()", err);
-		xnvme_buf_free(dev, *rbuf);
+		xnvme_buf_free(dev, *wbuf);
```
safl, line 62: Nice catch.

lib/xnvme_be_posix_async_aio.c
```diff
@@ -81,7 +81,7 @@ _posix_async_aio_poke(struct xnvme_queue *q, uint32_t max)
 	req = TAILQ_FIRST(&queue->reqs_outstanding);
 	assert(req != NULL);
 
-	while (req != NULL && completed < max) {
+	while (req != NULL && req->ctx != NULL && completed < max) {
```
safl, line 84: scan-build is a bit annoying here since this should never happen as entries entering ``reqs_outstanding`` always have a ``ctx``.  Hmm...

lib/xnvme_be_vfio_async.c
```diff
@@ -76,6 +76,9 @@ xnvme_be_vfio_queue_poke(struct xnvme_queue *queue, uint32_t max)
 		rq = __nvme_rq_from_cqe(q->sq, cqe);
 
 		ctx = (struct xnvme_cmd_ctx *)rq->opaque;
+		if (!ctx) {
```
safl, line 79: This one I am not fond of either, similar to the aio-controlblock.
Something to mull on..

lib/xnvme_be_posix_async_thrpool.c
```diff
@@ -203,10 +203,12 @@ _posix_async_thrpool_term(struct xnvme_queue *q)
 		XNVME_DEBUG("FAILED: pthread_mutex_unlock(), err_lock: %d", err_lock);
 	}
 
-	for (int i = 0; i < queue->nthreads; i++) {
-		pthread_join(queue->threads[i], NULL);
+	if (queue->threads) {
+		for (int i = 0; i < queue->nthreads; i++) {
+			pthread_join(queue->threads[i], NULL);
+		}
+		free(queue->threads);
```
safl, line 210: Guarding the ``free()`` should not be needed.

lib/xnvme_be_posix_async_thrpool.c
```diff
@@ -203,10 +203,12 @@ _posix_async_thrpool_term(struct xnvme_queue *q)
 		XNVME_DEBUG("FAILED: pthread_mutex_unlock(), err_lock: %d", err_lock);
 	}
 
-	for (int i = 0; i < queue->nthreads; i++) {
-		pthread_join(queue->threads[i], NULL);
+	if (queue->threads) {
+		for (int i = 0; i < queue->nthreads; i++) {
```
safl, line 207: As i recall, then ``queue->nthreads`` is ``0`` in case of error.
But I guess scan-build does not see that relation.

How about for "compactness" changing this to the following instead:
```
for (int i = 0; queue->threads && (i < queue->nthreads); i++) {
   thread_join(queue->threads[i], NULL);
}
```

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
```diff
@@ -131,6 +131,9 @@ def test_log(cijoe, device, be_opts, cli_args):
 @xnvme_parametrize(labels=["dev"], opts=["be", "admin"])
 def test_feature_get(cijoe, device, be_opts, cli_args):
 
+    if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
```
safl, line 134: great.

safl (commented):
This is excellent. However, there are two places where I am not super happy with the thing scan-build is considering an issue. I might of course be wrong.

Yet, is there a way to mark something as a false-positive? Some argument to the scan-build run?

.github/workflows/verify.yml
```diff
@@ -93,6 +93,36 @@ jobs:
     - name: Run pre-commit
       run: pre-commit run --all-files
 
+  #
+  # Check source-format using pre-commit
```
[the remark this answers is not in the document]
karlowich: Done

.github/workflows/verify.yml
```diff
@@ -93,6 +93,36 @@ jobs:
...
+  #
+  scan-build-check:
```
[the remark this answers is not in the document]
karlowich: Done

.github/workflows/verify.yml
```diff
@@ -93,6 +93,36 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: refenv/xnvme-devtools-ci:latest
+
```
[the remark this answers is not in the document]
karlowich: Done

lib/xnvme_be_linux_block.c
```diff
@@ -74,7 +74,7 @@ int
 xnvme_be_linux_sysfs_dev_attr_to_num(struct xnvme_dev *dev, const char *attr, uint64_t *num)
 {
 	const int buf_len = 0x1000;
-	char buf[buf_len];
+	char buf[0x1000] = {'\0'};
```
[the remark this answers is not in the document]
karlowich: Done

lib/xnvme_be_posix_async_thrpool.c
karlowich: Done

lib/xnvme_be_posix_async_thrpool.c
karlowich: Done

karlowich:
> Is there a way to mark something as a false-positive?

Unfortunately, I don't think there is. The best we can do is the `--exclude` but that exclude an entire file (or directory)

tests/lblk.c
```diff
@@ -50,14 +50,14 @@ boilerplate(struct xnvmec *cli, uint8_t **wbuf, uint8_t **rbuf, size_t *buf_nbyt
 
 	*wbuf = xnvme_buf_alloc(dev, *buf_nbytes);
 	if (!*wbuf) {
-		err = -errno;
+		err = -ENOMEM;
```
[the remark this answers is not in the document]
karlowich: No it is because scan-build assumes that errno could be 0 here, which would cause problems when we later try to use wbuf or rbuf.
Of course, we know it could never be 0 if the condition is true.

safl (changes requested):
Looks great, to move this along then:

- [ ] Drop the two "questionable" changes
- [ ] Add ``|| true`` to the ``scan-build`` invocation, thereby allow scan-build to find errors.

.github/workflows/verify.yml
```diff
@@ -93,6 +93,41 @@ jobs:
...
+        rm xnvme*.tar.gz
+
+    - name: install scan-build (clang-tools)
+      run: apt-get install -qy clang-tools
+
+    - name: Run config-debug
+      run: scan-build make config-debug
+
+    - name: Run scan-build
+      run: scan-build --exclude subprojects -o /tmp/scan-build-report --status-bugs make
```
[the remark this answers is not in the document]
karlowich: I just removed the `--status-bugs` which is the flag causing it to return an err if bugs are found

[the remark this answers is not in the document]
safl: great!

safl (approved):

PR #159, opened 2022-09-21
Update fio subprojects to v3.32

safl:
Bumping up fio such that we can compare built-in ``io_uring_cmd`` with xNVMe backend using ``io_uring_cmd``. The built of fio has ``--disable-xnvme`` as from the xNVMe side then the external engine is still in use until we get the Windows issue resolved.


Baekalfen (approved):

ankit-sam (commented):
I am bit late, but looks good

PR #161, opened 2022-09-21
meson: configure spdk/dpdk to build on generic arm64 (Apple Silicon)

Baekalfen:

### DIFF
--- a/subprojects/packagefiles/spdk/meson.build
+++ b/subprojects/packagefiles/spdk/meson.build
@@ -20,6 +20,15 @@ if get_option('build_subprojects') and not fs.exists('build' / 'lib' / 'src' / '
     check: true,
   ).stdout().strip()
 
+  message('Checking if generic platform (Apple Silicon)')
+  is_platform_generic = target_machine.cpu_family() == 'aarch64' and run_command([
+    import('python').find_installation('python3'),
+    '-c',
+    'print(open("/sys/devices/system/cpu/cpu0/regs/identification/midr_el1").read() == "")'
+    ],
+    check: true,
+  ).stdout().strip() == 'True'
+
   message('Building ..')
   run_command([
       find_program(host_machine.system() == 'freebsd' ? 'gmake' : 'make'),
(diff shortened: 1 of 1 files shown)


subprojects/packagefiles/spdk/meson.build
```diff
@@ -20,14 +20,28 @@ if get_option('build_subprojects') and not fs.exists('build' / 'lib' / 'src' / '
...
+
   message('Building ..')
   run_command([
       find_program(host_machine.system() == 'freebsd' ? 'gmake' : 'make'),
       '-j',
       cpu_count
     ],
     capture: true,
-    check: true
+    check: true,
```
safl, line 43: Shouldn't this still check error-value?

subprojects/packagefiles/spdk/meson.build
```diff
@@ -20,14 +20,28 @@ if get_option('build_subprojects') and not fs.exists('build' / 'lib' / 'src' / '
     check: true,
   ).stdout().strip()
 
+  message('Checking if generic platform (Apple Silicon)')
+  if target_machine.cpu_family() == 'aarch64'
```
safl, line 24: How about:
```
is_platform_generic = target_machine.cpu_family() == 'aarch64' and run_command([
    import('python').find_installation('python3'),
    '-c',
    'print(open("/sys/devices/system/cpu/cpu0/regs/identification/midr_el1").read() == "")'
  ],
  check: true,
).stdout().strip() == 'True'
```

?

safl (commented):
Great! Two questions/comments inline.

subprojects/packagefiles/spdk/meson.build
safl: Oh, it still does, not there is just ",". Misread, nevermind.

subprojects/packagefiles/spdk/meson.build
Baekalfen: You're right. I originally intended to return the string, that's why. Fixed now.

safl (approved):
Thanks!

PR #165, opened 2022-09-25
be:linux:libaio: add spin-while-waiting

safl:
The libaio backend implementation conservatively waits for at least a
single I/O to complete and a timespec to put an upper bound on how long
it waits. Such usage of the Linux aio interface is common as it
conserves CPU utilization.

However, one can trade CPU utilization for reduced latency and increased
IO rate by repeatedly calling io_getevents() with a min-value of 0. This
commit enables such behavior via the "poll_io" option for the libaio
backend implementation.

### DIFF
--- a/include/xnvme_be_linux_libaio.h
+++ b/include/xnvme_be_linux_libaio.h
@@ -11,7 +11,8 @@ struct xnvme_queue_libaio {
 	io_context_t aio_ctx;
 	struct io_event *aio_events;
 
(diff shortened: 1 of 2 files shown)


krish-reddy (approved):
LGTM

ankit-sam (approved):

safl:
Improve the commit message, no content change. Thus, re-using PR-review-approval.

PR #144, opened 2022-08-25
be:ramdisk add write-zeroes

karlowich:

PR #128, opened 2022-07-15
be:ramdisk: add iovec support

karlowich:
Add support for iovec payloads.

Before merging this:
- Wait for #132 to be merged
- [x] add testcases for iovec

PR #168, opened 2022-09-28
ci,toolbox: add ramdisk testing

safl:
There is currently no tests of the ramdisk functionality. This adds basic testing of it by:

* Add skip of tests for functionality the ramdisk does not implement
* Add the ramdisk to pytest-parametrization via xnvme_be_combinations
* Add a cijoe-configuration for running tests on local environment using the ramdisk
* Add a workflow running tests using the ramdisk

Additionally, the ``test-ramdisk.workflow`` is executed on Linux Distributions with recent Python and gcc.

This is useful in itself, and additionally a preparation for running full memory-leak checks with ``valgrind`` and other runtime-checkers as this will enable checks of the "upper-half" of xNVMe, that is, the helper-APIs and the backend factory.

This is also done as an enabler for PRs:

* https://github.com/OpenMPDK/xNVMe/pull/144
* https://github.com/OpenMPDK/xNVMe/pull/128

Such that there is infra for verifying the behavior of those two additions.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -466,6 +466,59 @@ jobs:
         python3 -m pip install -r python/xnvme-cy-bindings/xnvme/cython_bindings/tests/requirements.txt --user
         python3 -m pytest --pyargs xnvme.cython_bindings -k 'test_version' -s
 
+    #
+    # Ramdisk testing
+    #
+    # skipping on distros with too old gcc/clang to build fio, and too old a Python to run cijoe
+    #
+    - name: Remove any existing cijoe installation and install from toolbox
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      run: |
+        python3 -m pip uninstall cijoe cijoe-pkg-xnvme cijoe-pkg-linux cijoe-pkg-qemu cijoe-pkg-example -y
+        python3 -m pip list
+        python3 -m pip install cijoe --upgrade
+        pushd toolbox/cijoe-pkg-xnvme && make all && popd
+        cijoe -r
+
+    - name: Install fio
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      run: |
+        cd subprojects/fio
+        ./configure --disable-xnvme --prefix=/opt/fio
+        make -j $(nproc)
+        make install
+
+    - name: Adjust the cijoe-configuration to match the location of the fio engine
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      run: |
+        sed -i "s|path: /usr/local.*|path: $(pkg-config xnvme --variable=libdir)/libxnvme-fio-engine.so|" toolbox/workflow/configs/ramdisk.config
+
+    - name: CIJOE, run test-ramdisk.workflow
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      run: |
+        cd toolbox/workflow && cijoe \
+        --config "configs/ramdisk.config" \
+        --workflow "test-ramdisk.workflow" \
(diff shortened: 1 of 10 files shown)


lib/xnvme_buf.c
```diff
@@ -13,6 +13,7 @@ xnvme_buf_virt_alloc(size_t alignment, size_t nbytes)
 {
 	if (!nbytes) {
 		errno = EINVAL;
+		XNVME_DEBUG("nbytes(%zu)", nbytes);
```
karlowich, line 16: I think this is the right idea, however maybe the debug message could be expanded to be more useful.
Maybe something like:
`XNVME_DEBUG("FAILED: invalid value for nbytes: %zu", nbytes);`

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
```diff
@@ -118,6 +126,8 @@ def test_log(cijoe, device, be_opts, cli_args):
 
     if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
         pytest.skip(reason="[admin=block] does not implement get_log")
+    if "ramdisk" in device["labels"]:
+        pytest.skip(reason="[be=ramdisk] does not implement error-log")
```
karlowich, line 130: Should be `get_log` (or maybe `get-log`) instead of error-log

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
```diff
@@ -118,6 +126,8 @@ def test_log(cijoe, device, be_opts, cli_args):
 
     if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
         pytest.skip(reason="[admin=block] does not implement get_log")
```
karlowich, line 128: should this be `get-log` instead of get_log?

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
```diff
@@ -159,6 +175,8 @@ def test_padc(cijoe, device, be_opts, cli_args):
 
     if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
         pytest.skip(reason="[admin=block] does not support admin-passthru")
+    if "ramdisk" in device["labels"]:
+        pytest.skip(reason="[be=ramdisk] does not implement this!?")
```
karlowich, line 179: I am confused as to why the `!?` is here

karlowich (changes requested):
Looks good, I left a few comments

lib/xnvme_buf.c
safl: Agree, fixed!

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
safl: Changing to get-log. thanks!

toolbox/cijoe-pkg-xnvme/src/cijoe/xnvme/tests/tools/test_xnvme.py
safl: yeah, changed to "get-log".

karlowich (approved):

.github/workflows/verify.yml
```diff
@@ -431,6 +431,59 @@ jobs:
...
 
+    #
+    # Ramdisk testing
+    #
+    # skipping on distros with too old gcc/clang to build fio, and too old a Python to run cijoe
+    #
+    - name: Remove any existing cijoe installation and install from toolbox
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
```
Baekalfen, line 442: Could we define these once and store them in a variable? It's hard to tell if certain steps are meant for all or some of the containers.

toolbox/workflow/configs/debian-bullseye.config
```diff
@@ -137,6 +137,12 @@ devices:
   driver_attachment: "kernel"
   spdk_conf:
 
+- uri: "2GB"
+  nsid: 1
+  labels: ["dev", "bdev", "nvm", "ramdisk"]
+  driver_attachment: "kernel"
+  spdk_conf:
```
Baekalfen, line 144: Is it possible to just omit these requirement?

.github/workflows/verify.yml
```diff
@@ -431,6 +431,59 @@ jobs:
...
+    - name: CIJOE, upload test-ramdisk-report
+      uses: actions/upload-artifact@v2
+      if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
+        'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
+        '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      with:
+        name: test-ramdisk-${{ matrix.container.os }}-${{ matrix.container.ver }}
+        path: toolbox/workflow/test-${{ matrix.container.os }}-${{ matrix.container.ver }}/*
+        if-no-files-found: error
+
```
Baekalfen, line 486: Could we add it on macOS as well?

toolbox/workflow/configs/debian-bullseye.config
safl: The driver-helper needs to know about the "driver_attachment". So that one needs to be there.
The "spdk_conf" could be omitted.

.github/workflows/verify.yml
safl: Perhaps, I haven't looked into that.

This PR was just a means to un-block:

- https://github.com/OpenMPDK/xNVMe/pull/128
- https://github.com/OpenMPDK/xNVMe/pull/144

Such that there was some kind of testing of the ramdisk. A lot more can be done, but that was out of scope here.

PR #169, opened 2022-09-29
Misc. ci and testing

safl:
This PR contains a handful of changes to CI and testing which either improves or prepares for further improvement.
Please see the commit-messages for details.
Additional notes are also provided below.

# Reduce time spent on switching driver attachment

At the time of creating this PR then a testrun consumed about 2 hours of wall-clock time. This is an attempt at cutting that down.

- [x] Each test checks whether switching driver-attachment is needed and switches accordingly. This side-effect task consumes a lot of time. To improve this, then tests are grouped to run for all instances requiring a certain configuration. This has cut +30min of the testing time.

# Reduce time spent doing I/O

To reduce the time spent, the goal was to make emulated I/O faster. Currently attempted:

* Running qemu with qcow-images on ``tmpfs``
* Running qemu with qcow-images on ``/dev/shm``

Surprisingly, neither of these approaches, although much faster on a local system, does not yield the same benefits when executed via the GitHUB Runner. This might be due to virtualization and/or container-overhead for I/O. The root cause have not been found.

Another approach now, is to take the most time-consuming tests using ``fio`` and reduce the amount of I/O they do. By reducing the verify-job from 1GB to 64M. This is not ideal as it reduces the confidence of tests, it would be preferable to figure out the issue with the lack of memory-backed ``tmpfs`` or ``/dev/shm``.

# Artifact generation

When provisioning a guest locally for development, then using the ``provision.workflow`` requires "artfiacts" to exist in "/tmp/artifacts". The intent when provisioning locally is that one downloads the artifacts from a failed run to reproduce it locally. However, it is often useful to generate that same set of artifacts locally. In this PR there is a change providing a ``make gen-artifacts`` helper to do exactly that.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -503,6 +503,7 @@ jobs:
       if: (!(contains(matrix.container.ver, 'centos7') || contains(matrix.container.ver, 'opensuse') || contains(matrix.container.ver,
         'buster') || contains(matrix.container.ver, 'bionic') || contains(matrix.container.ver, 'focal') || contains(matrix.container.ver,
         '15.4') || contains(matrix.container.ver, '15.3') || contains(matrix.container.os, 'gentoo')))
+      continue-on-error: true
       run: |
         cd toolbox/workflow && cijoe \
         --config "configs/ramdisk.config" \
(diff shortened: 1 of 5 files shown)


safl:
- Split + qemu with nvme-on-tmpfs = 1hour and 29minutes, the test-run itself around 1hour and 14minutes

safl:
Running on ``"/tmp"`` does not seem to be any faster.
Reason is probably that github-workflow runs a docker-container, and within the container the ``/tmp`` is not  tmpfs mount.

Baekalfen (approved):
I would like to see us use 64M instead of 1G for the fio tests, as I find it unlikely that it will catch more bugs

karlowich (approved):

PR #171, opened 2022-10-04
Reduce CI wall-clock

safl:
Before this PR wall-clock was around 1hour and 30min for the debian testrun.
With this PR the wall-clock is down to 40min for the debian testrun.

Trying out:

- [x] enable ioeventfd on the emulated nvme controllers.
  - result: https://github.com/OpenMPDK/xNVMe/actions/runs/3180754612
  - no measurable impact, still at 1hour and 30min
- [x] qemu-images raw-format
  - result: they were already in raw-format, nothing to gain there.
- [x] retry /dev/shm + reducing fio work
  - result: https://github.com/OpenMPDK/xNVMe/actions/runs/3180985899
  - result: https://github.com/OpenMPDK/xNVMe/actions/runs/3181649145
  - runs out of memory with hugepages... even in combination with reducing fio-work, dropping this pursuit.
- [x] reduce fio worksize from 1GB to 64M + tweak fabrics TCP transport
  - result: https://github.com/OpenMPDK/xNVMe/actions/runs/3182172002
  - currently seeing a failure but expecting a reduction to 40 minutes
  - yup, after fix, down to 40min for the testrun.

Using ``/dev/shm`` is interesting as fio is not the only I/O test. However, fio is the one with the largest-working set e.g. 1GB. Thus reducing it should have an impact. However, won't improve anything for the other I/O test. Reducing fio might have the added benefit that less working-memory is needed, instead of writing 1GB to the device, only 64M is written, and thus, less use of ``/dev/shm``.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -317,8 +317,18 @@ jobs:
 
     - name: Python sdist-packages, build(xnvme-core, xnvme-cy-header, and xnvme-cy-bindings)
       run: |
-        cd python/xnvme-cy-bindings
-        make clean build-py-env build-sdist
+        pushd python/xnvme-core
+        make clean build
+        popd
+
+        pushd python/xnvme-cy-header
+        make clean build
+        popd
+
+        pushd python/xnvme-cy-bindings
+        # Explicitly avoid bdist build
+        make clean .build-py-env build-sdist
+        popd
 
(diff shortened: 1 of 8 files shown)


Baekalfen (approved):
Looks good, and great improvements!

karlowich (approved):
Looks good!

PR #172, opened 2022-10-04
mk:py: refactor python makefiles

Baekalfen:
Changed the ownership of the build process to a Makefile in python/ instead of xnvme-cy-bindings/Makefile taking control of the other packages.

Improved build-times by avoiding redundant rebuild of build env.

If you want to clean, build or install all of the Python packages, do it from the python/Makefile, and save time by only cleaning when you change the dependencies in the packages.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -1,0 +1,81 @@
+#
+# This Makefile serves as convenient command-line auto-completion when working on Python bindings
+#
+#
+
+define default-help
+# invoke: 'make uninstall', 'make install'
+endef
+.PHONY: default
+default: build
+	@echo "## py: make default"
+	@echo "## py: make default [DONE]"
+
+
+define all-help
+# Do the common task during development of: uninstall clean build install
+endef
+.PHONY: all
+all: uninstall clean build install
+
+define build-help
+# Call build on all subpackages
+endef
+.PHONY: build
+build:
+	@cd xnvme-core && make build
+	@cd xnvme-cy-header && make build
+	@cd xnvme-cy-bindings && make build
+
+
+define install-help
+# install all subpackages for current user
+endef
+.PHONY: install
+install:
+	@cd xnvme-core && make install
+	@cd xnvme-cy-header && make install
+	@cd xnvme-cy-bindings && make install
+
+define install-system-help
+# install all subpackages system-wide
+endef
+.PHONY: install-system
+install-system:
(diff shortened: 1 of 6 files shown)


python/Makefile
```diff
@@ -0,0 +1,87 @@
...
+.PHONY: clean
+clean:
+	@cd xnvme-core && make clean
+	@cd xnvme-cy-header && make clean
+	@cd xnvme-cy-bindings && make clean
+
+define help-help
+# Print the description of every target
+#
+# Print the description of every target
```
karlowich, line 74: This line is duplicated and can be removed

python/Makefile
```diff
@@ -0,0 +1,87 @@
...
+.PHONY: install
+install:
+	@cd xnvme-core && make install
+	@cd xnvme-cy-header && make install
+	@cd xnvme-cy-bindings && make install
+
+define install-system-help
+# install system-wide
+#
+# install system-wide
```
karlowich, line 43: This line is duplicated and can be removed

python/Makefile
```diff
@@ -0,0 +1,87 @@
...
+# Print the description of every target
+endef
+.PHONY: help
+help:
+	@./$(TOOLBOX_DIR)/print_help.py --repos .
+
+define help-verbose-help
+# Print the verbose description of every target
+#
+# Print the verbose description of every target
```
karlowich, line 83: This line is duplicated and can be removed

python/Makefile
```diff
@@ -0,0 +1,87 @@
...
+
+
+define all-help
+# Do the common task during development of: uninstall clean build install test
+endef
+.PHONY: all
+all: uninstall clean build install
+
+define build-help
+# Generate cython pxd, ctypes, patch them, then create a source package
```
karlowich, line 22: This seems a bit misleading, the source package doesn't seem to be created in this target

karlowich (changes requested):
Looks good, I especially like the addition of the `-f` for the `rm` commands.

I left a few comments

python/Makefile
Baekalfen: Updated

karlowich (approved):

safl (approved):
LGTM.

safl:
I pulled the trigger too quickly on this one, it break ``make gen-artifacts``.
@Baekalfen can you take a look?

PR #174, opened 2022-10-05
mk: fix gen_artifacts afte python refactor

Baekalfen:
gen_artifacts didn't correctly call build of all of the Python packages, and would fail to find the xnvme-core and xnvme-header dependencies.


safl (approved):
Great, thanks!

PR #178, opened 2022-10-10
Enable WQ sharing (ATTACH_WQ) for SQPOLL

safl:
This PR provides the means to and changes behavior of ``async=io_uring`` and ``async=io_uring_cmd`` in the following ways:

* Share the kernel-side sqpoll thread-pool among multiple ``xnvme_queue`` instances
  * Do this by default, disable with env(``XNVME_QUEUE_SQPOLL_AWQ=0``)
* Provides control of cpu affinity for the thread-pool via environment variable ``XNVME_QUEUE_SQPOLL_CPU``
* Ads ``SINGLE_ISSUER`` optimization

### DIFF
--- a/lib/xnvme_be_linux_async_liburing.c
+++ b/lib/xnvme_be_linux_async_liburing.c
@@ -63,8 +63,8 @@ xnvme_be_linux_liburing_init(struct xnvme_queue *q, int opts)
 {
 	struct xnvme_queue_liburing *queue = (void *)q;
 	struct xnvme_be_linux_state *state = (void *)queue->base.dev->be.state;
+	struct io_uring_params ring_params = {0};
 	int err = 0;
-	int iou_flags = 0;
 
 	if ((opts & XNVME_QUEUE_SQPOLL) || (state->poll_sq)) {
 		queue->poll_sq = 1;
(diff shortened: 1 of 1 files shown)


lib/xnvme_be_linux_async_liburing.c
```diff
@@ -80,18 +80,18 @@ xnvme_be_linux_liburing_init(struct xnvme_queue *q, int opts)
...
 
 	if (opts & XNVME_QUEUE_IOU_BIGSQE) {
-		iou_flags |= IORING_SETUP_SQE128;
-		iou_flags |= IORING_SETUP_CQE32;
+		ring_params.flags |= IORING_SETUP_SQE128;
+		ring_params.flags |= IORING_SETUP_CQE32;
 	}
 
-	err = io_uring_queue_init(queue->base.capacity, &queue->ring, iou_flags);
+	err = io_uring_queue_init_params(queue->base.capacity, &queue->ring, &ring_params);
```
ankit-sam, line 122: Looks good. I was comparing with FIO io_uring and it seems we are missing the new options. See these recent commits

https://github.com/axboe/fio/commit/e453f369ecf4db4913ef6916603037f2d7141035
https://github.com/axboe/fio/commit/4d22c1037dafdac3d9bf30a3fb39d88f4391e41a

especially the 2nd one, where Jens claims that it improves RR for QD8 from 680K to 870K

safl: I am taking a look!

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -80,6 +84,24 @@ xnvme_be_linux_liburing_init(struct xnvme_queue *q, int opts)
 	// Ring-initialization
 	//
 	if (queue->poll_sq) {
+		if (!g_sqpoll_ring_initialized) {
```
ankit-sam, line 87: I was wondering if we need a lock for this?

karlowich (approved):
Looks good!

lib/xnvme_be_linux_async_liburing.c
safl: Good point. I am wavering a bit between adding a lock or just adding a comment on  ``xnvme_queue_init()`` saying that it is not thread-safe. Since, currently none of the backends guard the ring/qpair/control-block/etc. initializations.
However, a quick skim indicates that they do not need to... so I guess the best thing would be to add a lock here, so be on the safe side, and then go over the implementations ensuring that they are thread-safe, fix them, and then note the state.

lib/xnvme_be_linux_async_liburing.c
safl: Seems like:

- ``IORING_SETUP_SINGLE_ISSUER``
- ``IORING_SETUP_DEFER_TASKRUN``

Are  liburing 2.2+, so I am creating an issue of things to address with the next liburing release.
Seems like:
- ``IORING_SETUP_COOP_TASKRUN | IORING_SETUP_TASKRUN_FLAG``
Is low-hanging fruit, man-page mentions a use-case where it is less ideal (one thread submitting another waiting on completions). But, does not say that it directly hurts performance in that case. Worth taking for a spin.

safl: Tried a different approach for ``IORING_SETUP_SINGLE_ISSUER``, added it under an ``ifndef`` and ``queue_init_params()``-retry in case of EINVAL. It seemed like low-hanging fruit to enable it in the context of SQPOLL done here.

lib/xnvme_be_linux_async_liburing.c
safl: Locking added and improved error-handling.

PR #182, opened 2022-10-14
be:win: added IORING async interface backend for Windows OS.

rs-sam:
note: currently supports only Read IO path.

Signed-off-by: Rishabh Shukla <rishabh.sh@samsung.com>

### DIFF
--- a/lib/xnvme_be_windows_async_ioring.c
+++ b/lib/xnvme_be_windows_async_ioring.c
@@ -1,0 +1,207 @@
+// Copyright (C) Rishabh Shukla <rishabh.sh@samsung.com>
+// SPDX-License-Identifier: Apache-2.0
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE 700
+#endif
+#include <xnvme_be.h>
+#include <xnvme_be_nosys.h>
+#ifdef XNVME_BE_WINDOWS_ASYNC_ENABLED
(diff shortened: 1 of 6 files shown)


lib/xnvme_be_windows_async_ioring.c
```diff
@@ -0,0 +1,254 @@
...
+#include <winternl.h>
+#include <time.h>
+#include <intrin.h>
+#include <ioringapi.h>
+#include <libxnvme_spec_fs.h>
+#include <xnvme_queue.h>
+#include <xnvme_dev.h>
+#include <xnvme_be_windows.h>
+
+#define MAX_IORING_SQ 0x10000
```
karlowich, line 26: I think it would be good to move line 26 to 69 to a header file in `include/`, perhaps it should be called `xnvme_be_windows_ioring.h`. That way this file is cleaner and follows the convention of fx. `lib/xnvme_be_linux_async_liburing.c`

karlowich (changes requested):
This looks good!
I added a single comment inline

lib/xnvme_be_windows_async_ioring.c
rs-sam: included a 'xnvme_be_windows_ioring.h' file as per the review comment. Thanks @karlowich

safl (approved):
Thanks @rs-sam !

safl:
@karlowich I move this ahead as @rs-sam had addressed your comments.

Issue #185, opened 2022-10-17
Cython breaks on Debian Bookworm


safl:
The basic Cython tests are failing on Debian Bookworm, see failure examples here:

* https://github.com/OpenMPDK/xNVMe/actions/runs/3262527630/jobs/5360758574
* https://github.com/OpenMPDK/xNVMe/actions/runs/3263404552/jobs/5362251918

It seems like Python package-dependency/in-compatibility issue as it has "arisen" unrelated to the code in CI of the PRs above.
Work-around on PR: https://github.com/OpenMPDK/xNVMe/pull/186

safl:
Fixed by: https://github.com/OpenMPDK/xNVMe/pull/187

PR #186, opened 2022-10-17
ci: Cython workaround and node-related-action-update

safl:
* Work-around to: https://github.com/OpenMPDK/xNVMe/issues/185
* Update of upload/download actions due to:
```Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/upload-artifact```

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -70,7 +70,7 @@ jobs:
         make clean gen-src-archive
 
     - name: Upload source archive
-      uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


karlowich (changes requested):
Looks good!
The workaround can be removed, since we now have #187

safl:
> Looks good! The workaround can be removed, since we now have #187

done.

karlowich (approved):

PR #187, opened 2022-10-18
ci: Fix setuptools on Debian Bookworm

Baekalfen:
Debian Bookworm has currently an issue with its python3-setuptools package. If we force-install a recent version from PyPi, it builds again.

Added debug information to see the current setuptools version.


karlowich (approved):
Nice!

safl (changes requested):
Looks like this break on Gentoo?

Baekalfen:
> Looks like this break on Gentoo?

I'm really starting to hate Linux...

Maybe fixed. Let's wait for the build to finish.

safl (commented):
Great, thanks!

safl (approved):

PR #183, opened 2022-10-14
Remove SLIST(sys/queue.h) from API

safl:
Now, two things:

* Removes the "SLIST" data structure from the public API
* Adds a missing header-inclusion in the Windows backend

See commit-messages for details.

PR #188, opened 2022-10-19
test,python: mark test_max_segments_ioctl as flaky

Baekalfen:
By marking the test as flaky, and defining the conditions for a rerun, we will reduce the chance that this test fails.


safl (commented):
The use of ``pytest-ignore-flaky`` seems like the right approach, however, on:

* https://github.com/OpenMPDK/xNVMe/pull/183

The test has failed three times in a row, so how about instead of failing after three attempts, then it ignores it?

Baekalfen:
> 

In that case, #189 is simpler with the same outcome

PR #189, opened 2022-10-19
test,python: test_max_segments_ioctl if allocation fails

Baekalfen:
We cannot always consistently get a segmented buffer, as there is a bit of chance involved.

If we fail to allocate correctly, we just skip the test. This way, we will optimistically test when the buffer allocates, and otherwise, allow to pass the test without it.


Baekalfen:
This is an alternative to #188

safl (approved):
LGTM, thanks!

PR #190, opened 2022-10-20
be: add earmark to dev_open() during enumeration

### LINKED ISSUE #127: SPDK backend fails enumeration
SPDK has an issue during enumeration, when it calls `xnvme_dev_open` here: https://github.com/OpenMPDK/xNVMe/blob/e507f2d1791163d0c2757f1b973cd2cf6dfcb573/lib/xnvme_be_spdk_dev.c#L611

It doesn't force `spdk` backend in the `opts` variable, therefore you may trigger `vfn` or another PCIe backend to try to open the device. This can happen if `spdk` fails to open the device, and `xnvme_dev_open` continues to probe other backends.

karlowich:
Pass a copy of opts, with the respective backend set, to xnvme_dev_open().
This is to avoid a case where a different backend is used for
enumeration and opening the device.

Closes #127

### DIFF
--- a/lib/xnvme_be_fbsd_dev.c
+++ b/lib/xnvme_be_fbsd_dev.c
@@ -20,6 +20,8 @@ xnvme_be_fbsd_enumerate(const char *sys_uri, struct xnvme_opts *opts, xnvme_enum
 		XNVME_DEBUG("FAILED: sys_uri: %s is not supported", sys_uri);
 		return -ENOSYS;
 	}
+	struct xnvme_opts tmp_opts = *opts;
+	tmp_opts.be = xnvme_be_fbsd.attr.name;
 
 	for (int cid = 0; cid < 256; cid++) {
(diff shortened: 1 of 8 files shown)


safl (commented):
This should do the trick :) Let's see what the CI says.

This brings up the need for some explicit testing of ``xnvme_dev_open()`` and what it manages to "realize". With the recent addition of ``xnvme_dev_get_opts()``, there is path to verify when explicitly specifying backend-options, that those requested interfaces are actually what the device handled ended up with.

Would you be up for adding a test like add?

Baekalfen:
> With the recent addition of `xnvme_dev_get_opts()`, there is path to verify when explicitly specifying backend-options, that those requested interfaces are actually what the device handled ended up with.

+1 for making tests for this.

Also looks like a good fix @karlowich

safl (approved):
Great, thanks!

PR #177, opened 2022-10-07
cli: add macro to simplify options

karlowich:
closes #175

Signed-off-by: Karl Bonde Torp <k.torp@samsung.com>

PR #191, opened 2022-10-21
be:spdk:dev: remove code duplication and general cleaning

karlowich:
This together with #177 is part of an effort to make the code easier to maintain.

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -27,8 +27,7 @@ static struct xnvme_be_spdk_ctrlr_ref g_cref[XNVME_BE_SPDK_CREFS_LEN];
 /**
  * look for a cref for the given given ident
  *
- * @return On success, a pointer to the cref is returned. When none existsWhen a
- * ref is found, 1 is returned cref is assigned, 0 otherwise.
+ * @return On success, a pointer to the cref is returned. When none exists, NULL is returned.
  */
 static struct spdk_nvme_ctrlr *
 _cref_lookup(struct xnvme_ident *id)
@@ -57,7 +56,7 @@ _cref_lookup(struct xnvme_ident *id)
 /**
  * Insert the given controller
  *
- * @return On success, 0 is return. On error, negated ``errno`` is return to
+ * @return On success, 0 is returned. On error, negated ``errno`` is returned to
  * indicate the error.
  */
 static int
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -361,6 +361,41 @@ _spdk_nvme_transport_id_compare_weak(const struct spdk_nvme_transport_id *trid1,
 	return spdk_nvme_transport_id_compare(&one, &other);
 }
 
+bool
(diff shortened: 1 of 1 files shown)


lib/xnvme_be_spdk_dev.c
```diff
@@ -361,6 +361,41 @@ _spdk_nvme_transport_id_compare_weak(const struct spdk_nvme_transport_id *trid1,
 	return spdk_nvme_transport_id_compare(&one, &other);
 }
```
safl, line 360: Looks good.

lib/xnvme_be_spdk_dev.c
```diff
@@ -217,8 +217,7 @@ _spdk_env_init(struct spdk_env_opts *opts)
 
 	if (g_xnvme_be_spdk_env_is_initialized) {
 		XNVME_DEBUG("INFO: already initialized");
-		err = 0;
-		goto exit;
+		return 0;
```
safl, line 219: This is probably a matter of taste whether it is simpler/cleaner.
I found that it was simpler to be able to always look for:
```
INFO: spdk_env_is_initialized: 0/1
```
In the output, to figure out whether initialization succeeded, rather than ``spdk_env_is_initialized`` sometimes being there and sometimes not.

Also that the debug would explicitly show that ``INFO: already initialized`` when re-using.

lib/xnvme_be_spdk_dev.c
```diff
@@ -27,8 +27,7 @@ static struct xnvme_be_spdk_ctrlr_ref g_cref[XNVME_BE_SPDK_CREFS_LEN];
 /**
  * look for a cref for the given given ident
  *
- * @return On success, a pointer to the cref is returned. When none existsWhen a
- * ref is found, 1 is returned cref is assigned, 0 otherwise.
+ * @return On success, a pointer to the cref is returned. When none exists, NULL is returned.
```
safl, line 30: Great.

safl (commented):
Looks good. Only a small thing regarding the debug output that I don't entirely agree with, a made a comment as to why it currently is as it is. Let me know your thoughts about it as you are now spending time on that code path.

lib/xnvme_be_spdk_dev.c
```diff
@@ -677,43 +677,15 @@ xnvme_be_spdk_enumerate(const char *sys_uri, struct xnvme_opts *opts, xnvme_enum
...
-				continue;
-			}
-			break;
-
-		case SPDK_NVME_TRANSPORT_FC:
-		case SPDK_NVME_TRANSPORT_CUSTOM:
-			XNVME_DEBUG("SKIP: enum. of trtype: %s not implemented",
-				    spdk_nvme_transport_id_trtype_str(trtype));
-			continue;
+		err = _xnvme_be_spdk_ident_to_trid(&ident, opts, &trid, trtype);
```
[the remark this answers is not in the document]
karlowich: Done

lib/xnvme_be_spdk_dev.c
karlowich: The commit has been dropped, I see what you mean

safl (approved):
The code needs some re-organization in this area, this is a great start.
Thanks!

PR #194, opened 2022-10-24
py: fix void_pointer casting

Baekalfen:
Fix issue with casting between encapsulated xNVMe objects in Python. Cython wouldn't do the proper casting, and cast a PyObject* instead of the integer value it represents.


karlowich (approved):
This looks good!

PR #198, opened 2022-10-27
test:enum: add test for individual backends

karlowich:
This will test if the same backend is used for enumerating and opening devices.

This is related to #127 and #190


karlowich:
This has been included in #190

PR #199, opened 2022-10-31
scripts: update xnvme-driver

karlowich:
The old xnvme-driver failed on FreeBSD because of a syntax error.
This updates the script to the latest SPDK version.


karlowich:
Currently, this breaks more than it fixes

safl:
As discussed offline, let's go with a quick-fix for FreeBSD, then we can consider doing a pure-Python tool for this aligned with the requirements of upstream SPDK / setup.sh. As all projects need the same thing then there is room to make something easily/self-contained deployment in a pure-Python package, that would be really convenient.

PR #200, opened 2022-11-07
docs/contributing: add PR checklist, DCO descr. and msg-format-change

safl:
This contains three changes to the contributors section.

1. Provide a check-list of what is expected on behalf of the contributor

2. Explicit mention of DCO, this is preparation for the license-change,
   as with it, then the CONTRIBUTING.md will refer to the online
   documentation and no longer have a mention of the DCO.

3. Change the "component" prefix convention for example:

-   from: ``be:linux:async:libaio: ...``
-   to: ``be/libaio: ...``

By doing so, the component prefix becomes a lot shorter and the "sub-component" uses "/" instead of ":". Then ":" is used exclusively for the separation of the component from the brief description.

This is done to make the messages shorter and align closer with the commit-style of other projects.

### DIFF
--- a/docs/contributing/index.rst
+++ b/docs/contributing/index.rst
@@ -4,7 +4,7 @@
  Contributing
 ==============
 
-**xNVMe** is hosted on `GitHUB`_ where you can raise `issues`_, start a
+**xNVMe** is hosted on `GitHUB`_ where you can raise `issues`_, start
 `discussions`_, or set up a `pull-request`_ for contributing your changes.
 This page contains information related to the latter. Specifically, the
 :ref:`sec-contributing-process` is of relevance to you and the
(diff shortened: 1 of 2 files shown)


docs/contributing/index.rst
```diff
@@ -23,37 +23,79 @@ Contribution Process
...
 ---------------
 
 1. The first line is subject/title
 
    - Keep it at a max. of 72 chars, if possible
-   - Prefix with the compoent e.g. ``be:linux: fix error-handling``
    - Lower-case is preferred
-   - Use the "imperative mood" e.g. ``fix`` rather ``fixes``/``fixed``
    - Do **not** end with a punctuation
+   - Use the **imperative mood** e.g. ``add`` rather ``added``/``adds``
```
karlowich, line 70: Maybe change
> rather

to **rather than**

docs/contributing/index.rst
```diff
@@ -23,37 +23,79 @@ Contribution Process
...
+    - In case of multiple components, then separate with a comma e.g.
+      ``ident,be/linux: ...```
+   - **Example**: ``be/libaio: add spin-while-waiting for opts.poll_io``
 
 2. Then an empty line.
 
 3. Then a description (may be omitted for truly trivial changes).
 
-   - Should explain what and why not how
+   - Should motivate the change and explain what and why not how
```
karlowich, line 86: Maybe add a comma after why to make the sentence clearer, like this:

Should motivate the change and explain what and **why,** not how

karlowich (changes requested):
This is looking good, I have left two comments

docs/contributing/index.rst
safl: Fixed, thanks!

docs/contributing/index.rst
safl: Fixed, thanks!

karlowich (approved):

PR #205, opened 2022-11-15
Change format-specifiers to PRI-macro

safl:
The format-specificers used for hex-values e.g. lx/llx does not match the on Mac/clang. This changes the the use of those format-specifiers to use the PRIx64/PRIu64/PRIx16 macros.

### DIFF
--- a/lib/xnvme_be.c
+++ b/lib/xnvme_be.c
@@ -87,12 +87,12 @@ xnvme_lba_fpr(FILE *stream, uint64_t lba, enum xnvme_pr opts)
 
 	switch (opts) {
 	case XNVME_PR_TERSE:
-		wrtn += fprintf(stream, "0x%016lx", lba);
+		wrtn += fprintf(stream, "0x%016" PRIx64, lba);
 		break;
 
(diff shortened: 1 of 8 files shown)


niclashedam (approved):
Format-specifier warnings are now resolved on MacOS Ventura 13.0 (22A380).

karlowich (approved):
Looks good!

One unrelated thing I noticed:
There are a LOT of "TODO"s in `lib/xnvme_spec.c`, I think it would good to remove them and create an issue on GitHub instead.

safl:
> 

Yes, it would be nice with a combover of the the ``lib/xnvme_spec.c``, however, it should be refactored based on Python/C++/Rust/Zig FFI compatibility and general portability. We still need some further "data" for that, and then ideally, then ``lixnvme_spec.h`` and ``xnvme_spec.c`` would be items we could emit code for based on [ NVMe-Lint ](https://github.com/linux-nvme/nvme-lint) 

But yeah, these things should be described in an issue :)

niclashedam (approved):

PR #208, opened 2022-11-15
ci: update upload/download-artifact actions

safl:
The actions use a feature which will be deprecated, thus the CI shows a bunch of deprecation warnings. The latest latest versions 3.1.1 and 3.0.1 have adjusted and updated them should get rid of the deprecation messages.


karlowich (approved):
Nice!

PR #212, opened 2022-11-19
Fix macOS warnings

### LINKED ISSUE #206: Deprecation-warnings on MacOS
In addition to format-specifiers giving build-warnings (this is addressed here: https://github.com/OpenMPDK/xNVMe/pull/205), then the following deprecation-warnings and additional warnings need some attention:

```
[24/202] Compiling C object lib/libxnvme-shared.dylib.p/xnvme_be_macos_dev.c.o
../lib/xnvme_be_macos_dev.c:33:42: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        matching_dictionary = IOBSDNameMatching(kIOMasterPortDefault, 0, basename(devname));
                                                ^~~~~~~~~~~~~~~~~~~~
                                                kIOMainPortDefault
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:133:19: note: 'kIOMasterPortDefault' has been explicitly marked deprecated here
const mach_port_t kIOMasterPortDefault
                  ^
../lib/xnvme_be_macos_dev.c:34:49: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        ioservice_device = IOServiceGetMatchingService(kIOMasterPortDefault, matching_dictionary);
                                                       ^~~~~~~~~~~~~~~~~~~~
                                                       kIOMainPortDefault
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:133:19: note: 'kIOMasterPortDefault' has been explicitly marked deprecated here
const mach_port_t kIOMasterPortDefault
                  ^
../lib/xnvme_be_macos_dev.c:78:7: warning: incompatible pointer types passing 'IONVMeSMARTInterface ***' (aka 'struct IONVMeSMARTInterface ***') to parameter of type 'LPVOID *' (aka 'void **') [-Wincompatible-pointer-types]
                                         nvme_smart_interface);
                                         ^~~~~~~~~~~~~~~~~~~~
../lib/xnvme_be_macos_dev.c:178:19: warning: incompatible pointer to integer conversion passing 'IONVMeSMARTInterface **' (aka 'struct IONVMeSMARTInterface **') to parameter of type 'io_object_t' (aka 'unsigned int') [-Wint-conversion]
                IOObjectRelease(state->nvme_smart_interface);
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:268:14: note: passing argument to parameter 'object' here
        io_object_t     object );
                        ^
../lib/xnvme_be_macos_dev.c:262:27: warning: incompatible pointer to integer conversion assigning to 'io_object_t' (aka 'unsigned int') from 'void *' [-Wint-conversion]
                state->ioservice_device = NULL;
                                        ^ ~~~~
5 warnings generated.
[94/202] Compiling C object lib/libxnvme-static.a.p/xnvme_be_macos_dev.c.o
../lib/xnvme_be_macos_dev.c:33:42: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        matching_dictionary = IOBSDNameMatching(kIOMasterPortDefault, 0, basename(devname));
                                                ^~~~~~~~~~~~~~~~~~~~
                                                kIOMainPortDefault
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:133:19: note: 'kIOMasterPortDefault' has been explicitly marked deprecated here
const mach_port_t kIOMasterPortDefault
                  ^
../lib/xnvme_be_macos_dev.c:34:49: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        ioservice_device = IOServiceGetMatchingService(kIOMasterPortDefault, matching_dictionary);
                                                       ^~~~~~~~~~~~~~~~~~~~
                                                       kIOMainPortDefault
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:133:19: note: 'kIOMasterPortDefault' has been explicitly marked deprecated here
const mach_port_t kIOMasterPortDefault
                  ^
../lib/xnvme_be_macos_dev.c:78:7: warning: incompatible pointer types passing 'IONVMeSMARTInterface ***' (aka 'struct IONVMeSMARTInterface ***') to parameter of type 'LPVOID *' (aka 'void **') [-Wincompatible-pointer-types]
                                         nvme_smart_interface);
                                         ^~~~~~~~~~~~~~~~~~~~
../lib/xnvme_be_macos_dev.c:178:19: warning: incompatible pointer to integer conversion passing 'IONVMeSMARTInterface **' (aka 'struct IONVMeSMARTInterface **') to parameter of type 'io_object_t' (aka 'unsigned int') [-Wint-conversion]
                IOObjectRelease(state->nvme_smart_interface);
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:268:14: note: passing argument to parameter 'object' here
        io_object_t     object );
                        ^
../lib/xnvme_be_macos_dev.c:262:27: warning: incompatible pointer to integer conversion assigning to 'io_object_t' (aka 'unsigned int') from 'void *' [-Wint-conversion]
                state->ioservice_device = NULL;
```

niclashedam:
This PR resolves #206 and the code is based on documentation of macOS Ventura 13.0 (22A380) and compiles successfully on said device.


karlowich (approved):
This looks good!
I have started the tests

niclashedam:
Looks like the CI encountered a segmentation fault. I don’t have a Mac with and NVMe device, so I cannot reproduce. Is it possible to get the backtrace from the CI?

karlowich:
@Baekalfen can you help with this?

Baekalfen:
I can't reproduce it on my machine. There's only a few lines of code in your PR. You could try reverting one change at a time.

niclashedam:
Hi @karlowich, @safl. I tried dereferencing a pointer that may need to be dereferenced. Can you restart the workflow, so I can see if it worked?

lib/xnvme_be_macos_dev.c
```diff
@@ -75,7 +75,7 @@ _get_nvme_smart_interface(io_object_t ioservice_device,
 		(**plugin_interface)
 			->QueryInterface(*plugin_interface,
 					 CFUUIDGetUUIDBytes(kIONVMeSMARTInterfaceID),
-					 nvme_smart_interface);
+					 (void **)*nvme_smart_interface);
```
Baekalfen, line 78: This doesn't look right. It's supposed to be a pointer-pointer. https://developer.apple.com/documentation/iokit/ionvmesmartinterface/3521170-queryinterface

I'm getting segfaults on this locally, while reverting it to `(void **)nvme_smart_interface` is fine.

niclashedam: Then that's not the problem. Thanks for testing.

Baekalfen: No problem :)

niclashedam:
@safl: Regarding the change from `IOObjectRelease` to the release function. `IONVMeSMARTInterface` is not an `IOObject` (`io_object_t`), so that call wasn't right. I'll try and add version 12. Can anyone trigger the build?

karlowich:
@niclashedam Started.
Sorry for the slow response :)

niclashedam:
It seems like it doesn’t work on either systems. I have an intel-based Mac somewhere. I’ll try and see if it works on that.

safl:
> It seems like it doesn’t work on either systems. I have an intel-based Mac somewhere. I’ll try and see if it works on that.
> 

I have added you as a contributor on GitHUB, so you can now re-run failed jobs.

Baekalfen:
> > It seems like it doesn’t work on either systems. I have an intel-based Mac somewhere. I’ll try and see if it works on that.
> 
> I have added you as a contributor on GitHUB, so you can now re-run failed jobs.

You can comment out the Linux and Windows builds in GHA. There's a limited number of runners, so you can end up waiting if you run them in quick succession.

Baekalfen:
> @safl: Regarding the change from `IOObjectRelease` to the release function. `IONVMeSMARTInterface` is not an `IOObject` (`io_object_t`), so that call wasn't right. I'll try and add version 12. Can anyone trigger the build?

You're right, it seems incorrect to use `IOObjectRelease` for that. I've found that smartmontools do as you've done here: [os_darwin.cpp](https://github.com/mirror/smartmontools/blob/d3abf1eca60eede88b995a99d345c8cc1593526e/os_darwin.cpp#L289)

safl:
@niclashedam that alpine build-issue is now fixed on ``next`` by this: https://github.com/OpenMPDK/xNVMe/pull/213
Rebase your branch and you should be back.

niclashedam:
> > It seems like it doesn’t work on either systems. I have an intel-based Mac somewhere. I’ll try and see if it works on that.
> 
> I have added you as a contributor on GitHUB, so you can now re-run failed jobs.

Great. Thanks and thank you for the trust!

niclashedam:
Thanks @Baekalfen !
I managed to get it to run now. I can see that the macOS 12 runner is `x86_64-apple-darwin21.6.0`, so it is not silicon-based. This is good since it is then not a CPU problem but an OS problem. It also passes on macOS 12, so I suppose this is a changed interface. I've gone ahead and removed 11 from the workflow since macOS 11 is deprecated in favour of Monterey and now also Ventura.

If you disagree, we can look for a solution, but I think macOS 12+ is a reasonable minimum requirement.

safl:
> 

That is perfectly fine. To clarify, was the fix to just drop macOS-11, or were there changes to the backend as well?

niclashedam:
My initial solution was OK on macOS 12+. Dropping macOS 11 was enough.

Baekalfen:
> Thanks @Baekalfen ! I managed to get it to run now. I can see that the macOS 12 runner is `x86_64-apple-darwin21.6.0`, so it is not silicon-based. This is good since it is then not a CPU problem but an OS problem. It also passes on macOS 12, so I suppose this is a changed interface. I've gone ahead and removed 11 from the workflow since macOS 11 is deprecated in favour of Monterey and now also Ventura.
> 
> If you disagree, we can look for a solution, but I think macOS 12+ is a reasonable minimum requirement.

Fantastic!

Did you find out what part of the code that caused this? Could it be [`kIOMainPortDefault`](https://developer.apple.com/documentation/iokit/kiomainportdefault?language=objc) that requires "macOS 12.0+"? It's weird that macOS 11 will happily build it and then crash in runtime.

If it is `kIOMainPortDefault`, could we do an `ifdef` on the deployment target? Or even just set the deployment target `MACOSX_DEPLOYMENT_TARGET=11.0` and continue using `kIOMasterPortDefault`?

If the setting the deployment target doesn't work, then let's just say macOS 12.0+.

safl:
@niclashedam the Python issue should now be fixed on ``next``. Please rebase :)

While doing that, find a bit on feedback on the commits below. I am trying to keep the documentation on the contribution-process updated, however, is currently not entirely there... so the following are some changes that would be great.

For all the commits, remember to sign them off, e.g. via: ``git commit -s``, this is the convention stating that you have read and abide to the DCO (https://developercertificate.org/).

### The commit: ``be_macos: Fix deprecation warnings``

Change something like: ``be/macos: adjust to macOS API changes``

And then in the commit-message describe along the lines:

```
On macOS compiler-warnings are emitted as deprecated API macros are used.
This change adjusts them to the latest definition and thereby eradicating compiler-warnings.
```

### The commit: ``be_macos: Fix type warnings``

Rewrite the short-message to something like: ``be/macos: <very brief description>``

Instead of just "fix" then a brief description of what it does and why would be really nice.
That helps share the knowledge of where the warnings came from and how they were removed.

### The commit: ``Add Niclas Hedam to list of contributors``

Change the short-message to: ``CHANGELOG: add Niclas Hedam``

# The commit: ``workflows: add macOS 12 as Runner``

This touches on a bunch of components, a prefix/short message like:

``ci,docs,toolbox/pkgs/macos: replace macOS 11 with macOS 12``

And btw. thanks for looking into this, it is highly appreciated!

niclashedam:
No problem. It's been fun. I've updated the commits.

How does this look? @safl

Baekalfen:
> No problem. It's been fun. I've updated the commits.
> 
> How does this look? @safl

We are not quite ready to remove support for macOS, if it can be avoided. I think we can keep the support, while removing the warning with this:

```
diff --git a/.github/workflows/verify.yml b/.github/workflows/verify.yml
index 368f51df..9e86017f 100644
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -572,6 +572,7 @@ jobs:
       matrix:
         runner:
         - {os: 'macos', ver: '11'}
+        - {os: 'macos', ver: '12'}

     steps:
     - name: Retrieve the source-archive
@@ -585,12 +586,12 @@ jobs:

     - name: Install build-requirements
       run: |
-        source toolbox/pkgs/${{ matrix.runner.os }}-${{ matrix.runner.ver }}.sh || true
+        source toolbox/pkgs/${{ matrix.runner.os }}.sh || true

     - name: Configure, Build, and Install
       env:
         BSCRIPT_DEF: toolbox/pkgs/default-build.sh
-        BSCRIPT: toolbox/pkgs/${{ matrix.runner.os }}-${{ matrix.runner.ver }}-build.sh
+        BSCRIPT: toolbox/pkgs/${{ matrix.runner.os }}-build.sh
       run: |
         if [[ -f "${BSCRIPT}" ]]; then source ${BSCRIPT}; else source ${BSCRIPT_DEF}; fi

diff --git a/Makefile b/Makefile
index 07eedf03..93a018bc 100644
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,8 @@ GIT = $$( \
 BUILD_DIR?=builddir
 TOOLBOX_DIR?=toolbox
 PROJECT_VER = $$( python3 $(TOOLBOX_DIR)/xnvme_ver.py --path meson.build )
+# Set deployment target to avoid deprecation warnings on macOS
+export MACOSX_DEPLOYMENT_TARGET=11.0

 define default-help
 # invoke: 'make info', 'make tags', 'make git-setup', 'make config' and 'make build'
diff --git a/toolbox/pkgs/macos-11-build.sh b/toolbox/pkgs/macos-11-build.sh
index 4cc34804..50bc21e0 100755
--- a/toolbox/pkgs/macos-11-build.sh
+++ b/toolbox/pkgs/macos-11-build.sh
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
+
+# Set deployment target to avoid deprecation warnings
+export MACOSX_DEPLOYMENT_TARGET=11.0
+
 # configure xNVMe and build dependencies (fio, libvfn, and SPDK)
 meson setup builddir
```

Then you'll just have to rename the `toolbox/pkgs/macos-*` files to `macos.*`. And change `kIOMainPortDefault` back to `kIOMasterPortDefault`.

niclashedam:
I don't understand what "remove support for macOS" means since I am not proposing to stop supporting macOS but rather stopping support for macOS 11 specifically. Within roughly nine months, Apple will also mark Big Sur as EOL. Your solution is not solving the deprecation problems; it's hiding them under the carpet and making the code less correct, as the code does not follow the current Apple documentation.

Since @safl stated that it's OK to drop macOS 11, I prefer not to introduce your proposed change.

toolbox/pkgs/macos-11.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+# This is done to avoid appleframework deprecation warnings
+export MACOSX_DEPLOYMENT_TARGET=11.0
```
safl, line 4: This should be in the ``macos-11-build.sh`` build script instead.

toolbox/pkgs/macos-12-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# configure xNVMe and build dependencies (fio, libvfn, and SPDK)
```
safl, line 2: And also here for macos-12:
```
# This is done to avoid appleframework deprecation warnings
export MACOSX_DEPLOYMENT_TARGET=11.0
```

safl: Since it is in the build-script that the variable is utilized.

lib/xnvme_be_macos_dev.c
```diff
@@ -30,8 +30,8 @@ _get_ioservice_device(char *devname)
 	CFMutableDictionaryRef matching_dictionary;
 	io_object_t ioservice_device;
 
-	matching_dictionary = IOBSDNameMatching(kIOMasterPortDefault, 0, basename(devname));
-	ioservice_device = IOServiceGetMatchingService(kIOMasterPortDefault, matching_dictionary);
+	matching_dictionary = IOBSDNameMatching(kIOMainPortDefault, 0, basename(devname));
```
safl, line 33: With the environment variable set, then I think this should be dropped, as the toolchain for 11 would then not understand the "new" ``kIOMainPortDefault``.

safl (commented):
This is looking great, thanks for those history edits and adjusting according to the feedback on the backwards-compatibility to macOS 11. Very nice! I have left a couple of comments inline that I think might fix the build issue that is popping up now.

niclashedam:
I am not opposing supporting old-non EOL versions. However, I have difficulty doing it if it hurts development and deployment on newer versions. With the suggestions above, we must accept an unnecessary warning on macOS 12+. Alternatively, we'd have to introduce compiler-if statements with different constants based on the target system. For me, neither is acceptable nor good coding discipline.

I hear from both of you that macOS 11 support is essential. That's fair. Therefore, I've instead introduced a new static constant called `darwin_default_master_port`. This constant equals 0 since both `kIOMasterPortDefault`, and `kIOMainPortDefault` are synonyms for 0. Ultimately, this makes all happy. We have macOS 11 support, and I don't feel like we are hurting macOS 12+. When Big Sur becomes EOL, we can return to `kIOMainPortDefault`.

safl:
> I am not opposing supporting old-non EOL versions. However, I have difficulty doing it if it hurts development and deployment on newer versions. With the suggestions above, we must accept an unnecessary warning on macOS 12+. Alternatively, we'd have to introduce compiler-if statements with different constants based on the target system. For me, neither is acceptable nor good coding discipline.
> 
> I hear from both of you that macOS 11 support is essential. That's fair. Therefore, I've instead introduced a new static constant called `darwin_default_master_port`. This constant equals 0 since both `kIOMasterPortDefault`, and `kIOMainPortDefault` are synonyms for 0. Ultimately, this makes all happy. We have macOS 11 support, and I don't feel like we are hurting macOS 12+. When Big Sur becomes EOL, we can return to `kIOMainPortDefault`.

I think @Baekalfen suggestion can work without the ``darwin_default_master_port`` e.g. without changing the deprecated code. Added an experiment here: https://github.com/OpenMPDK/xNVMe/pull/215/commits
It is your PR (#212) with the following changes:
* Dropped the first commit (be/macos: adjust to macOS API changes)
* Added:
```
# This is done to avoid appleframework deprecation warnings
export MACOSX_DEPLOYMENT_TARGET=11.0
```
to Makefile and build-scripts.

Basically, instead of if-def'ing, or in other ways, handling those deprecation warnings in the code, then the tool-chain is instead instructed to treat the code as "v11" and "somewhat" described in the documentation.

safl (commented):
This is looking mighty fine, the last final thing is the changes experimented with on PR #215 
What do you think of those? They seem to do the trick:
https://github.com/OpenMPDK/xNVMe/actions/runs/3543500907/jobs/5950031610

Baekalfen:
> This is looking mighty fine, the last final thing is the changes experimented with on PR #215
> What do you think of those? They seem to do the trick:
> https://github.com/OpenMPDK/xNVMe/actions/runs/3543500907/jobs/5950031610

I think that looks good. We should have set the deployment target regardless. Normally Xcode would do this for you when you create a project on macOS.

niclashedam:
I've pulled in the changes from #215. However, I've removed the export in the build bash file since having it in the Makefile is better. We want the CI to resemble reality as much as possible, and having it in the Makefile should be sufficient on its own.

safl:
@niclashedam The motivation for putting it into the the ``*build.sh`` **and** into the Makefile are:

* The ``*build.sh`` are for the documentation of build-instructions provided here:
  - https://xnvme.io/docs/next/getting_started/index.html#macos
  - additionally, then they are used by the CI system, such that the documentation for the common-case of building and installing xNVMe is in itself tested, thus the instructions won't go stale, since that would mean that the CI would fail.
* The Makefile is a "helper" showing how to instrument ``meson`` and useful as a "shorthand" for various ``meson`` configurations and instrumenting additional tools from the ``toolbox``

So, please have it in both places :)

niclashedam:
Alright, gotcha. They are in the PR again.

toolbox/pkgs/macos-12-build.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# This is done to avoid appleframework deprecation warnings
+export MACOSX_DEPLOYMENT_TARGET=12.0
```
safl, line 4: This should be ``11.0`` to tell the toolchain that the code follows the old API, otherwise it will yield deprecation warnings here.

safl (commented):
Almost there :)

niclashedam:
Updated!

safl (approved):
Thanks!

PR #216, opened 2022-11-28
be/spdk: add transport ack timeout to fabrics

### LINKED ISSUE #130: xnvme_queue_drain hangs when connection is dropped in spdk backend
When I try to make io commands while the connection to the spdk device is dropped the function xnvme_queue_poke hangs, this happens in the spdk backend but not when using nvme-cli. The behavior I expect is that the function should return an error that I would be able to handle on my side.

when using nvme-cli this is what I get in the debug log
```
# DBG:xnvme_be_linux_nvme.c:ioctl_wrap-137: INFO: ioctl(NVME_IOCTL_IO64_CMD), err(-1), errno(11)
# DBG:xnvme_be_linux_nvme.c:ioctl_wrap-144: INFO: retconv: -1 and set errno
# DBG:xnvme_be_linux_nvme.c:xnvme_be_linux_nvme_cmd_io-255: FAILED: ioctl_wrap(), err: -11
# DBG:xnvme_be_posix_async_emu.c:_posix_async_emu_poke-136: FAILED: sync.cmd_io{v}(), err: -11
# ERR: 'got completion errors': {errno: 5, msg: 'Input/output error'}
```

Here is how to reproduce the issue:
1) create spdk device
```
./build/bin/nvmf_tgt &
./scripts/rpc.py nvmf_create_transport -t TCP -u 16384 -m 8 -c 8192
./scripts/rpc.py nvmf_create_subsystem nqn.2016-06.io.spdk:cnode1 -a -s SPDK00000000000001 -d SPDK_Controller1
./scripts/rpc.py nvmf_subsystem_add_listener nqn.2016-06.io.spdk:cnode1 -t tcp -a  127.0.0.1 -s 4420
./scripts/rpc.py bdev_null_create Null0 8589934592 4096
./scripts/rpc.py nvmf_subsystem_add_ns nqn.2016-06.io.spdk:cnode1 Null0
```

2) run examples/xnvme_single_async (change opts.nsid to 1) and stop with a breakpoint at line 96 (before issuing the command)
3) configure iptables to drop incoming connection
```
sudo iptables -I INPUT -p tcp -i lo --dport 4420 -j DROP
```
Then xnvme_queue_drain should hang

To restore the ip tables run:
```
sudo iptables -D INPUT -p tcp -i lo --dport 4420 -j DROP
```

karlowich:
Resolves #130 

This is to avoid a situation where xnvme hangs because the fabrics connection is dropped.

This solution was chosen over
spdk_nvme_ctrlr_register_timeout_callback(),
since it calls the completion callback function when it fails and allows the user to handle the error as they see fit.

The downside is that the user will potentially have to wait for the timeout twice, once when checking for command completion and again when closing the device.

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -380,6 +380,7 @@ _spdk_setup_controller_opts(struct xnvme_opts *opts, const struct spdk_nvme_tran
 		ctrlr_opts->header_digest = 1;
 		ctrlr_opts->data_digest = 1;
 		ctrlr_opts->keep_alive_timeout_ms = 0;
+		ctrlr_opts->transport_ack_timeout = 16;
 		if (opts->hostnqn) {
 			strncpy(ctrlr_opts->hostnqn, opts->hostnqn, SPDK_NVMF_NQN_MAX_LEN);
 		}


karlowich:
As mentioned here, #130, this doesn't handle every case. Therefore, #135 will be used.

PR #217, opened 2022-11-28
toolbox/xnvme-driver: adapt syntax to bash 2.5.3

karlowich:
xnvme-driver fails on FreeBSD in the CI, this is because FreeBSD is using bash 2.5.3. Which has deprecated the use of parentheses around brackets. Every instance of `([0-9])` has been changed to `[0-9]`


safl (approved):
LGTM!

safl:
@karlowich thanks

PR #221, opened 2022-12-20
Add FDP changes in xNVMe

ankit-sam:
This series adds FDP support in xNVMe.
The initial 2 commits modify the log pages and set / get feature commands by adding a prep function. This is done to avoid changes to the existing api so that it does not become too bulky.
The next 4 commit introduces the required changes in spec, tools, cli and api.

This series is missing changes to toolbox (test) which I will update once Qemu changes are upstream

### DIFF
--- a/include/libxnvme_adm.h
+++ b/include/libxnvme_adm.h
@@ -83,6 +83,20 @@ xnvme_adm_idfy_ns_csi(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint8_t csi,
 		      struct xnvme_spec_idfy *dbuf);
 
 /**
+ * Prepare NVMe Get Log Page command
+ *
+ * @param ctx Pointer to ::xnvme_cmd_ctx
+ * @param lid Log Page Identifier for the log to retrieve entries for
+ * @param lsp Log Specific Field for the log to retrieve entries for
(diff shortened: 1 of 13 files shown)


karlowich:
This is looking good.
I have fixed the issue causing the test to fail here: #237

include/libxnvme_adm.h
```diff
@@ -100,6 +114,31 @@ int
...
+xnvme_prep_adm_gfeat(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint8_t fid, uint8_t sel);
+
+/**
+ * Prepare NVMe Set Feature (sfeat) command
+ *
+ * @param ctx Pointer to ::xnvme_cmd_ctx
+ * @param nsid Namespace identifier
+ * @param fid Feature identifier (see NVMe 1.3; Figure 84)
+ * @param feat Structure defining feature attributes
+ * @param save
```
karlowich, line 136: I think an explanation could be good here. Maybe something like:
"specify that the controller shall save the attribute"

ankit-sam: Sure, will update this

safl (approved):
Thanks @ankit-sam, this LGTM!

PR #224, opened 2023-01-11
py/cython-bindings: handle const qualifiers in function signatures

karlowich:
autopxd2 v2.3.0 added support for `const` qualifiers. These caused the auto-generation of the `.pyx` files to fail. This has been resolved now.
Unfortunately, this does not resolve any of the `[-Wdiscarded-qualifiers]` warnings.

Currently the only `const` qualifiers that end up in the `.pyx` files are `const char *`, this is because python/cython complains if I put them in front of pyObjects, i.e. `xnvme_void_p`. 
Unless we can find a way to do that, perhaps it is better to just discard the const qualifiers instead, to keep the complexity of `autopxd_py.py` down.

@Baekalfen @safl can I get your opinion on this?

### DIFF
--- a/python/xnvme-cy-bindings/auxiliary/autopxd_py.py
+++ b/python/xnvme-cy-bindings/auxiliary/autopxd_py.py
@@ -54,7 +54,7 @@ def get_definitions(pxd_f):
 
         args = []
         for arg in re.finditer(
-            r"([_a-zA-Z0-9\*]+) ([_a-zA-Z0-9]+)(?:, )?", arg_body.strip("()")
+            r"(?:const )?([_a-zA-Z0-9\*]+) ([_a-zA-Z0-9]+)(?:, )?", arg_body.strip("()")
         ):
             arg_type, arg_name = arg.groups()
             args.append((arg_type, arg_name))


python/xnvme-cy-bindings/auxiliary/autopxd_py.py
```diff
@@ -53,11 +53,14 @@ def get_definitions(pxd_f):
         _type, __func_name, func_name, arg_body = func.groups()
 
         args = []
-        for arg in re.finditer(
-            r"([_a-zA-Z0-9\*]+) ([_a-zA-Z0-9]+)(?:, )?", arg_body.strip("()")
```
Baekalfen, line 57: I think the best solution would be to add `(?:const )?` to the beginning of the regex, and keep status quo.

Baekalfen, line 57: I think the best solution would be to add `(?:const )?` to the beginning of the regex, and keep status quo.

Baekalfen (changes requested):

python/xnvme-cy-bindings/auxiliary/autopxd_py.py
karlowich: I agree

karlowich: Done

Baekalfen: Beautiful 👍

Baekalfen (approved):

safl (approved):
Great, thanks!

PR #236, opened 2023-01-18
be/spdk: remove patches for verify and simple copy

karlowich:
These patches should be removed in order to get closer to the goal of not bundling SPDK.

The functionality is very minor and it is not part of any tests.


safl (approved):
Thanks for your relentless effort on this.
Let's see how far we can get tomorrow with deprecating that SPDK bundling!

PR #237, opened 2023-01-20
ci,toolbox/cijoe: Don't try to save non-saveable feat

karlowich:
The test is trying to set a Temperature Threshold, which is a non-saveable feature. As such, it should not try to save it.

The only reason this is not failing currently is because the `xnvme` cli is hardcoded to never save. However, #221 changes this to be implemented correctly and then the test fails.


safl (approved):
Great, thanks!

PR #240, opened 2023-01-27
ci: Switch containers from refenv to ghcr

karlowich:

PR #241, opened 2023-01-30
precommit: bump isort to 5.12

karlowich:


karlowich:
This is included here: #240, since the two PRs are solving the problem the other is having

PR #242, opened 2023-02-01
xnvmec: rename xnvme_enumeration* to xnvmec_enumeration*

karlowich:
xnvme_enumeration belongs to libxnvmec, thus it has the wrong prefix. This commit corrects this prefix.


safl (approved):
Thanks!

PR #243, opened 2023-02-02
meson: don't bundle SPDK

karlowich:
This removes the bundling of SPDK. Instead SPDK is now a regular dependency that needs to be installed independently.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -11,12 +11,15 @@ examples_source = [
   'zoned_io_sync.c',
 ]
 
+xnvme_dep = get_variable('lib_' + get_option('default_library'))
+
 foreach example_source : examples_source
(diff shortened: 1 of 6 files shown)


karlowich:
This now works with shared libraries, but it requires installing:
DPDK v22.11
SPDK v22.09

If on runtime, the loader is missing some shared library run:
`export LD_LIBRARY_PATH=/usr/local/lib/`

safl:
This is really awesome! The need to set ``LD_LIBRARY_PATH`` will likely vary based on Linux Distribution, similar to how PKG_CONFIG and other things may-or-may-not need to be set.

karlowich:
This leads to a deadend.
#256 will solve the problem instead.

PR #245, opened 2023-02-13
rpm: Add rpm spec file and support with Makefile

sc108-lee:


meson.build
```diff
@@ -107,6 +107,16 @@ if is_windows
   conf_inc += [ include_directories('third-party'/'windows') ]
 endif
 
+rpm_conf = configuration_data()
+rpm_conf.set('NAME',    meson.project_name())
+rpm_conf.set('VERSION', meson.project_version())
+rpm_conf.set('LICENSE', meson.project_license())
```
karlowich, line 113: This line seems to be the one breaking the tests.
`meson.project_license()` returns an array of strings and the `rpm_conf.set()` expects a single string.

eli-schwartz: Having more than one is deprecated. You could index into it and only use the first one, safe in the assumption that there's no additional information being dropped.

sc108-lee: @karlowich @eli-schwartz Thanks for the comment. 
I just added using first index of license

safl (approved):
Thanks @sc108-lee !

PR #247, opened 2023-02-17
meson: only check for CentOS 7 if sys == "linux"

karlowich:
Currently this check is run regardless of the system, however on some systems the command either fails or doesn't exist. This can cause the build to fail.

Instead this should only be run if the system is linux.


safl (approved):
Thanks

PR #256, opened 2023-02-27
Refactor build of xNVMe

karlowich:
A refactoring of the build system for xNVMe is required to get closer to the goal of distributing xNVMe in the native package system of linux distros, like Fedora and Debian.

The main goal is to move away from bundling libraries into xNVMe by default. Instead, xNVMe should produce a "clean" library by default and only allow for bundling when building from source.
A consequence of this is that SPDK will only be available when using the bundled library built from source.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -424,6 +424,13 @@ jobs:
       run: |
         if [[ -f "${BSCRIPT}" ]]; then source ${BSCRIPT}; else source ${BSCRIPT_DEF}; fi
 
+    - name: Execute 'ldconfig'
+      # ldconfig doesn't work on alpine
+      # centos7 needs "ldconfig /usr/lib/local" for python3 to work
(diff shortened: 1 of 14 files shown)


safl:
Yesterday the self-hosted xNVMe github-actions-runners were "migrated" from one hosting provider to another.
Consequently, then the changes described on https://github.com/OpenMPDK/xNVMe/pull/276 were needed.

To avoid blocking progress on this PR (https://github.com/OpenMPDK/xNVMe/pull/256), due to CI-failures, then this PR was rebased.

Curiously, the rebase of this PR shows in a surprising way on the history. That is, the two commits introduced on https://github.com/OpenMPDK/xNVMe/pull/276 looks like they are part of the PR, as the first two commits. Even though they have the same SHAs as the commits on ``next``. It seems inconsequential, e.g. something cosmetically wrong in the GitHUB UI, regardless, writing this to let you know.

safl (approved):
Awesome!

PR #260, opened 2023-03-08
be/linux: Use io_uring_free_probe instead of free

Joelgranados:
Liburing uses nolibc by default in some architecture. This means that it might have its own version of free. Use the liburing free version for the uring probe so we do not run into incompatibilities

### DIFF
--- a/lib/xnvme_be_linux_async_ucmd.c
+++ b/lib/xnvme_be_linux_async_ucmd.c
@@ -41,7 +41,7 @@ _linux_liburing_noptional_missing(void)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


karlowich (approved):
Looks good!

Joelgranados:
There are some failures on the CI but it seems that it is because of misconfigured test systems. The messages that I see are regarding the failure to find 'cc' in the environment. could my patch have broken this?????

safl (approved):
Thanks @Joelgranados.
The CI has some issues which are being worked out. Once they are fixed, then this goes in.

PR #269, opened 2023-03-20
api,tools: replace 'nlb' with 'llb' in dsm-range

safl:
The length-in-logical-blocks inside the dsm-range is a one-based value. However, it is named 'nlb', which is used everywhere else for zero-based values. This is a short-coming of the specification not defining a name for the field, thus inventing 'llb' for this field.

The PR also has a commit removing some deprecated files for pre-commit/clang-format. This commit is piggy-backed on this PR to spend less time in the CI-queue for this trivial removal.


ankit-sam (approved):
Looks fine to me, I did some basic testing with qemu as well as on actual hardware.

safl:
Thanks @ankit-sam !

PR #271, opened 2023-03-22
Jag/fix dsm command

Joelgranados:


lib/xnvme_nvm.c
```diff
@@ -123,7 +123,7 @@ xnvme_nvm_dsm(struct xnvme_cmd_ctx *ctx, uint32_t nsid, struct xnvme_spec_dsm_ra
 	ctx->cmd.dsm.idw = idw;
 	ctx->cmd.dsm.idr = idr;
 
-	return xnvme_cmd_pass(ctx, (void *)dsm_range, sizeof(*dsm_range) * nr, NULL, 0);
+	return xnvme_cmd_pass(ctx, (void *)dsm_range, sizeof(*dsm_range) * (nr + 1), NULL, 0);
```
safl, line 126: Nice, catch!

tools/xnvme.c
```diff
@@ -656,7 +656,7 @@ sub_dsm(struct xnvmec *cli)
 	dsm_range->llb = llb;
 	dsm_range->slba = slba;
 
-	err = xnvme_nvm_dsm(&ctx, nsid, dsm_range, nr, ad, idw, idr);
+	err = xnvme_nvm_dsm(&ctx, nsid, dsm_range, nr - 1, ad, idw, idr);
```
safl, line 659: i would prefer that this be left as is, that is just pass ``nr``.
Then, instead change the value of  ``nr`` in its initialization, and possibly add a comment as well.

safl: Up at around line 630.

lib/xnvmec.c
```diff
@@ -1561,13 +1561,13 @@ xnvmec_assign_arg(struct xnvmec *cli, struct xnvmec_opt_attr *opt_attr, char *ar
 		args->nsr = num;
 		break;
 	case XNVMEC_OPT_AD:
-		args->ad = arg ? num : 0;
+		args->ad = arg ? num : 1;
```
safl, line 1564: Generally, then the values parsed from the cli are 0 by default.
This would flip, such that arguments which are not provided will carry the value 1 by default.

In case it should be restricted to the set of values 1 or 0, then how about changing this to:
```
args->idw = arg ? 1 : 0;
```

Or, am I missing something here?

safl (commented):
Thanks!
The way I read this, then this fixes an issue in ``lib/xnvme_nvm.c``.
The other changes I have a couple of comments to.

tools/xnvme.c
safl: Something like noting ``nr = 0; ///< Zero-based value``

Joelgranados: If you do this, you need to change line 648 to `dsm_range = xnvme_buf_alloc(dev,sizeof(*dsm_range) * (nr +1));`. Does it still make sense to you @safl ?

lib/xnvmec.c
Joelgranados: I followed what was done for XNVMEC_OPT_HELP which is also one that does not carry an argument. Should we change that one to `arg ? 1 : 0` as well then?

tools/xnvme.c
Joelgranados: Actually, why is there an `nr` variable to begin with. As I read it, this function will only send a dsm for one range. So it would be better to just remove the `nr` variable and pass a 0 to the xnvme_nvm_dsm call.

safl: I agree. I guess I lean towards your latter approach of removing the ``nr`` and just passing 0.

lib/xnvmec.c
safl: Hmm, I need to take another look, seems like there is something I have forgotten about the cli-parser code here.
As I recall, then point of ``args->attribute = arg ? num : k;`` is that the logic can be used for arguments which are used as either value-options or flag-options. When used as flags, then I have forgotten why there is a discrepancy in the code-base of using:

```
args->ad = arg ? num : 1;
```
in some cases, and:
```
args->ad = arg ? num : 0;
```
In others... I would expect the latter form to be "correct", that is, default to 0.

safl: One approach here, is make the change that works for you for the ``dsm`` functionality, these are only available on ``next`` not in any release. Thus, their behavior will not break anything. Then, I or you could write an issue about the need to investigate this and cleanup/refactor the code for proper common operation.

safl: What I mean by this is:

* Add the cli-parsing that you need
* Don't change the behavior of the other arguments (to avoid regression)
* Open an issue, to clean the argument parser "up" / re-factor

Joelgranados: Just gave the code a quick run and here is what I understand is happening:
1. With XNVMEC_LFLG type arguments arg is 0x0 (because there is no accompanying  arg) and num is always 0 because the code that sets it is skipped in line 1207.
2. Every time you are in a case of a XNVMEC_LFLG type arg you should just consider it being in the arguments

IMO the refactoring should be changing all the XNVMEC_LFLG type args to args->ARG_NAME = 1; or whatever should happen when the arg is present in the command line.

Joelgranados: > What I mean by this is:
> 
> * Add the cli-parsing that you need
OK

> * Don't change the behavior of the other arguments (to avoid regression)
Do you mean IDW and IDR here?  I'm ok with that, but they will be broken if you don't do anything.

> * Open an issue, to clean the argument parser "up" / re-factor
I can do that know if you want.

PR #275, opened 2023-04-03
Windows: SPDK backend support 

rs-sam:
Adding support for SPDK backend for Windows OS, using WPDK repo.
Have a look at this initial work to do support SPDK backend for xNVMe on Windows OS.

Created a new subproject called 'spdk-win'. It used mix environment MSYS2-MinGW and WSL for the compilation.
A helper 'spdk_win_build.bat' batch file is used to do compilation of dependency repos like
WPDK, DPDK and SPDK.


safl:
Yesterday the self-hosted xNVMe github-actions-runners were "migrated" from one hosting provider to another.
Consequently, then the changes described on https://github.com/OpenMPDK/xNVMe/pull/276 were needed.

To avoid blocking progress on this PR (https://github.com/OpenMPDK/xNVMe/pull/275), due to CI-failures, then this PR was rebased.

Curiously, the rebase of this PR shows in a surprising way on the history. That is, the two commits introduced on https://github.com/OpenMPDK/xNVMe/pull/276 looks like they are part of the PR, as the first two commits. Even though they have the same SHAs as the commits on ``next``. It seems inconsequential, e.g. something cosmetically wrong in the GitHUB UI, regardless, writing this to let you know.

safl:
HI @rs-sam I rebased this, to get the changes to work around the ci-blockage.

rs-sam:
I have rebased this PR with latest next. It has a new issue that is latest SPDK repo has a non windows friendly file name "https://github.com/spdk/spdk/blob/master/test/common/config/pkgdep/patches/ice/0001-stats-fetch_*_irq.patch"
due to this spdk-win.wrap not able to clone it properly, looking for any workaround for this.

subprojects/spdk-win.wrap
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+[wrap-git]
+url = https://github.com/spdk/spdk.git
+revision = v22.05
+patch_directory = spdk-win
+clone-recursive = true
```
karlowich, line 9: I believe we can avoid the issue if we use the following instead of `[wrap-git]`:
```
[wrap-file]
source_url = https://github.com/spdk/spdk/archive/refs/tags/v22.05.tar.gz
source_filename = spdk-22.05.tar.gz
```

safl: Looks like there is an issue with the tar-all as it does not include the git-submodules e.g. DPDK etc.

subprojects/packagefiles/wpdk/patches/wpdk-0001-fixed-clang-compilation-warning.patch
```diff
@@ -0,0 +1,261 @@
+From 9fa031ce3564b3d6902b3cc7822325379eb9c017 Mon Sep 17 00:00:00 2001
```
safl: I think these could/should be upstreamed to SPDK/WPDK. Looks like good improvements.

- [ ] Upstream patch fixing compilation-warnings

meson.build
```diff
@@ -93,10 +93,11 @@ conf_data.set('XNVME_BE_MACOS_ENABLED', is_darwin)
 conf_data.set('XNVME_BE_FBSD_ENABLED', is_freebsd)
 
 conf_data.set('XNVME_BE_SPDK_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
```
safl, line 95: Here you could now do:

```
conf_data.set('XNVME_BE_SPDK_ENABLED', get_option('with-spdk') and (!is_darwin))
```

meson.build
```diff
@@ -93,10 +93,11 @@ conf_data.set('XNVME_BE_MACOS_ENABLED', is_darwin)
...
-conf_data.set('XNVME_BE_SPDK_TRANSPORT_PCIE_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
+conf_data.set('XNVME_BE_SPDK_TRANSPORT_PCIE_ENABLED', (is_linux or is_freebsd or is_windows) and (get_option('with-spdk') or get_option('with-spdk-win')))
 conf_data.set('XNVME_BE_SPDK_TRANSPORT_TCP_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
 conf_data.set('XNVME_BE_SPDK_TRANSPORT_RDMA_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
 conf_data.set('XNVME_BE_SPDK_TRANSPORT_FC_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
+conf_data.set('XNVME_BE_SPDK_WIN_ENABLED', (is_windows) and get_option('with-spdk-win'))
```
safl, line 100: I think you do not need an additional option? How about just re-using ``with-spdk``?

safl: Would this:
``
XNVME_BE_SPDK_WPDK_ENABLED
``
be preferable to:
``
XNVME_BE_SPDK_WIN_ENABLED
``
?

meson_options.txt
```diff
@@ -1,6 +1,7 @@
 option('with-liburing', type: 'boolean', value: true)
 option('with-libaio', type: 'boolean', value: true)
 option('with-spdk', type: 'boolean', value: true)
+option('with-spdk-win', type: 'boolean', value: true)
```
safl, line 4: I think you can drop this one.

safl (changes requested):
To avoid multiple SPDK subprojects e.g. ``subprojects/spdk`` and ``subprojects/spdk-win`` then it would be great to combine these into a single subproject.
However, ss I understand it then WPDK currently uses v22.05 and xNVMe uses v22.09.
Thus, we need to track to different SPDK versions. Let's make sure this is documented, and then we should work on updating both xNVMe and WPDK to the latest SPDK version ``v23.??``.

safl (changes requested):
Code LGTM, however, we seem to face some CI issues with this.
I would propose:

- [ ] Remove call of ``xnvme enum``
- [ ] Fix what is causing ``xnvme library-info`` to fail
- [ ] Drop support / testing for Windows 2019
  - [ ] Remove from .gtihub workflow, documentation, etc.
- [ ] Determine whether this can be built using ``clang`` instead of ``gcc``
  - Compilation issue needs to be resolved
  - advantage: "direct" use of ``clang`` + ``meson``
  - advantage: we would not need to use **wsl**
  - advantage: a lot of scripts patching dos2unix etc. would no longer we needed

safl (approved):

subprojects/packagefiles/spdk-win/meson.build
```diff
@@ -0,0 +1,95 @@
...
+  version: '22.05',
+)
+fs = import('fs')
+
+if get_option('build_subprojects') and not fs.exists('build' / 'lib' / 'libspdk_nvme.a')
+  message('Patching ..')
+  run_command('spdk_win_patches.sh', capture: true, check: true)
+  message('Configuring and building..')
+  run_command('spdk_win_build.bat', capture: true, check: true)
+endif
```
karlowich, line 13: It seems that there is a big overlap between the spdk-win `meson.build` and the regular SPDK `meson.build`.
I think it would be great if we could just have the single spdk subproject, then you can add the few things you need on windows using if checks for: `is_windows`.
Do you think this is possible?

karlowich (commented):
Everything looks good, I just have a single comment

subprojects/packagefiles/spdk-win/meson.build
safl: @karlowich I think the issue is that we need different **revisions** of the SPDK repository. Is that correct @rs-sam ?

karlowich: @safl Oh you're right. That is a soft requirement from WPDK, that is super annoying.

karlowich: In that case, I approve of this PR

karlowich (approved):

PR #279, opened 2023-04-11
rpmbuild: fix directory conflicts error

sc108-lee:
when directory is created in some other package it occur below error file /usr/local/lib64/pkgconfig from install of xnvme-0.6.0-0.x86_64 conflicts

Add executable permission for library, bin files


safl (approved):
Thanks @sc108-lee

PR #282, opened 2023-04-12
be/spdk: remove unnecessary patch

karlowich:
Now that we are building spdk correctly and no longer bundling. We do not need to force driver registrations.


safl (commented):
Exciting :) With this gone, then the following code calling the driver-hook should be removed as well:

```c
#ifdef SPDK_REGHOOK_NVME_PCIE_DRIVER
	XNVME_DEBUG("INFO: SPDK NVMe PCIe Driver registration -- BEGIN");
	if (!spdk_nvme_transport_available_by_name("PCIe")) {
		XNVME_DEBUG("INFO: registering...");
		spdk_reghook_nvme_pcie_driver();
	} else {
		XNVME_DEBUG("INFO: skipping, already registered.");
	}
	XNVME_DEBUG("INFO: SPDK NVMe PCIe Driver registration -- END");
#endif
```

That is in ``xnvme_be_spdk_dev.c``.

Also, perhaps change the short-message to: ``be/spdk: remove pcie-driver-registration patch`` since it describes what is removed, rather than a trait (unnecessary) of the patch. Then the description that follows describes that the patch is unnecessary and why it is removed.

safl:
Hi @karlowich , i have added a test on top, everything looks great. Plrase have a look.

safl (approved):
Thanks @karlowich, this LGTM.
Note, that I have added a test verifying: dynamic loading, linking shared + static.

In case you disagree with something in the tests, then I can fix it up, otherwise I find this is good to go.

cijoe/tests/linkandload/test_xnvme_consumption.py
```diff
@@ -0,0 +1,103 @@
...
+        Path(libdir) / line
+        for line in state.output().split()
+        if "libxnvme" in Path(line.strip()).stem
+    ]
+
+    return {suffix[lib.suffix]: lib for lib in paths}
+
+
+def test_library_consumption(cijoe):
+    """Verify that the shared library is consumable via dynamic loading"""
```
karlowich, line 63: This seems to be inaccurate. It tests all 3 types right?

cijoe/tests/linkandload/test_xnvme_consumption.py
```diff
@@ -0,0 +1,103 @@
+"""
+This tests whether:
+
+* It is possible to load the shared library
+* It is possible to link with the shared library
+* It is possible to link with the static library
+
+It does so by compiling programs and running them, and additionally, checking
```
karlowich, line 8: It looks like it is not actually running the compiled programs. Either that should be implemented or the docstring should be updated

karlowich (commented):
This is looking good, I have left couple comments.

Also what is the purpose of the "mini.toml"?

cijoe/tests/linkandload/test_xnvme_consumption.py
safl: Right! Nice catch. Re-phrased!

cijoe/tests/linkandload/test_xnvme_consumption.py
safl: Right, now it ought to actually run them....

safl:
Thanks!

PR #283, opened 2023-04-12
build/libvfn: Enable support for external libvfn

karlowich:
Now Meson is able to find a system-installed libvfn and compile with that instead of the subproject.

This relies on Meson actually being able to find the system-installed libvfn. If it cannot find it, the subproject will be built and installed. This can lead to Meson accidentally overwriting the system-installed libvfn as in the example below:

* libvfn is installed in /usr/local/lib
* /usr/local/lib/pkgconfig is not part of the PKG_CONFIG_PATH
* During installation of xNVMe Meson cannot find libvfn
* Meson installs xNVMe and the libvfn subproject to /usr/local/lib

This was also a problem before this patch.
However, since we used to always install the libvfn subproject it was consistent. Ideally, we could disable installation of the libvfn subproject, but this is not currently supported in Meson.


eli-schwartz:
> Ideally, we could disable installation of the libvfn subproject, but this is not currently supported in Meson.

You could disable installation of the subproject with current versions of meson, with a hack. Update the libvfn build files to conditionally if/else functions such as `install_*()`, and for functions with an `install: <bool>` kwarg, pass a conditional variable instead of just "true".

The conditional could be an option defined in meson_options.txt or just be `not meson.is_subproject()` on the theory that one never wants to install libvfn when it is merely vendored into another project (probably correct).

safl:
> > Ideally, we could disable installation of the libvfn subproject, but this is not currently supported in Meson.
> 
> You could disable installation of the subproject with current versions of meson, with a hack. Update the libvfn build files to conditionally if/else functions such as `install_*()`, and for functions with an `install: <bool>` kwarg, pass a conditional variable instead of just "true".
> 
> The conditional could be an option defined in meson_options.txt or just be `not meson.is_subproject()` on the theory that one never wants to install libvfn when it is merely vendored into another project (probably correct).

Thanks for the input @eli-schwartz , how recent a meson version would this require?

karlowich:
@safl I have implemented the changes @eli-schwartz suggested. However, now the source-archive won't build because there are uncommitted changes in `subprojects/libvfn` from the patch. There are also uncommited changes in `subprojects/spdk` but these don't cause issues. Do you have an idea what the difference is / how to fix it?

safl:
> @safl I have implemented the changes @eli-schwartz suggested. However, now the source-archive won't build because there are uncommitted changes in `subprojects/libvfn` from the patch. There are also uncommited changes in `subprojects/spdk` but these don't cause issues. Do you have an idea what the difference is / how to fix it?

Hmm. The main difference is that  ``subprojects/spdk`` is a not a meson project, and thus not a "proper" subproject. Consequentially, then spdk is patched and built at configuration-time, whereas ``subprojects/libvfn`` is a "proper" meson project and is built at built-time. They are probably handlded differently by meson in many ways. The patching done on SPDK is also not "proper", the right meson-way would be some "code-generator" logic before the build-stage.
Just a couple of thoughts, I don't have an immediate idea on how to address the issue.

karlowich:
@safl 
> The patching done on SPDK is also not "proper"

I have changed the way we patch SPDK to use meson functionality, so there's no difference there.
Perhaps, the solution is to just enable "--allow-dirty"?
That would also make it more comfortable to use during development.

safl:
> 



> @safl
> 
> > The patching done on SPDK is also not "proper"
> 
> I have changed the way we patch SPDK to use meson functionality, so there's no difference there. Perhaps, the solution is to just enable "--allow-dirty"? That would also make it more comfortable to use during development.

Hmm, how about sending these changes to libvfn: https://github.com/OpenMPDK/libvfn
See what @birkelund thinks about these changes. Then there is no need to do the "patching".

I am a bit nerved by the bump of meson to 0.63, for projects using xNVMe on embedded systems and such...

eli-schwartz:
Sorry, to be clear my original intention was to make the canonical libvfn support this out of the box.

I actually forgot that diff_files existed, since I never use them and I'm not the one who implemented them. I'm also skeptical about the value of having meson support build overlays / patches on a git repository. :P

It's entirely possible that the combination of "git repo", "uses overlays/patches" and "try to run the dist tarball creation command" is just broken. If so, that's @xclaesse's fault. :D

birkelund:
meson >=0.58 has --skip-subprojects for install. I'm not too happy about sprinkling "not subproject" guards everywhere in my meson build files.

xclaesse:
> You could disable installation of the subproject with current versions of meson, with a hack. Update the libvfn build files to conditionally if/else functions such as `install_*()`, and for functions with an `install: <bool>` kwarg, pass a conditional variable instead of just "true".

That's not a good advice at all, don't patch meson.build files like that.

There is `meson install --skip-subprojects` if you really want to skip all/some subprojects from installation. But what you probably want is to skip installation of devel files (headers, static libs, etc) in which case I would suggest using `meson install --tags runtime` because subprojects might still need to install some runtime files even if you static link them. That depends on why you want to skip installation of libvfn?

eli-schwartz:
I did say it was a hack.

The problem with --skip-subprojects is that whether or not to do so is properly a choice to be made by the top-level project, not a choice of the person doing the build (who surely expects things to be non-broken by default, rather than an opt-in). Unfortunately there are differing ideas of what subprojects are supposed to be for, so the "solution" we got instead was a manual one.

eli-schwartz:
> in which case I would suggest using meson install --tags runtime because subprojects might still need to install some runtime files even if you static link them. 

Maybe if I say it a few thousand more times, you'll eventually realize I'm telling the truth?

Skipping subprojects isn't a solution. Only installing runtime files isn't a solution. The former won't install runtime files from the subproject, and the latter won't install development files from the parent project.

And when you're doing wrap fallbacks, in most cases you are doing fallback of a C/C++ library and you know, beyond all shadow of a doubt, that there's no such thing as runtime files.

Period. End of story. Runtime doesn't exist.

GLib special snowflake cases can be handled by, well, doing the meson.is_subproject() dance for exactly devel files and nothing else. :) :) :) :)

...

Or we could actually solve it for real in meson but that will apparently never happen.

xclaesse:
> Maybe if I say it a few thousand more times, you'll eventually realize I'm telling the truth?

Repeating that does make it true.
 
> Only installing runtime files isn't a solution. (...) won't install development files from the parent project.

Which is the whole point. You should not install devel files from parent project without devel from subprojects. xnvme.pc has `Requires: libvfn`. If you skip install of devel files from libvfn you break xnvme.pc file. Even if libvfn is an internal dependency not exposed in libxnvme API, you still need libvfn.a to link with libxnvme.a.

> GLib special snowflake cases

What?

toolbox/pkgs/emitter/templates/src-libvfn.sh.jinja
```diff
@@ -0,0 +1,12 @@
+#
+# Clone, build and install libvfn
+#
+# Assumptions:
+#
+# - These commands are executed with sufficient privileges (sudo/root)
+#
+git clone https://github.com/OpenMPDK/libvfn.git
```
birkelund, line 8: This should probably checkout the right version of libvfn and not master.

karlowich: Good point, fixed!

safl:
@karlowich I dropped the most recent commit as it did not live up to the promise of: ``# Avoid error with "libpython3.7m.so.1.0: cannot open shared object file: No such file or directory"``.  The PR is otherwise great, so about to force push and merge this in to get the pipeline going again.

safl:
This currently fails due to issue with cython etc. thus, expediting the changes on: https://github.com/OpenMPDK/xNVMe/pull/287 As these fix this in the consolidation into a single Python offering.

safl (approved):
This currently fails due to issue with cython etc. thus, expediting the changes on: https://github.com/OpenMPDK/xNVMe/pull/287 As these fix this in the consolidation into a single Python offering. Found a couple of other minor things and put fixes on top. Will add it to ``next`` upon success. Thanks for this @karlowich !

PR #285, opened 2023-05-01
be:spdk: Add option to control admin and io command timeout

LiadOz:
We have found that for certain applications, we would want to change/remove the timeout on admin commands and io commands. This commit allows to choose the command timeout from the xnvme_opts

### DIFF
--- a/include/libxnvme.h
+++ b/include/libxnvme.h
@@ -68,15 +68,17 @@ struct xnvme_opts {
 	struct {
 		uint32_t value : 31;
 		uint32_t given : 1;
-	} css;                 ///< SPDK controller-setup: do command-set-selection
-	uint32_t use_cmb_sqs;  ///< SPDK controller-setup: use controller-memory-buffer for sq
-	uint32_t shm_id;       ///< SPDK multi-processing: shared-memory-id
-	uint32_t main_core;    ///< SPDK multi-processing: main-core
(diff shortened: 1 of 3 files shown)


karlowich:
@LiadOz Thanks for this, I think it is a good idea.
Two things:
- The tests are failing, however I don't think that has anything todo with your changes. I will need to investigate
- I think it makes sense to have the previous values as default for the timeouts, do you agree?

LiadOz:
> @LiadOz Thanks for this, I think it is a good idea. Two things:
> 
>     * The tests are failing, however I don't think that has anything todo with your changes. I will need to investigate
> 
>     * I think it makes sense to have the previous values as default for the timeouts, do you agree?

I think it may be better to not enforce by default a certain timeout, but I don't have a strong opinion about it. Do you want me to change it?

karlowich:
I think the timeouts matching the kernel are a sensible default.
Besides, if I'm not mistaken, there are more use cases where a timeout is desirable than not.

karlowich (approved):
Thanks, this is great!
The CI failing has nothing to do with the changes you made, sorry for that.

LiadOz:
@karlowich the pipeline has passed, will this be merged into next soon?

safl (approved):
Thanks @LiadOz and @karlowich .
LGTM.

PR #290, opened 2023-05-26
fix(cijoe/tests): Speed up fabrics tests

karlowich:
By applying a CPU mask to the SPDK NVMe-oF target we can massively speed up the fabrics tests.

Additionally, we no longer need to adjust the parameters of the transport, and they can remain in the default configuration.


karlowich:
This speeds up the total running time of the CI from ~50 min to ~30 min

safl (approved):
Great!

PR #291, opened 2023-05-29
vfio/mmap: memory map and unmap support on vfio.

a-malakar:
Added support for memory map and unmap in vfio.


lib/xnvme_be_vfio_mem.c
```diff
@@ -84,6 +84,53 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
 	errno = ENOSYS;
 	return -ENOSYS;
 }
+
+void *
+xnvme_be_vfio_buf_map(const struct xnvme_dev *dev, size_t nbytes, uint64_t *XNVME_UNUSED(phys), void *vaddr)
+{
+	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
+	struct vfio_state *vfio = &state->ctrl->pci.vfio;
+  struct vfio_iommu_mapping *mapping;
```
ankit-sam, line 93: indentation, same comment at 2 more places in `xnvme_be_vfio_buf_unmap`

include/libxnvme.h
```diff
@@ -199,6 +199,37 @@ xnvme_buf_realloc(const struct xnvme_dev *dev, void *buf, size_t nbytes);
...
+/**
+ * Map a buffer for IO with the given device
+ *
+ * The buffer will be aligned to device geometry and DMA allocated if required by the backend for
+ * command payloads
+ *
+ * @note
+ * nbytes must be greater than zero and a multiple of minimal granularity
+ * @note
+ * De-allocate the buffer using xnvme_buf_free()
```
ankit-sam, line 211: I think this note about deallocating buffer is not required

ankit-sam:
@Baekalfen for context libblkio guys are trying to use xNVMe with libvfn backend. They already have a merge request  [libblkio](https://gitlab.com/libblkio/libblkio/-/merge_requests/168). As xNVMe doesn't have support for map / unmap which is one of the requirements for them. Kindly have a look into this.

lib/xnvme_be_vfio_mem.c
```diff
@@ -84,6 +84,53 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+	if (vfio_map_vaddr(vfio, vaddr, nbytes, NULL)) {
+		XNVME_DEBUG("FAILED: vfio_map_vaddr(): %s\n", strerror(errno));
+		return NULL;
+	}
+
+	return vaddr;
+}
+
+void
+xnvme_be_vfio_buf_unmap(const struct xnvme_dev *dev, void *buf)
```
birkelund, line 115: You think you want to be as generic as possible in the xNVMe API. Thus, you cannot rely on all implementations of mapping/unmapping buffer to be aware of the mapping length internally and should add a `length` parameter on this functions as well as allowing it to return an error code (mapping does not exist, unmapping failed, etc.).

lib/xnvme_be_vfio_mem.c
```diff
@@ -84,6 +84,53 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+	
+	XNVME_DEBUG("xnvme_be_vfio_buf_unmap(%p, %p)", dev, buf);
+	
+  mapping = vfio_iommu_find_mapping(&vfio->iommu, buf);
+	if (!mapping) {
+			XNVME_DEBUG("no mapping for %p\n", buf);
+		return;
+	}
+
+	if (vfio_unmap_vaddr(vfio, buf)) {
```
birkelund, line 129: In libvfn v2.0.0 this function will take an additional `size_t *len` parameter. If not `NULL` it will be set to the length of the mapping being unmapped. This helps the application deallocate the memory if it did so with, say, `mmap()`.

birkelund (changes requested):

lib/xnvme_be_vfio_mem.c
birkelund: Thinking about this some more, I am not 100% convinced that my suggestions is sound.... It *is* as hazzle for the user to keep track of buffer mapping lengths, and you might want to just decide that if the backend does not implement it, then the xnvme "glue" has to take care of it. Not sure.

docs/capis/xnvme.rst
```diff
@@ -205,6 +205,22 @@ xnvme_buf_free
 
 .. doxygenfunction:: xnvme_buf_free
 
+   
+.. _sec-c-api-xnvme-func-xnvme_buf_map:
+
+xnvme_buf_map
+---------------
```
safl, line 212: For the public API, to completely separate the notion of command buffer, usually a single allocated region of memory, against that of "mapped memory", which might be larger ranges of memory, used by multiple buffers. Then, let's provide a API seperate from buffers. Let's introduce: ``xnvme_mem`` with the methods map/unmap:

- [ ] Add memory-mapping API function ``xnvme_mem_map()``
- [ ] Add memory-mapping API function ``xnvme_mem_unmap()``

a-malakar:
@birkelund @safl, I've addressed all the changes requested and review comments. Kindly check.

safl:
Thanks @a-malakar ! Once these are merged:

* https://github.com/OpenMPDK/xNVMe/pull/287
* https://github.com/OpenMPDK/xNVMe/pull/283
* https://github.com/OpenMPDK/xNVMe/pull/292

Then I'll take another look. The first two **should** take care of the CI issues in this PR. The third provides an update to the libvfn backend using libvfn 2.0 and thereby changes to the API the backend is utilizing.

safl:
@a-malakar I have rebased on the ``next`` which fixes the CI issues you saw and additionally upgraded libvfn to v2.0.2. Thus, the changes @birkelund suggestion for libvfn v2 can now be made. I took the liberty of:

* Squashing in a correction to the ``xnvme_be_nosys_mem_map()`` correcting the returned value
* Adding a description of the change provided in the commit-message

lib/xnvme_be_vfio_mem.c
```diff
@@ -71,6 +71,41 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+int
+xnvme_be_vfio_mem_map(const struct xnvme_dev *dev, void *vaddr, size_t nbytes, uint64_t *phys)
+{
+	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
+	struct vfio_container *vfio = state->ctrl->pci.dev.vfio;
+	int err;
+
+	XNVME_DEBUG("xnvme_be_vfio_mem_map(%p, %p)", dev, vaddr);
+
+	nbytes = ALIGN_UP(nbytes, __VFN_PAGESIZE);
```
birkelund, line 84: This should be changed to a check on the alignment instead. That is, we should not map more than what the user requested and if the user did not provide a properly aligned number of bytes, then we bail out. Actually, `vfio_map_dma()` will check this (that `vaddr` **and** `nbytes` are aligned) and bail out if that is not the case.

birkelund (changes requested):

a-malakar:
@birkelund @safl, I've added a check for alignment. Please review.

include/xnvme_be.h
```diff
@@ -119,6 +119,16 @@ struct xnvme_be_mem {
 	 * Free a buffer usable for NVMe commands
 	 */
 	void (*buf_free)(const struct xnvme_dev *, void *);
+
+	/**
+	 * Map a buffer usable for NVMe commands
+	 */
+	int (*mem_map)(const struct xnvme_dev *, void *, size_t, uint64_t *);
```
birkelund, line 126: Name the parameters. Then they document themselves.

lib/xnvme_be_vfio_mem.c
```diff
@@ -71,6 +71,44 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+int
+xnvme_be_vfio_mem_map(const struct xnvme_dev *dev, void *vaddr, size_t nbytes, uint64_t *phys)
+{
+	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
+	struct vfio_container *vfio = state->ctrl->pci.dev.vfio;
+	int err;
+
+	XNVME_DEBUG("xnvme_be_vfio_mem_map(%p, %p)", dev, vaddr);
+
+	if (!ALIGNED(nbytes, __VFN_PAGESIZE)) {
```
birkelund, line 84: You should also check `vaddr` actually, i.e.

    if (!ALIGNED(vaddr | nbytes, __VFN_PAGESIZE)) { ...

lib/xnvme_be_vfio_mem.c
```diff
@@ -71,6 +71,44 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+		XNVME_DEBUG("FAILED: nbytes not page aligned\n");
+		return -EINVAL;
+	}
+
+	XNVME_DEBUG("vfio_map_vaddr(%p, %p, %ld, NULL)", vfio, vaddr, nbytes);
+	err = vfio_map_vaddr(vfio, vaddr, nbytes, phys);
+	if (err) {
+		XNVME_DEBUG("FAILED: vfio_map_vaddr(): %s\n", strerror(errno));
+		return err;
```
birkelund, line 93: I believe the xnvme convention is to return negative errno. libvfn returns negative (typically `-1`) on error and sets errno. So the above should be `return -errno;`.

birkelund (changes requested):

lib/xnvme_be_vfio_mem.c
```diff
@@ -71,6 +71,44 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME
...
+{
+	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
+	struct vfio_container *vfio = state->ctrl->pci.dev.vfio;
+	int err;
+
+	XNVME_DEBUG("xnvme_be_vfio_mem_map(%p, %p)", dev, vaddr);
+
+	if (!ALIGNED(nbytes, __VFN_PAGESIZE)) {
+		XNVME_DEBUG("FAILED: nbytes not page aligned\n");
+		return -EINVAL;
```
birkelund, line 86: Also set errno to `EINVAL`? I think?

safl (changes requested):
@a-malakar this is looking great, a couple of things.

- [x] @birkelund has some feedback on buffer-alignment and return-value convention. That would be great to address.
- [x] Add ``libxnvme_mem.h`` to ``include/meson.build``.
- [x] There are some CI-failures, these are fixed on: https://github.com/OpenMPDK/xNVMe/pull/293 thus need to get that in at this rebased.

safl:
Thanks @a-malakar @ankit-sam @birkelund !

PR #292, opened 2023-06-05
be/vfio: bump libvfn to v2.0.0

birkelund:
Update the libvfn-based backend to libvfn v2.0.0.

### DIFF
--- a/lib/xnvme_be_vfio_async.c
+++ b/lib/xnvme_be_vfio_async.c
@@ -94,7 +94,7 @@ xnvme_be_vfio_queue_poke(struct xnvme_queue *queue, uint32_t max)
 
 int
 xnvme_be_vfio_async_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes, void *mbuf,
-			   size_t XNVME_UNUSED(mbuf_nbytes))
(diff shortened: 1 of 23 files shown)


safl:
This was rebased on ``next`` and force-pushed in order to get the recent related changes from https://github.com/OpenMPDK/xNVMe/pull/283 and https://github.com/OpenMPDK/xNVMe/pull/287 changing the consumption of libvfn as "external" / on-system library instead of subproject.

Additionally, then I took the liberty of squashing in a change to the toolbox/pkgs bumping up libvfn to v2.0.0. This is needed now since it is no longer managed via the meson-subproject. The docker-imaages have been rebuilt with the scripts integraed.

There seems to be some issue now on ``xnvme_enumerate()``, looks like the vfio/libvfn backend is erroring out when it should not need to?

safl:
@birkelund altough the git-tag v2.0.1 is there, then the meson-file is not bumped: https://github.com/OpenMPDK/libvfn/blob/v2.0.1/meson.build

birkelund:
> @birkelund altough the git-tag v2.0.1 is there, then the meson-file is not bumped: https://github.com/OpenMPDK/libvfn/blob/v2.0.1/meson.build

Yeah, I'm an idiot. Fixed now in v2.0.2...

PR #298, opened 2023-06-26
Python package deployment

safl:
This provides a collection of fixes, the most pressing is the issue introduced by:

* https://github.com/ofek/userpath/issues/50

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -70,11 +70,15 @@ jobs:
       run: |
         make clean gen-src-archive
 
+    - name: Rename to un-versioned xnvme-src.tar.gz
+      run: |
+        mv builddir/meson-dist/xnvme-0.7.0.tar.gz builddir/meson-dist/xnvme-src.tar.gz
+
     - name: Upload source archive
       uses: actions/upload-artifact@v3.1.1
       with:
-        name: archive-src
-        path: builddir/meson-dist/*.tar.gz
+        name: xnvme-src-archive
+        path: builddir/meson-dist/xnvme-src.tar.gz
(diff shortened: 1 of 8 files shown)


cijoe/scripts/gha_prepare_python.py
```diff
@@ -42,8 +42,20 @@ def main(args, cijoe, step):
         log.error(f"invalid step({step})")
         return errno.EINVAL
 
+    # TODO: because pipx / userpath only adds PATH to files read by
+    # login-shells, then we cannot use 'pipx ensurepath' anymore.
+    # Issue has been reported here:
+    #
+    # * https://github.com/pypa/pipx/pull/1021
+    # * https://github.com/ofek/userpath/issues/50
+    #
```
karlowich, line 51: I think this should be an issue on github instead of a todo in the code

.github/workflows/verify.yml
```diff
@@ -101,15 +105,14 @@ jobs:
     container: ghcr.io/xnvme/xnvme-deps-citools-latest:next
 
     steps:
-    - name: Retrieve the source-archive
+    - name: Retrieve, the xNVMe source archive
```
karlowich, line 110: The added comma here doesn't make a lot of sense. Same goes for the rest of the cases

cijoe/scripts/gha_prepare_python.py
safl: Yeah, changed to a NOTE instead. As the code needed a comment...

ankit-sam (approved):
Looks good

karlowich (approved):

PR #301, opened 2023-07-11
libconfig: remove deprecated subprojects from library info

ankit-sam:
fio and libvfn are no longer subprojects after xnvme v0.7.0 release. Remove them from library-info


safl (approved):
Great, thanks.

karlowich (approved):
Looks good!

PR #303, opened 2023-07-20
feat(be/aio): add aio ring to improve performance

karlowich:
By copying io_events from memory instead of using io_getevents() during poke, we get a significant performance improvement.

This is part of an effort to improve the performance of the xNVMe integrated into SPDK, and with this addition the performance of bdev_xnvme is comparable to bdev_aio.

The graphs below show  IOPS with 4k block-size and varying queue-depth:

| Devices | Without AIO ring | With AIO ring |
| --- | --- | --- |
|  1  | bdevperf_lineplot_libaio | bdevperf_lineplot_libaio |
| 7  |  bdevperf_lineplot_libaio | bdevperf_lineplot_libaio |

### DIFF
--- a/lib/xnvme_be_linux_async_libaio.c
+++ b/lib/xnvme_be_linux_async_libaio.c
@@ -11,10 +11,27 @@
 #ifdef XNVME_BE_LINUX_LIBAIO_ENABLED
 #include <errno.h>
 #include <libaio.h>
+#include <stdatomic.h>
 #include <xnvme_queue.h>
 #include <xnvme_be_linux.h>
 #include <xnvme_be_linux_libaio.h>
 #include <xnvme_dev.h>
+
+struct xnvme_aio_ring {
+	uint32_t id;
+	uint32_t nr;
+	uint32_t head;
+	uint32_t tail;
+
+	uint32_t magic;
+	uint32_t compat_features;
+	uint32_t incompat_features;
+	uint32_t header_length;
+
+	struct io_event events[0];
+};
+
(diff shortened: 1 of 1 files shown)


lib/xnvme_be_linux_async_libaio.c
```diff
@@ -57,17 +74,41 @@ int
...
+		min = queue->poll_io ? 0 : 1;
+		completed = io_getevents(queue->aio_ctx, min, max, queue->aio_events, &timeout);
+		if (completed < 0) {
+			XNVME_DEBUG("FAILED: completed: %d, errno: %d", completed, errno);
+			return completed;
+		}
+
+		/* Otherwise copy events from memory */
+	} else {
+		uint32_t current = ring->head;
```
birkelund, line 97: Do you need to use an atomic load with acquire semantics here?

safl: > Do you need to use an atomic load with acquire semantics here?

Hmm, I guess it is not required, since worst case then we read an "older" ``head`` and consequently process less I/O. However, it should not lead to erroneous behavior? Or, am I missing here?

safl: Like, would there be a re-ordering issue her?

birkelund: No, I think you are right, sorry. The store release should be enough.

safl (approved):
Excellent @karlowich !

PR #308, opened 2023-07-24
allow overriding library loading by env var

jwdevantier:
**Use-case**
My system might differ from yours, I may not have pkg-config, or I may have a copy of xNVMe installed elsewhere.

This change permits me to override the library-loading logic by exporting `XNVME_LIBRARY_PATH`, pointing to the library file I want to load.
This change even allows users of any downstream applications to work around issues of locating the xNVMe library on their systems, without the application making special provisions.


safl (approved):
This looks good to me, thanks!

PR #312, opened 2023-07-27
fix(gha): adjust conditions for build-and-test

safl:
With these changes, then it should possible to enable the github merge-queue. This means that:

* When a PR for the ``next`` branch meets all requirements
  * Is rebased / up-to-date
  * Has been reviewed
  * Passes "fast" checks (code-format and build-checks)

Then it can go into the "merge-queue", when entering the merge-queue, additional longer-running checks are performed, specifically: spinning up a virtual machine (qemu) with NVMe device emulation and running the full functional test-suite on it. In case the testing passes, then the PR is "automatically" merged into ``next``. In case it fails, then it is removed from the merge-queue and the submitter can choose to:

* Reproduce the issue locally
* Make changes
* Send to the merge-queue again

The merge-queue is introduced to make it more transparent when multiple PRs are ready to be merged but then they each have to sit and wait until it is their turn. In this flow, one has to constantly rebase as PRs trickle in. This also means re-running the long-running tests again and again. Thus, this integration-test-phase is **hopefully** now something that can be manage by the merge-queue, where PR as put on top of each other, and then tested.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -118,154 +118,6 @@ jobs:
       run: make gen-man-pages
 
   #
-  # Code-analysis using scan-build
-  #
-  analyze-scan-build:
-    needs: source-archive
-    runs-on: ubuntu-latest
-    container: ghcr.io/xnvme/xnvme-deps-citools-latest:next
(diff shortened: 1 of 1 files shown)


ankit-sam (approved):
Looks good

PR #315, opened 2023-08-01
vfio/map unmap tests

a-malakar:
Added test case to verify memory map/unmap in vfio. 
Also changed memory unmap return type.


tests/map.c
```diff
@@ -0,0 +1,93 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <errno.h>
+#include <sys/mman.h>
+#include <libxnvme.h>
+#include <xnvme_be_vfio.h>
```
ankit-sam, line 8: builds are failing for macos and apline linux because of its inclusion. I don't see any need for this header file inclusion.

tests/map.c
```diff
@@ -0,0 +1,97 @@
...
+			nerr += 1;
+			continue;
+		}
+	}
+
+	if (nerr) {
+		xnvme_cli_pinf("--={[ Got Errors - see details above ]}=--");
+		xnvme_cli_pinf("nerr: %d out of count: %zu", nerr, count);
+	} else {
+		xnvme_cli_pinf("LGMT: xnvme_mem_{map,unmap}");
```
karlowich, line 58: >"LGMT"

Should be `LGTM`

karlowich (changes requested):
This is looking good, I just have two comments

lib/xnvme_be_vfio_mem.c
```diff
@@ -107,7 +107,10 @@ xnvme_be_vfio_mem_unmap(const struct xnvme_dev *dev, void *buf)
 
 	if (vfio_unmap_vaddr(vfio, buf, NULL)) {
 		XNVME_DEBUG("FAILED: vfio_unmap_vaddr(-, %p): %s\n", buf, strerror(errno));
+		return errno;
```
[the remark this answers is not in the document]
safl: I missed this earlie. So, it would be good to also change the ``xnvme_be_vfio_mem_map()`` function as well.

safl (changes requested):
This is looking great, can you also add a Python test (invoking the binary ``xnvme_tests_map``. In the folder: ``cijoe/tests/logic/xnvme_tests_map.py`` something like:

```python
import pytest

from ..conftest import xnvme_parametrize


@xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
def test_verify(cijoe, device, be_opts, cli_args):
    err, _ = cijoe.run(f"xnvme_tests_map verify {cli_args}")
    assert not err
```

This will make sure that the added tests, written in C, is executed as part of the CI.

safl (approved):

PR #316, opened 2023-08-02
Rust bindings

s-rajpoot:
Added rust bindings for dev and queue

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,3 +75,18 @@ repos:
     args:
     - --extend-ignore=E203,F401,F811,E501
     exclude: (ctypes_mapping\/api\.py)$
+
+# Rust tools
+- repo: https://github.com/doublify/pre-commit-rust
+  rev: v1.0
+  hooks:
+  - id: fmt
+    name: Rust-fmt-xnvme
+    args: ['--verbose', '--manifest-path', 'rust/xnvme/Cargo.toml', '--']
+
+- repo: https://github.com/doublify/pre-commit-rust
+  rev: v1.0
+  hooks:
+  - id: fmt
+    name: Rust-fmt-xnvme-sys
+    args: ['--verbose', '--manifest-path', 'rust/xnvme-sys/Cargo.toml', '--']
--- a/rust/xnvme-sys/Cargo.toml
+++ b/rust/xnvme-sys/Cargo.toml
@@ -1,10 +1,17 @@
 [package]
 name = "xnvme-sys"
-version = "0.0.0"
+version = "0.7.1"
 edition = "2021"
-description = "Raw/direct/unsafe bindings to the xNVMe C Library. NOTE: This is a crate-namespace-reservation"
+description = "Raw/direct/unsafe bindings to the xNVMe C Library."
 license = "BSD-3-Clause"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+
+[build-dependencies]
+system-deps = "6"
+bindgen = "0.63.0"
+
+[package.metadata.system-deps]
+xnvme = "0.7.1"
--- a/rust/xnvme-sys/build.rs
+++ b/rust/xnvme-sys/build.rs
@@ -1,0 +1,29 @@
+fn main() {
+    // Emit link options for linking system dependencies
+    let deps = system_deps::Config::new().probe().unwrap();
+
+    // Generate include arguments for clang
+    let flags: Vec<String> = deps
+        .all_include_paths()
+        .iter()
+        .map(|x| format!("-I{}", x.to_str().unwrap()))
+        .collect();
+
+    // Generate bindings for xnvme
+    let bindings = bindgen::Builder::default()
+        .header("../../include/libxnvme.h")
+        .allowlist_function("xnvme_.*")
+        .allowlist_type("xnvme_.*")
+        .clang_args(flags)
+        // Tell cargo to invalidate the built crate whenever any of the
+        // included header files changed.
+        .parse_callbacks(Box::new(bindgen::CargoCallbacks))
+        .generate()
+        .expect("Failed to generate bindings");
+
+    // Write bindings to file
+    let out_path = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap());
+    bindings
+        .write_to_file(out_path.join("bindings.rs"))
+        .expect("Failed to write bindings to file");
+}
--- a/rust/xnvme/Cargo.toml
+++ b/rust/xnvme/Cargo.toml
@@ -5,12 +5,3 @@ edition = "2021"
 description = "Safe and idiomatic xNVMe-bindings built on top of xnvme-sys. NOTE: This is a crate-namespace-reservation"
 license = "BSD-3-Clause"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[dependencies]
-xnvme-sys = { path = "../xnvme-sys" }
-anyhow = "1"
-thiserror = "1"
-derive_builder = "0.12"
-const-str = "0.5.3"
-clap = { version = "4", features = [ "derive" ] }
-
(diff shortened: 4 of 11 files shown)


rust/xnvme/src/device.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub fn open(device_uri: &str, options: &crate::options::Options) ->
+                                                                Result<Self> {
+        let device_uri_c = std::ffi::CString::new(device_uri)?;
+
+        let dev = unsafe { xnvme_sys::xnvme_dev_open(
+                                        device_uri_c.as_ptr(),
+                                        &mut options.as_raw())
+                            };
+
+        // Todo: check non null
```
metaspace, line 39: Resolve this TODO

rust/xnvme/src/device.rs
```diff
@@ -0,0 +1,374 @@
...
+}
+
+/*
+ * Returns the geometry of the given device
+ *
+ * @param dev Device handle obtained with xnvme_dev_open()
+ *
+ * @return The geometry (struct xnvme_geo) of given device handle
+ */
+pub fn dev_get_geo(dev: *const xnvme_dev) -> *const xnvme_geo {
```
metaspace, line 57: The rest of the functions in this file seem to be rather pointless. What is the purpose of these? The documentation is a copy of the C documentation which is already rendered as part of `xnvme_sys`.

metaspace (changes requested):
I think this needs some more work. I think we should add some CI hooks as mentioned in my review comments.

safl (changes requested):
In addition to the changes noted by @metaspace  the let's reduce the scope of the Rust support to the ``xnvme-sys`` crate. Thus, drop the "idiomatic" bindings, that is, the crate named ``xnvme``.
The latter crate requires a lot more iterations and design, whereas the first crate should be ready to go (with the minor changes).

safl (approved):

safl:
@metaspace the PRs has been sliced down to only provide the bindgen-emitted library-bindings. Could I bother you for a quick look?

safl (approved):
This is fine as is, adding simple ``raw`` /  ``xnvme-sys`` bindings.
The interface is **very** "rough around the edges", however, this is something we wish to improve with yace (https://github.com/safl/yace).

.pre-commit-config.yaml
```diff
@@ -75,3 +75,18 @@ repos:
...
     - --extend-ignore=E203,F401,F811,E501
     exclude: (ctypes_mapping\/api\.py)$
+
+# Rust tools
+- repo: https://github.com/doublify/pre-commit-rust
+  rev: v1.0
+  hooks:
+  - id: fmt
+    name: Rust-fmt-xnvme
+    args: ['--verbose', '--manifest-path', 'rust/xnvme/Cargo.toml', '--']
```
metaspace, line 85: Will this format the code in place, or will it error out if formatting is wrong? I guess the latter is what we want?

rust/xnvme-sys/Cargo.toml
```diff
@@ -1,10 +1,18 @@
 [package]
 name = "xnvme-sys"
-version = "0.0.0"
+version = "0.7.1"
```
metaspace, line 3: It looks like we are locking version of Rust bindings to xnvme version. Does xnvme use proper semver? If we find a bug in the Rust bindings and we need to bump patch version for the Rust crate, will we do a release of C xnvme with same patch level bump?

rust/xnvme/Cargo.toml
```diff
@@ -1,10 +1,16 @@
...
-
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+xnvme-sys = { path = "../xnvme-sys" }
+anyhow = "1"
+thiserror = "1"
+derive_builder = "0.12"
+const-str = "0.5.3"
+clap = { version = "4", features = [ "derive" ] }
```
metaspace, line 15: We have no code in this crate, so we probably do not need these dependencies.

metaspace (changes requested):
Looks good, nice with the CI stuff. I have a few comments inline.

rust/xnvme-sys/Cargo.toml
safl: yup, xNVMe subscribes to the dogma of SemVer. That is why xNVMe is currently at < 1.0, to have the freedom granted by paragraph 4 and 5 of SemVer.

Yup, I find that locking the ``xnvme-sys`` crate to the xnvme version is sensible, since it is such as "raw" binding to the xNVMe library, and yes, the intent is to bump xNVMe version in case issues are found in the Rust bindings. This also holds for the Python bindings.

For the potential ``xnvme`` create, e.g. the idiomatic-bindings, then this might have its own versioning-number, since it will introduce Rust-specific and self-contained constructs. E.g. once xNVMe itself hits v1, v2 etc. then the idiomatic bindings might not be ready for that...

rust/xnvme/Cargo.toml
safl: Right! Let me fix that, thanks!

.pre-commit-config.yaml
safl: Did a quick check. It does both. That is, when running pre-commit it will fix the format, and it will output that it needed to. E.g. like this:

```
make format-all
## xNVMe: format-all
C-format-headers.........................................................Passed
C-format-source..........................................................Passed
reuse....................................................................Passed
YAML-format-check........................................................Passed
YAML-format..............................................................Passed
Python-format-black......................................................Passed
Python-format-isort......................................................Passed
Python-lint-flake8.......................................................Passed
Rust-fmt-xnvme...........................................................Failed
- hook id: fmt
- files were modified by this hook

[bin (2021)] "/Users/safl/git/xnvme/rust/xnvme/src/main.rs"
rustfmt --edition 2021 rust/xnvme-sys/src/bin/xnvme_rust_example01.rs rust/xnvme-sys/build.rs rust/xnvme-sys/src/lib.rs rust/xnvme/src/main.rs /Users/safl/git/xnvme/rust/xnvme/src/main.rs

Rust-fmt-xnvme-sys.......................................................Passed
make: *** [format-all] Error 1
```

 Thus, it is useful for cleaning up the code and for the ci-task or checking whether cod eon a PR has formatting issues.

rust/xnvme/Cargo.toml
safl: done.

safl (dismissed):
LGTM. Thanks @s-rajpoot for your work on this!

rust/xnvme-sys/src/bin/xnvme_rust_example01.rs
```diff
@@ -0,0 +1,12 @@
...
+fn main() -> ExitCode {
+    println!(
+        "v{}.{}.{}",
+        unsafe { xnvme_sys::xnvme_ver_major() },
+        unsafe { xnvme_sys::xnvme_ver_minor() },
+        unsafe { xnvme_sys::xnvme_ver_patch() },
+    );
+
+    ExitCode::from(0)
+}
```
metaspace, line 12: This should probably be an example, see [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples)

metaspace (changes requested):
LGTM, but we probably should move the example to the example directory.

rust/xnvme-sys/build.rs
```diff
@@ -0,0 +1,29 @@
+fn main() {
+    // Emit link options for linking syste dependencies
```
karlowich, line 2:
```suggestion
    // Emit link options for linking system dependencies
```

rust/xnvme-sys/Cargo.toml
karlowich, line 3: Is there a way to store the version in a variable to make sure we don't have a case where it is updated here but not in the bottom?
```
[package.metadata.system-deps]
xnvme = "0.7.1"
```

karlowich (commented):
Two tiny comments

rust/xnvme-sys/Cargo.toml
metaspace: I don't think so. You could use a template engine to generate the manifest maybe? Probably not worth it.

karlowich: Nevermind then :)

karlowich (approved):

PR #318, opened 2023-08-07
feat(be/uring): Improve performance with batching

karlowich:
The performance of uring can be greatly improved by batching commands by
postponing submission until poke() is called.

This is part of an effort to improve the performance of the xNVMe
integrated into SPDK, and with this (and the addition of cqe_peek()
instead of cqe_wait()), the performance of bdev_xnvme is equal or better
compared to bdev_uring.

The behavior is enabled by setting the environment variable
XNVME_QUEUE_BATCHING.

The graphs below show the improvement, both are with 7 devices and bs = 4096.
(SOP = batching; CCPU = POLL_SQ instead of POLL_IO)

`modprobe nvme poll_q=0`
bdevperf_lineplot_io_uring

`modprobe nvme poll_q=1`
bdevperf_lineplot_io_uring


safl (approved):
Excellent, thank you @karlowich

PR #322, opened 2023-08-08
feat(be): expand backend-interface with pseudo-commands

safl:
This should enable the following:

- [ ] Show controller-registers
- [ ] Controller-reset
- [ ] Subsystem-reset
- [ ] Namespace-rescan
- [ ] Consider whether this code-infra can replace how the associated dev-namespace-id is handled

The list above will most likely expand.

TODO

* First implementation will be for the Linux backend (``xnvme_be_linux.c``)
* Add code-path going into the backend
  - example of how to encode "show-registers" in a "pseudo-command" and how the backend should process it
  - example of how to encode "controller-reset" in a "pseudo-command" and how the backend should process it
  - ``xnvme_topology.c`` will do the "pseudo-command", then every backend will have processing in e.g. ``xnvme_be_linux_admin_pseudo.c``


include/xnvme_be.h
```diff
@@ -77,6 +77,14 @@ struct xnvme_be_admin {
 	 */
 	int (*cmd_admin)(struct xnvme_cmd_ctx *, void *, size_t, void *, size_t);
 
+	/**
+	 * Pass pseudo admin-commands
+	 *
+	 * That is representation of various operations in NVMe commands, that look
+	 * like but are not actual NVMe commands.
+	 */
+	int (*cmd_pseudo)(struct xnvme_cmd_ctx *, void *, size_t, void *, size_t);
```
ankit-sam, line 86: This requires a flag to differentiate what the pseudo command needs to do (subsystem reset, or controller reset etc).

safl: The idea is that we would encode this in the NVMe-command. That is, in ``ctx->cmd``. E.g. somewhere, library-internal, we can keep a list of opcodes for the different operations. That is, we assign opcodes to the operations: (subsysystem-reset, controller-reset, namespace-rescan, show-reqs.). Then, in e.g. ``lib/xnvme_topology.c`` the functions would be implemented by in a form similar to command-construction for NVMe-command-sets.

Something like:

```c
  struct xnvme_cmd_ctx ctx = { 0 };
  int err;
  
  ctx.cmd.common.opcode = XNVME_SPEC_PSEUDO_OPC_CONTROLLER_RESET;

  err = xnvme_cmd_pass_pseudo(&ctx, NULL, 0, NULL, 0);

  return err | ctx.cpl.status.val;
```

Let me add those operations as pseudo-opcodes, and fill them out to better demonstrate this.

safl: Expanded with opcodes, xnvme_cmd_pass_pseudo(), and usage-examples.

ankit-sam: Yeah this should work

ankit-sam (approved):
Looks good

PR #325, opened 2023-08-11
fix(build/fedora): add zlib-devel dependency

karlowich:
The `test_linkandload` fails on fedora since zlib-devel is missing. 
With this addition, that is no longer the case.


safl (approved):
LGTM, great!

PR #328, opened 2023-08-15
feat(build/spdk): Enable support for external SPDK

karlowich:
Now Meson is able to find a system-install SPDK and compile with that.
Thus, the SPDK subproject has been removed since it is no longer needed.

This relies on Meson actually being able to find the system-installed
SPDK. If 'XNVME_BE_SPDK_ENABLED' is true and it cannot find
SPDK, the build will fail.

This might require setting `export PKG_CONFIG_PATH` before building and
`export LD_LIBRARY_PATH` before running, depending on where SPDK is
installed.

Additionally, this PR changes installation of  both SPDK and DPDK to be from src.
This is to make sure that xNVMe is able to find both SPDK and DPDK libraries on
runtime. If DPDK is built through SPDK, the libraries are never
installed and thus cannot be located through LD.


safl (commented):
Thanks for the effort on this! A lot is going on here and I have to questions / concerns with the approach of relying on SPDK being installed on the system as a proper link-target, as well as questions to the changes in general.

1) Regarding the introduction of using ``--prefix=/usr`` by default, e.g.

	meson setup builddir --prefix=/usr

Is a bit nerving to me. One of the great feats of meson is that it **should** take care of the correct paths, adhering to OS / Distribution conventions.
Slapping on ``--prefix=/usr`` seems to me like dropping some of the Meson niceties or what am I missing here? Why is this needed?

2) Where can these be located when relying on a system-install of SPDK:

* nvmf_tgt
* dpdk/build/kmod
* spdk/scripts/rpc.py

My assumption is that SPDK does not install these on the system, right? This is not an issue for xNVMe and consumption of the NVMe-driver, however, to my knowledge, then common practice for SPDK applications is to fork the SPDK repository and add the application within, thus, the SPDK repository is usually always "around" together with the application? Similar to our usage, where we are using the fabrics-target application, as well as the rpc.py scripts for bdev setup etc.

I am mentioning this as, it might not be the most common practice to link with SPDK "installed on the system", however, rather that one wants to link with SPDK in the location where the SPDK repository is located. Since, one needs a bunch of tooling, at runtime, which is only available in the repository.

3) From reading the changes, then it looks like xNVMe will with this, always link against SPDK as a shared library? That is, instead of linking against static-libraries, and thereby potentially reaping some LTO-benefits, then with this, xNVMe will rely on the at-runtime availability and loadability of the SPDK shared library.

On this, then it was a great improvement when we dropped the "bundling" of the SPDK into xNVMe and I think the replacement with static-libraries via pkg-config is great as it solves the challenge for the user. However, dropping static libraries, in favor of SPDK as a shared library, here I am wavering a bit...

4) The need to explicitly manage the build and installation of DPDK on the system is adding to my uncertainty on this...

Reservations
============

Dropping the SPDK subproject is great! This should be done, and there are a ton of excellent changes going on in this PR. However, I am uncertain whether consuming SPDK via "installed-on-the-system" is the right replacement. Due to the stuff above, then I am currently leaning towards not linking with SPDK "installed-on-the-system" but instead "link-with-spdk-in-a-given-folder".

That is, changing the meson-option "with-spdk" to be a path, and then link with it, in the same manner as we do now (with the static libraries), and generate the pkg-config in the same manner (adding the SPDK static libraries). By doing so, I see these benefits:

* Potential LTO on static libraries
  - For xNVMe itself and consumers of xNVMe

* Today, xNVMe provides the means for the user to consume the user-space SPDK NVMe-driver, without having to install DPDK + SPDK on their system
  - For some, then this is a nice feature, that is reducing what is required to install on the system, yet still enabling them to use it
  - For others, then it is nice because there might be a conflicting version already partly installed on the system for a different DPDK application
  - In other words, it is less intrusive to link with DPDK/SPDK from a non-system folder

* Users of xNVMe can provide their custom version of SPDK

  - These are usually proprietary or customization not publically available
  - With a ``-Dwith-spdk=PATH`` then it is easy to point to it
  - Multiple SPDK versions / customizations can exist on a given system

* We can provide a custom version of SPDK with not-yet-upstreamed patches

  - We currently need this for the patches causing failure (idfy-passthru / get-zone-change etc.)
  - In the CI we can have multiple SPDK versions
    * Tracking latest stable upstream
    * Custom local with non-yet-upstream patches
    
* Linking "with-spdk-in-a-given-folder" **might** also align better with https://github.com/OpenMPDK/xNVMe/pull/275
  - E.g. provide a different SPDK version when building on Windows
  - More convenient to locate the SPDK build on Windows when explicitly assigning it

karlowich:
@safl Let me start by addressing your questions:

> Regarding the introduction of using --prefix=/usr by default, e.g.

I use this to change the prefix from `/usr/local` to `/usr`, it doesn't change the structure of the directories within the folder, libraries still go to `lib` (or `lib64`) and headers to `include` and so on.
This is not required by a user. On most `/usr/local` just requires that the user adds the environment variables `export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig` (for compile time) and `export LD_LIBRARY_PATH=/usr/local/lib/` (for run time) to a bashrc or similar.
We don't have the luxury of a config in our CI and thus it easier to adjust the install path than the `export` statements everywhere.

> Where can these be located when relying on a system-install of SPDK

The executables reside inside their project directories, the path is exactly the same as before except that it is not wrapped in the `xnvme/subprojects/` directory. I don't see this being an issue, it is not exactly intuitive to find `kmod` in `xnvme/subprojects/spdk/dpdk/build/kmod` which is the current location.

> From reading the changes, then it looks like xNVMe will with this, always link against SPDK as a shared library?

This is correct, there is currently no way to elegantly link the SPDK static library with the required `-Wl,--whole-archive` through meson.
I don't know much about LTO, but it would be interesting to compare performance between this branch and `next`.
Secondly, is LTO even relevant when using `-Wl,--whole-archive`?

> The need to explicitly manage the build and installation of DPDK on the system is adding to my uncertainty on this...

This is a non-issue, the version of DPDK is locked and I don't see why this would cause a problem.

a16bitsysop:
there is an spdk issue about the scripts not being installed with `make install` https://github.com/spdk/spdk/issues/3143

karlowich:
Hi @a16bitsysop, thanks for the heads up - that looks very interesting!
For now, this PR is on hold, since we have some patches in the internal SPDK, which are yet to be merged in upstream SPDK.
Do you have a need/use-case for the support for external SPDK?

a16bitsysop:
I have packaged xnvme and libvfn for alpine linux, which has spdk and dpdk packaged as well.  So for me to use spdk/dpdk it is better to use the packaged ones, which is the same for most distros.

karlowich:
First off, thank you for your packaging effort, that is much appreciated!
Knowing that someone has a use for this change, I will give it another shot in the future (Most likely when the next SPDK release is made)

safl:
> I have packaged xnvme and libvfn for alpine linux, which has spdk and dpdk packaged as well. So for me to use spdk/dpdk it is better to use the packaged ones, which is the same for most distros.

@a16bitsysop thanks for the packaging effort, it is highly appreciated!

Regarding the consumption of SPDK via on-system installation / packaged SPDK. It would be preferable for xNVMe to consume SPDK like it does with libaio, liburing, libvfn, etc. However, SPDK, unlike DPDK, is just not readily available in packaged form on the Linux distros that xNVMe targets, and linking it against the on-system installation of SPDK has its drawbacks/limitations, as previously mentioned. Thus, the current state of things and this PR.

Having said this, the mentioned limitations/drawbacks could be changed/mitigated by the effort you referred to (https://github.com/spdk/spdk/issues/3143), and perhaps we could contribute to that.

Currently, I would propose that if you @karlowich or somebody else take up this PR again, that it would be taken in the direction of:

* When meson detects SPDK on the system ==> then link with and use that.
  * With this xNVMe can conveniently consume on-system-installed / packaged versions of SPDK
* When meson cannot detect SPDK ==> then use the subproject (what we do now)
  * With this xNVMe can conveniently consume the SPDK NVMe/driver with (arguably) minimal hassle
* When meson cannot detect SPDK ==> meson-option to use SPDK from user-given path instead of subproject
  * With this, then users can conveniently provide their own customized SPDK

The different approaches each trade-off certain things and thus, I would not like to drop one for the other, but rather use all in a fallback-manner with user-override via meson-option.

karlowich:
This PR is so outdated, that there is no use in keeping around.
Instead, I will create an Issue for tracking this.

PR #329, opened 2023-08-16
be:macos: add missing include of mach_error.h

Baekalfen:
Compilation fails without this include

### DIFF
--- a/lib/xnvme_be_macos_admin.c
+++ b/lib/xnvme_be_macos_admin.c
@@ -6,6 +6,7 @@
 #include <xnvme_be.h>
 #include <xnvme_be_nosys.h>
 #ifdef XNVME_BE_MACOS_ENABLED
+#include <mach/mach_error.h>
 #include <CoreFoundation/CoreFoundation.h>
 #include <IOKit/storage/nvme/NVMeSMARTLibExternal.h>
 
--- a/lib/xnvme_be_macos_dev.c
+++ b/lib/xnvme_be_macos_dev.c
@@ -6,6 +6,7 @@
 #include <xnvme_be.h>
 #include <xnvme_be_nosys.h>
 #ifdef XNVME_BE_MACOS_ENABLED
+#include <mach/mach_error.h>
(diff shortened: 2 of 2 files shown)


safl:
Since it passes on:

* My physical mac environment (Ventura @ M1)
* GitHUB CI (MacOS 11/12/13 on Intel)

Then I am curious, in which environment does this fail the compilation?

Baekalfen:
On my Ventura M1, if I checkout `next` (df37dd669180332d774e03a988f74a5253c23ced), and do `make clobber`, `make config-debug`, `make`, it fails like this:

```
../lib/xnvme_be_macos_admin.c:49:51: error: call to undeclared function 'mach_error_string'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                        XNVME_DEBUG("Error in GetLogPage: %s", (char *)mach_error_string(err));
```

But doesn't it also make sense that it fails? Where would it get the definition of `mach_error_string` from otherwise?

It works on GHA because it's not building with debug on, so the call is eliminated.

safl:
> On my Ventura M1, if I checkout `next` ([df37dd6](https://github.com/OpenMPDK/xNVMe/commit/df37dd669180332d774e03a988f74a5253c23ced)), and do `make clobber`, `make config-debug`, `make`, it fails like this:
> 
> ```
> ../lib/xnvme_be_macos_admin.c:49:51: error: call to undeclared function 'mach_error_string'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
>                         XNVME_DEBUG("Error in GetLogPage: %s", (char *)mach_error_string(err));
> ```
> 
> But doesn't it also make sense that it fails? Where would it get the definition of `mach_error_string` from otherwise?
> 
> It works on GHA because it's not building with debug on, so the call is eliminated.

Ah, it is when debugging is turned on. Well, that makes perfect sense then.

safl (approved):
LGTM

karlowich (approved):
Great!

@safl I have seen #330 and I agree that is a good idea

PR #335, opened 2023-09-01
Toolbox cleanup and third-party integration

safl:
This does a bunch of fixes and cleanup of meson and the toolbox. A lot of technical debt is payed-off here, as well some preparation for potential changes.

- [x] Change optional-dependencies from booleans to features
  - This is a requirement to achieve the proper graceful degradation of the optional dependencies.
- [x] Graceful degradation of libaio / liburing / libvfn
  - The build fails when optional libraries are not available. This is convenient in the sense that one is informed ahead-of-time, that a given feature is not available. However, it is not customary, and it is **not** convenient for those who just want to grab xNVMe, build, and install it to use whatever is available on the system. Perhaps only NVMe-driver ioctls and async. emulation.
- [x] Test that xNVMe is linked as expected
  - The change to graceful degradation of library dependencies motivates this.
  - That is, libaio/liburing/libvfn and SPDK, on the systems where these are available in sufficient versions. Then xNVMe ought to be able to link with them are exposed that via ``xnvme library-info``. We have not been testing this, and thus, libvfn had been disabled on Alpine when it should been enabled. Also, for the potential removal of SPDK as a subproject, then a test is definately needed, instead of manually inspecting build-status.
- [x] Build source-package in ``toolbox/third-party``
  - Move the ``third-party`` folder into ``toolbox``, this provides the use of the ``toolbox`` in a "standalone" manner. That is, in the CI/docker-images, the toolbox will with this be deployed at ``/opt/toolbox`` and not the entire xNVMe repository
  - This changes the described way, and the way used by CI, on where src-build-install of optional dependencies are retrieved to and built.
  - This provides for infrastructure that could replace the use of subprojects


toolbox/pkgs/emitter/templates/pkgman-dnf-oraclelinux-9.sh.jinja
```diff
@@ -0,0 +1,9 @@
+# This repos has CUnit-devel + meson
```
ankit-sam, line 1: Small nit

These repos
or
This repo

toolbox/pkgs/emitter/templates/pkgman-dnf-rockylinux-9.2.sh.jinja
```diff
@@ -1,6 +1,6 @@
-# This repos has CUnit-devel
-dnf install -y 'dnf-command(config-manager)' || true
-dnf config-manager --set-enabled powertools || true
+# This repos has CUnit-devel + meson
```
ankit-sam, line 1: Same as above

ankit-sam (dismissed):
Few nitpicks but, else looks ok.

toolbox/pkgs/emitter/templates/pkgman-dnf-oraclelinux-9.sh.jinja
safl: Fixed.

toolbox/pkgs/emitter/templates/pkgman-dnf-rockylinux-9.2.sh.jinja
safl: fixed.

ankit-sam (approved):

PR #341, opened 2023-09-05
vfio: fix completion result for sync io

Baekalfen:
The VFIO/libvfn interface used a deprecated member 'cqe_sync' on the state struct. This is changed to the correct command context, so the completion result is returned.

Unused members 'queue_id' and 'cqe_sync' removed from VFIO state.

### DIFF
--- a/include/xnvme_be_vfio.h
+++ b/include/xnvme_be_vfio.h
@@ -15,13 +15,11 @@
 struct xnvme_be_vfio_state {
 	struct nvme_ctrl *ctrl;
 	unsigned long long qidmap; // Queue identifier bit map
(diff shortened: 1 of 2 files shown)


safl (commented):
LGTM. Can you make minor change to the commit-message, we are trying to adopt the "Conventional Commits" (https://www.conventionalcommits.org/en/v1.0.0/#summary).
So, specifically change to something like:

``fix(be/vfio): ... brief description ...``

Baekalfen:
@safl I updated the commit headline

safl (approved):
LGTM, thanks!

PR #342, opened 2023-09-08
fix: Adjust format-strings

### LINKED ISSUE #98: Adjust format-strings
Currently the code-base uses format-specifiers which do not always correctly match the width of the printf() arguments.
E.g. ``%zu`` is used for ``uint64_t``.
This should be cleaned up such that all format-specifiers for fixed-width types used the macros defined in ``<inttypes.h>``.

karlowich:
Now a far greater amount of the print statements uses the correct size format-specifier via the PRI* macros from `<inttypes.h>`

Resolves #98


safl (approved):
LGTM, thanks!

PR #345, opened 2023-09-13
feat(be/spdk): enable KV over nvmf

karlowich:
These additions allow for the creation of KV namespaces in the `nvmf_tgt` app, and the submission of KV cmds over nvmf.

To be clear, all of these patches only affect the SPDK `nvmf_tgt` app, and not the xNVMe SPDK backend.

### DIFF
--- a/cijoe/configs/debian-bullseye.toml
+++ b/cijoe/configs/debian-bullseye.toml
@@ -164,6 +164,12 @@ labels = [ "dev", "fabrics", "zns" ]
 driver_attachment = "userspace"
 
 [[devices]]
+uri = "127.0.0.1:4420"
+nsid = 3
+labels = [ "dev", "fabrics", "kvs" ]
+driver_attachment = "userspace"
+
+[[devices]]
 uri = "/tmp/xnvme-testfile.bin"
 nsid = 1
 labels = [ "file" ]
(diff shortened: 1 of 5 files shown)


safl (changes requested):
This is excellent, however, looks like the existing SPDK patches do not apply cleanly.
They need to be rebased for v23.01

cijoe/configs/debian-bullseye.toml
```diff
@@ -157,6 +157,14 @@ nsid = 2
 labels = [ "dev", "fabrics", "zns" ]
 driver_attachment = "userspace"
 
+[[devices]]
+uri = "127.0.0.1:4420"
+nsid = 3
+labels = [ "dev", "fabrics", "kvs" ]
+driver_attachment = "userspace"
+
+
```
sbates130272, line 166: Seems like too many newlines here?

karlowich: Thanks! Fixed.

safl:
@karlowich since the bump of SPDK is also needed by https://github.com/OpenMPDK/xNVMe/pull/350
How about you extract that into it's own PR? Then we can move that onto ``next`` for this PR as well as the other to consume?

karlowich:
@safl I agree that is a good idea. However, I think a new version of SPDK is just around the corner, and I think moving to that version is preferred.

safl:
> @safl I agree that is a good idea. However, I think a new version of SPDK is just around the corner, and I think moving to that version is preferred.

23.09 is out now, how about a PR taking that bump all the way? :)

safl (approved):
LGTM, thanks!

PR #346, opened 2023-09-15
feat(build/spdk): bump to v23.05

karlowich:
This requires a new dependency 'libarchive' for DPDK v23.03 which comes with SPDK v23.05.

Besides this, the `spdk_configure.sh` script has been updated to disable more things we don't need.


karlowich:
Something causes the ZNS tests to fail with 23.05.
However, they work with 23.01 so that is preffered.
The relevant commit is included in #345

PR #348, opened 2023-09-19
API Refactoring for portability

safl:
This is an exploration on how to improve the portability of the **xNVMe** C API and the language bindings for **Python** (https://pypi.org/project/xnvme/) and **Rust** (https://crates.io/crates/xnvme-sys).

# The need for API and FFI Improvements

**xNVMe** is using macros for various helpers in the API, this is great when consuming **xNVMe** from C and C++, however they are generally not available over FFI boundaries. These should be replaced with actual functions, such that FFI usage e.g. Python has access to these helpers as well.

In C, bit-field accessors are implementation-specific and translate poorly across FFI boundaries. Therefore, the goal is to either remove them entirely or provide equivalent accessors for FFI consumption. For NVMe definitions, specific helpers (e.g., for command construction) will be useful, where bit-shift manipulation and data structure access can be implemented.

Autodocs, the current use of bindgen and ctypeslib for generating bindings leaves docstrings behind. That is, when sitting in a Jypyter environment with the Python bindings, then there is absolutely no documentation for the functions other than their function signature. And in general, when a developer is writing code, then their LSP should be able to provide must better API descriptions.

C supports nested records (structs/unions), inline definitions, and anonymous records. While this works fine when consuming **xNVMe** via C++, it often translates poorly across FFI boundaries. Binding generators typically "hoist" inline/anonymous definitions to the top scope, assigning names like nameOfStruct_anon1. While this approach "works," it would be preferable to handle this directly in the C API with proper names, rather than relying on such workarounds.

# The NVMe Definitions

The NVMe specification is constantly evolving, and projects often manually translate the specification into source code, which is time-consuming and error-prone. To streamline this process, a tool (https://github.com/safl/senfd) is being developed to extract data from the specification documents for use by a code-emitter tool.

**xNVMe** utilizes this tool along with a code-emitter (https://github.com/safl/yace) to generate:

* Definitions and structures
* Command constructors/preppers
* Pretty-printers for structures (e.g., JSON and YAML output)

## Example of Generating the NVMe Model in JSON Format

Follow these steps to generate the NVMe model in .json format:

* Check out this PR.
* Navigate to the API directory: ``cd xnvme/emitter``
* Read the ``README.rst`` there (https://github.com/safl/xNVMe/blob/apigen/emitter/README.rst)

For more details on the commands being executed, refer to the ``xnvme/emitter/Makefile`` (https://github.com/safl/xNVMe/blob/apigen/emitter/Makefile).


safl:
One day there will hopefully be cycles to clean this up along with the improvements expect to come out of the https://github.com/safl/senfd and https://github.com/safl/yace

PR #351, opened 2023-09-26
Misc. fixes and refactoring

safl:
While experimenting with code-generation (https://github.com/OpenMPDK/xNVMe/pull/348) of the xNVMe C API and FFIs for Python and Rust, then several challenges appeared. These are addressed by refactoring. Misc. fixes are also provided which addresses build-warnings. more will come, however, no reason to hold these back.

### DIFF
--- a/lib/xnvme_spec.c
+++ b/lib/xnvme_spec.c
@@ -708,6 +708,7 @@ xnvme_spec_idfy_cs_fpr(FILE *stream, const struct xnvme_spec_idfy_cs *idfy, int 
 		wrtn += fprintf(stream, "iocsci: %d, ", i);
 		wrtn += fprintf(stream, "val: 0x%" PRIx64 ", ", iocscv->val);
 		wrtn += fprintf(stream, "nvm: %d, ", iocscv->nvm);
+		wrtn += fprintf(stream, "kv: %d, ", iocscv->kv);
 		wrtn += fprintf(stream, "zns: %d", iocscv->zns);
 		wrtn += fprintf(stream, " }");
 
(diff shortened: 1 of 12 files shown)


karlowich (dismissed):
This looks good!
I agree that this adds more clarity to the code

lib/xnvme_spec.c
```diff
@@ -43,7 +43,7 @@ xnvme_spec_log_health_fpr(FILE *stream, const struct xnvme_spec_log_health_entry
 
 	wrtn += fprintf(stream, "\n");
 	wrtn += fprintf(stream, "  crit_warn: %" PRIu8 "\n", log->crit_warn);
-	wrtn += fprintf(stream, "  comp_temp: %" PRIu16 "\n", log->comp_temp - 273u);
+	wrtn += fprintf(stream, "  comp_temp: %" PRIu16 "\n", (uint16_t)(log->comp_temp - 273u));
```
birkelund, line 46: By subtracting 273 you are converting from K to C. The result may very well be negative. Just let the compiler promote `comp_temp` to `int`.

lib/xnvme_spec.c
```diff
@@ -67,7 +67,7 @@ xnvme_spec_log_health_fpr(FILE *stream, const struct xnvme_spec_log_health_entry
 	wrtn += fprintf(stream, "  crit_comp_temp_time: %" PRIu32 "\n", log->crit_comp_temp_time);
 	for (int i = 0; i < 8; ++i) {
 		wrtn += fprintf(stream, "  temp_sens%u: %" PRIu16 "\n", i + 1,
-				log->temp_sens[i] ? log->temp_sens[i] - 273u : 0);
+				(uint16_t)(log->temp_sens[i] ? log->temp_sens[i] - 273u : 0));
```
birkelund, line 70: Same here.

safl: Right, thanks! Fixed.

lib/xnvme_spec.c
safl: Right, thanks! Fixed.

PR #352, opened 2023-09-27
feat(build/spdk): use debug for SPDK if enabled for xNVMe

karlowich:
If xNVMe is built with buildtype=debug, configure SPDK to use debug too.

I believe this is useful for debugging


safl (approved):
Great idea.

PR #350, opened 2023-09-25
add xnvme ctrl get registers

hmi-jeon:
add xnvme ctrl get registers

1. feat: add xnvme_ctrlr_get_registers to read controller registers
2. be:spdk: add function to get controller registeres
3. be:linux: add function to get controller registeres

PR #353, opened 2023-09-28
Pseudo commands support

vikash-k5:
Pseudo commands (ctrl reset, ns rescan and subsystem reset) support in xNVMe.

1. Adding ctrl reset, ns rescan and subsystem reset.
2. Add pseudo commands testing tool.

### DIFF
--- a/cijoe/tests/tools/test_xnvme.py
+++ b/cijoe/tests/tools/test_xnvme.py
@@ -236,3 +236,42 @@ def test_dsm(cijoe, device, be_opts, cli_args):
         f"xnvme dsm {cli_args} --nsid {device['nsid']} --ad --idw --idr --slba 0 --llb 1"
     )
     assert not err
+
+
+# For these three tests we request a device with labels 'ctrlr'
+@xnvme_parametrize(labels=["ctrlr"], opts=["be", "admin"])
+def test_subsystem_reset(cijoe, device, be_opts, cli_args):
+    if be_opts["be"] == "vfio":
+        pytest.skip(reason="[be=vfio] does not support pseudo commands")
+
+    err, _ = cijoe.run(
+        f"xnvme subsystem-reset {device['uri']} --be {be_opts['be']} --admin {be_opts['admin']}"
+    )
+
+    assert not err
+
+
+@xnvme_parametrize(labels=["ctrlr"], opts=["be", "admin"])
+def test_ctrlr_reset(cijoe, device, be_opts, cli_args):
+    if be_opts["be"] == "vfio":
+        pytest.skip(reason="[be=vfio] does not support pseudo commands")
+
+    err, _ = cijoe.run(
+        f"xnvme ctrlr-reset {device['uri']} --be {be_opts['be']} --admin {be_opts['admin']}"
+    )
+
+    assert not err
+
+
+@xnvme_parametrize(labels=["ctrlr"], opts=["be", "admin"])
+def test_namespace_rescan(cijoe, device, be_opts, cli_args):
+    if be_opts["be"] == "vfio":
+        pytest.skip(reason="[be=vfio] does not support pseudo commands")
+    if be_opts["be"] == "spdk":
+        pytest.skip(reason="[be=spdk] does not support namespace rescan")
+
+    err, _ = cijoe.run(
+        f"xnvme ns-rescan {device['uri']} --be {be_opts['be']} --admin {be_opts['admin']}"
+    )
+
+    assert not err
--- a/lib/xnvme_be_linux_nvme.c
+++ b/lib/xnvme_be_linux_nvme.c
@@ -14,6 +14,7 @@
 #include <linux/nvme_ioctl.h>
 #include <xnvme_be_linux.h>
 #include <xnvme_be_linux_nvme.h>
+#include <xnvme_spec.h>
 
 #ifdef XNVME_DEBUG_ENABLED
 static const char *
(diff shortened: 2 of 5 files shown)


safl (changes requested):
This is great, thank you @vikash-k5 as discussed offline, then I had some change-requests, I have applied them now. They include:

- [x] Correct formating (make format-all)
- [x] Remove NVME_IOCTL_* from public header
    - These should be provided by the on-system UAPI (``linux/nvme_ioctl.h``)
- [x] Dropped changes to ``lib/xnvme_topology.c``
    - These changed the error-handling. Although we currently do no make use of completion status-codes, then for other back-ends this might be useful. Thus, keeping that code in place as it does not prevent / interfere with Linux provided implementation.
- [x] Move cli-logic from test into "tools/xnvme.c"
    - The code added to the ``tests/pseudo_commands.c`` were exactly what the command-line tool needs, thus moved it to the cli instead. This can then be tested via the cijoe testcase.
- [x] Expand CIJOE tests with verification of the new 'xnvme' cli-tool sub-commands
    - ``cijoe/tests/tools/test_xnvme.py``
    - Expand with invocation of the new sub-commands
    
There is one item left, which will ensure that this runs as part for the CI-testing. Can you look into that one?

tools/xnvme.c
```diff
@@ -1551,6 +1599,36 @@ static struct xnvme_cli_sub g_subs[] = {
...
+		"Resets the subsystem",
+		"Resets the subsystem",
+		subsystem_reset,
+		{
+			{XNVME_CLI_OPT_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_URI, XNVME_CLI_POSA},
+		},
+	},
+	{
+		"ctrl-reset",
```
karlowich, line 1613: I think we should rename this and the other occurrences to "ctrlr-reset"

tools/xnvme.c
```diff
@@ -1551,6 +1599,36 @@ static struct xnvme_cli_sub g_subs[] = {
...
 		},
 	},
+	{
+		"subsystem-reset",
+		"Resets the subsystem",
+		"Resets the subsystem",
+		subsystem_reset,
+		{
+			{XNVME_CLI_OPT_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_URI, XNVME_CLI_POSA},
```
karlowich, line 1609: For each of these, we should add `XNVME_CLI_ADMIN_OPTS` to be able to specify `be` etc.

cijoe/tests/tools/test_xnvme.py
```diff
@@ -236,3 +236,35 @@ def test_dsm(cijoe, device, be_opts, cli_args):
...
+# For these three tests we request a device with labels 'bdev' and
+# "write_zeroes", as this will indirectly give us a kernel-managed NVMe device.
+@xnvme_parametrize(labels=["bdev", "write_zeroes"], opts=["be", "admin"])
+def test_subsystem_reset(cijoe, device, be_opts, cli_args):
+    err, _ = cijoe.run(f"xnvme subsystem-reset {device['uri']}")
+
+    if be_opts["be"] in ["linux"] and be_opts["admin"] in ["nvme"]:
+        assert not err, "Backend should support this, yet, yields an error"
+    else:
+        assert err, "Backend does not support this, yet, does not emit an error"
```
karlowich, line 250: These assert statements make no sense, we are currently not passing any options along with the command

lib/xnvme_be_linux_nvme.c
```diff
@@ -283,6 +284,41 @@ xnvme_be_linux_nvme_cmd_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf
...
+	}
+
+	err = ioctl(state->fd, ioctl_req);
+	if (err < 0) {
+		XNVME_DEBUG("FAILED: ioctl() err: %d", err);
+		return err;
+	}
+
+	ctx->cpl.status.sc = 0;
+	ctx->cpl.status.sct = 0;
```
karlowich, line 317: Do we need to set these explicitly? Perhaps we should use `ioctl_wrap()` instead of `ioctl()`?

cijoe/tests/tools/test_xnvme.py
```diff
@@ -236,3 +236,35 @@ def test_dsm(cijoe, device, be_opts, cli_args):
...
+
+
+@xnvme_parametrize(labels=["bdev", "write_zeroes"], opts=["be", "admin"])
+def test_subsystem_reset(cijoe, device, be_opts, cli_args):
```
karlowich, line 254: This test has the same name as the one above, as such we never actually test `subsystem-reset`

karlowich (changes requested):
This is looking pretty good. However it is weird that it fails like this:
```
# ERR: 'xnvme_controller_reset()': {errno: -1, msg: 'Operation not permitted'}
# ERR: 'ctrl-reset': {errno: -1, msg: 'Operation not permitted'}
```

I have left some comments.

karlowich:
It seems that part of the problem here lies in the fact that these commands are not supposed to be sent to namespaces e.g. `nvme0n1` but rather to the controller e.g. `nvme0`.
I am basing this on the behaviour of nvme-cli where:
`nvme reset /dev/nvme0n1` gives `Error: requesting reset on non-controller handle` 
`nvme reset /dev/nvme0` succeeds

karlowich:
For the subsystem reset, it would be very useful if we had support for something like the nvme-cli `nvme show-regs` or spdk `spdk_nvme_ctrlr_get_regs_*`

Then we could do something like this:
```
cap = spdk_nvme_ctrlr_get_regs_cap(ctrlr);
if (cap.bits.nssrs == 0) {
	return -ENOTSUP;
}
```
(from spdk [source code](https://github.com/spdk/spdk/blob/986fb7b3ef33fe066da9bc9cb91d86f8db4da0b1/lib/nvme/nvme_ctrlr.c#L1860))

karlowich:
Hi @vikash-k5,

I have addressed my own comments and pushed the changes required

safl:
> For the subsystem reset, it would be very useful if we had support for something like the nvme-cli `nvme show-regs` or spdk `spdk_nvme_ctrlr_get_regs_cap`
> 
> Then we could do something like this - getting the Controller Capabilities (figure 36 in base spec) and checking for support:
> 
> ```
> cap = spdk_nvme_ctrlr_get_regs_cap(ctrlr);
> if (cap.bits.nssrs == 0) {
> 	return -ENOTSUP;
> }
> ```
> 
> (from spdk [source code](https://github.com/spdk/spdk/blob/986fb7b3ef33fe066da9bc9cb91d86f8db4da0b1/lib/nvme/nvme_ctrlr.c#L1860))

That is what we have here: https://github.com/OpenMPDK/xNVMe/pull/350

We just needed this PR. to get in first, since it uses the restructured pseudo-command-infra, such that #350 could utilize it. So, once #353 is in, then we can rebase and refactor #350 to make use of pseudo-commands.

safl (approved):

karlowich (dismissed):

safl (dismissed):
Thanks @vikash-k5 and @karlowich !

I took the liberty of forcing a change on the PR, that is, use ``ioctl()`` rather than ``ioctl_wrap()``. This is because ``ioctl_wrap()`` is intended for command-passthru and the behavior of that specific ioctl, thus, it does not behave properly not the other ioctls.

In addition, then this PR https://github.com/OpenMPDK/xNVMe/pull/385 was added in preparation.

PR #354, opened 2023-09-29
feat(cijoe): add workflow for provisioning via git

karlowich:
This is useful for debugging in case creating/downloading artifacts is not an option


safl (approved):
LGTM

PR #356, opened 2023-10-04
feat(build/spdk): bump to v23.01

karlowich:
Bump SPDK since #350 uses a patch that is available in SPDK v23.01.

This requires a new dependency 'libarchive' for the version of DPDK which comes
with SPDK v23.01.

Besides this, the `spdk_configure.sh` script has been updated to disable
more things we don't need.


safl (approved):
LGTM, thanks!

PR #364, opened 2023-10-13
feat(build): only install manpages/completions for installed binaries

karlowich:
This avoids the case where manpages/completions are installed for binaries which have not been installed.

NOTE: With this change manpages/completions will always be installed for installed binaries. 
The reasoning for each is in the commit messages.


meson.build
```diff
@@ -176,13 +176,6 @@ if bash_completion_dep.found()
   subdir('toolbox'/'bash_completion.d')
 endif
 
-# man-pages: run the Makefile target 'make gen-man-pages' to update the man-pages
```
safl, line 179: How about, instead of removing the option of not installing man-pages, then leave it as is and instead replace``man/meson.build`` with:

```meson
if get_option('tests')
  subdir('tests')
endif
if get_option('tools')
  subdir('tools')
endif
if get_option('examples')
  subdir('examples')
endif
```

meson_options.txt
```diff
@@ -23,5 +23,4 @@ option('build_subprojects', type : 'boolean', value : true, yield : true)
 
 option('enable-ysnp', type: 'boolean', value: true)
 
-option('manpages', type: 'boolean', value: true, description: 'Install manpages for all CLI tools/examples/tests')
-option('bash_completion', type: 'feature', value: 'disabled', description: 'Install bash-completions for all CLI tools/examples/tests')
+option('force_completions', type: 'boolean', value: false, description: 'Install bash-completions at the default location if bash-completion is not available on the system')
```
safl, line 26: Perhaps rephrase as:

``Install bash-completions at the default location even when Bash is not available on the system``

Also, you remove two build-options and add another one. Thus, this change is closer to a refactoring and it is a build-system-API-breaking one. Thus, the commit message should signify that with:

``refactor(meson)!: replace options and only install manpages/completions for installed binaries``

meson.build
safl: Similar could be done for the completions.

I propose this, since it would be less of a functional change, in case the meson-options were still usable, then it would not be a 'build-api-breaking-change'. But just a ``fix(meson): ...``.

Anyway, either direction is fine with me.

meson_options.txt
karlowich: > Install bash-completions at the default location even when Bash is not available on the system

On Arch linux Bash does not include `bash-completion`, they are separate packages.
In the `meson.build` we check for `bash-completion`

karlowich:
I am unsure of what the right move is. 
I think for simplicity - the fewer meson options, the better - however, it would be stupid to remove an option that someone is actually using. 

Regardless of what we end up choosing, perhaps I should keep any removed options in the `meson_options.txt` saying that it is deprecated. That way we don't cause any build failures.

meson_options.txt
karlowich: > Also, you remove two build-options and add another one. 

I will edit this, so removing the man-pages option goes into the correct commit

safl:
> I am unsure of what the right move is. I think for simplicity - the fewer meson options, the better - however, it would be stupid to remove an option that someone is actually using.
> 

I agree, fewer options makes the build simpler. Once, the FreeBSD ports, RPM and Debian packaging starts to use build-options, then changing things increases "churn". Thus, since xNVMe is still < 1.0, then API breakage of various forms are to be expected. Thus, now is great time to remove the options. 

> Regardless of what we end up choosing, perhaps I should keep any removed options in the `meson_options.txt` saying that it is deprecated. That way we don't cause any build failures.

I would prefer a "hard" build-error to a silent-acceptance.  E.g. telling meson to not produce man-pages, and then it still does... that would just be annoying.

meson_options.txt
```diff
@@ -19,6 +19,4 @@ option('tools', type: 'boolean', value: true)
 
 option('build_subprojects', type : 'boolean', value : true, yield : true)
 
-option('enable-ysnp', type: 'boolean', value: true)
```
safl, line 24: Right, this seemed like a great idea a long time ago, however, it has not really been exercised, and gone stale.
When removing the meson-option, then also remove it in the code-base:
```
include/xnvme.h
include/xnvme_ysnp.h
```

safl (approved):
Great, thanks!

PR #365, opened 2023-10-16
Support nvme with char-device

hmi-jeon:
/dev/nvme0 is nvme char device.
But nsid of /dev/nvme0 is returned as 0xffffffff.
So check the uri to support nvme char device.


safl:
Hi @hmi-jeon,
Thanks for this, however, when reviewing it I came across a couple of things for this to function for a couple of different use-cases, thus integrated this PR https://github.com/OpenMPDK/xNVMe/pull/385 providing char-device support as well.
Thus closing this PR.

PR #368, opened 2023-10-17
feat(be/liburing): turn on batching by default

karlowich:
Instead of requiring `XNVME_QUEUE_BATCHING = 1` to enable batching, now we require `XNVME_QUEUE_BATCHING_OFF = 1` to disable batching


safl (approved):
LGTM.

PR #370, opened 2023-10-25
fix(cijoe,python): add 'vfio' to backend-test-matrix

safl:
This turned out to uncover a handful of issues:

- [x] Bump libvfn to v3 rc2
- [x] Fix usage of libvfn (incorrect usage of return-value)
- [x] Fix error handling in libvfn backend
- [ ] Feat: add support for iovec in libvfn backend
- [ ] Fix: enable use of libvfn backend with Python bindings
The two latter items are captured in issues.


karlowich:
This command fails with the vfio backend for some reason: `zoned mgmt-reset 0000:03:00.0 --dev-nsid 2 --be vfio --admin vfio --sync vfio --slba 0x0`

karlowich:
The latest commit fixes the issue

safl:
> The latest commit fixes the issue on the xNVMe side, unfortunately there is a bug in libvfn which needs to be resolved before this can work. This addressed here: #374

Thanks for getting to the bottom of this. The docker images are re-spinning now. Expecting this to pass once they are uploaded.

karlowich (approved):
Awesome!

PR #372, opened 2023-10-30
feat(build/spdk): bump to v23.09

karlowich:
This is currently blocked due to a SPDK limit on number of IOVECS for nvmf passthru which causes our Report Zones tests to fail.


karlowich:
The patch for SPDK is not finalized yet, however I put it here so the progress is visible.

safl (approved):
Awesome @karlowich !

PR #373, opened 2023-11-02
feat(be/spdk): support ctrlr and subsystem reset

karlowich:
Currently the SPDK api doesn't have an equivalent to rescan, thus this will give -ENOSYS.


karlowich:
Perhaps this should be included in #353

karlowich:
Closing this, since it has been included in #353

Issue #375, opened 2023-11-06
Consider refactoring vectored IO function signatures to use a metadata buffer


karlowich:
Currently they take metadata iovec, however according to the spec this can only have a single entry.
Because of this it seems simpler to just use a buffer instead.

Additionally, this aligns with the function signature for using iovecs with SPDK (which is not yet upstream - but very close).

karlowich:
As #376 has been merged, this has been resolved

PR #376, opened 2023-11-07
feat(be/spdk): add async/sync vectored io

karlowich:
This is dependent on the changes made in #372.

### DIFF
--- a/lib/xnvme_be_spdk_async.c
+++ b/lib/xnvme_be_spdk_async.c
@@ -44,6 +44,12 @@ xnvme_be_spdk_queue_init(struct xnvme_queue *q, int XNVME_UNUSED(opts))
 		return -ENOMEM;
 	}
 
+	queue->iov_payloads = malloc(sizeof(*queue->iov_payloads) * qopts.io_queue_size);
+	if (!queue->iov_payloads) {
+		XNVME_DEBUG("FAILED: malloc()");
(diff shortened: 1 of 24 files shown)


lib/xnvme_be_spdk_async.c
```diff
@@ -146,13 +146,97 @@ xnvme_be_spdk_async_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nb
...
+	}
+
+	state->iov = dvec;
+	state->iov_cnt = dvec_cnt;
+	state->iov_pos = 0;
+	state->iov_offset = 0;
+
+	queue->base.outstanding += 1;
+
+	err = spdk_nvme_ctrlr_cmd_iov_raw_with_md(
```
safl, line 222: We need to double-check the lifetime of the attributes accessed on ``state``. E.g. if it is used **after** the return of ``spdk_nvme_ctrlr_cmd_iov_raw_with_md()``, then we will have races on the structure.

karlowich:
This PR now addresses issue #375 as well.

This addition also brings to light that we never test the use of metadata - this should probably be fixed.

safl (approved):
This is great, thank you.
I have created an issue to remind us to add the missing tests https://github.com/OpenMPDK/xNVMe/issues/415 however let's go ahead and get this in.

PR #378, opened 2023-11-09
End to end data protection support

ankit-sam:
This series provides end to end data protection support in xNVMe.
This only support 16 and 64 bit protection information formats without storage tags.

This borrows all the helper functions required for E2E data protection, and big and little endian conversion from SPDK


lib/xnvme_crc.c
```diff
@@ -0,0 +1,587 @@
...
+		crc = crc64_rocksoft_refl_table[(uint8_t)crc ^ byte] ^ (crc >> 8);
+	}
+
+	return ~crc;
+}
+
+uint64_t
+xnvme_crc64_nvme(const void *buf, size_t len, uint64_t crc)
+{
+	return crc64_rocksoft_refl_base(crc, (const uint8_t *)buf, len);
```
birkelund, line 586: **Nitpick**. The CRC64 used in NVM Express is NOT called "Rocksoft". It's just called the "NVM Express 64b CRC" (it uses an NVMe unique polynomial). I know the above is because isa-l calls it that, but could do a `#define` here to point out that all the other implementations misunderstood something ;)

ankit-sam: Yeah correct, Keith also had the same comment in fio patches.

karlowich:
Hi @ankit-sam, the tests are failing because you are not installing the `libxnvme_pi.h` header, you need to add a line in `include/meson.build`.

safl (commented):
This is looking great. A couple of things.

- [ ] Conventional Commits
For the conventional-commits-style (https://www.conventionalcommits.org/en/v1.0.0/#summary), then it would be great with a rewording like the following: add "feat()" on the "scopes" to indicate that these things are extensions/features and use "/" as separator instead of ":". e.g. change from:
```
tools:lblk: add a sample test for end to end data protection
```
to
```
feat(tools/lblk): add a sample test for end to end data protection
```
- [ ] pi-namespace
On the introduction of the ``xnvme_pi`` namespace, then the following two functions should be renamed:

* ``xnvme_generate_pi()`` to ``xnvme_pi_generate()``
* ``xnvme_verify_pi()`` to ``xnvme_pi_verify()``

- [ ] ci-failure / build-issue on macOS
Then there is a build issue on macOS. Although macOS is not of that high priority, then the build-issue might be symptop of another problem, thus we should get it fixed.

- [ ] ISA-L fallback

I think it would be worth investigating whether we could use a, probably slow, fallback for the CRC algorithms, thus, when compiling xNVMe on a system where ISA-L is not available then it will use a fallback algorithm.

- [ ] Testing PI-support
Then last, it would be great with some form of tests of the PI-functionality that we can add to the CI. This could be a simple thing like invoking the ``lblk`` cli-tool and verifying the output and exit-code. Have a look at: ``cijoe/tests/tools/test_lblk.py``

In addition to this, then an explicit verification of functionality, such as ``xnvme_pi_generate()`` / ``xnvme_pi_verify()``, would be very useful as well, possiblt as an example-application or a test-application. However, we can add that on later as long as we have a basic verification in place now.

ankit-sam:
@safl isa-l is only enabled for 16 bit guard CRC, and this is only done if isal dependency is found. If isal is not present we fallback to the slower XOR based CRC calculation mechanism (borrowed from SPDK).
isal has support for 64 bit CRC, but we are currently not using it for xNVMe. I saw that even spdk is not using it. I will try to find a way to use it with xNVMe in future. 

I have made changes to the function names and commit messages as per your feedback, please take a look once if I addressed everything. I have also added a basic test in` cijoe/tests/tools/test_lblk.py`.

safl:
We should add a test of the e2e functionality along the lines of https://github.com/OpenMPDK/xNVMe/issues/404

safl:
@ankit-sam @karlowich just force-pushed a rebase with conflicts resolved after merge of https://github.com/OpenMPDK/xNVMe/pull/409

safl (approved):

PR #383, opened 2023-11-13
fix(rust): use header found via system_deps()

safl:
The path to the xNVMe library header was hardcoded. By doing so, 'cargo
build' would generate bindings. However, the build would fail when
invoked via 'cargo publish' as the hardcoded path would be incorrect.

Thus, instead of generating the bindings based on the relative path, the
bindings are generated based on the header found via
system_deps/include_paths. That is, against the library header, the
Rust-crate "system_deps" has verified that it is available in the
correct version.

### DIFF
--- a/rust/xnvme-sys/build.rs
+++ b/rust/xnvme-sys/build.rs
@@ -1,6 +1,26 @@
-fn main() {
+use std::error::Error;
+use std::io;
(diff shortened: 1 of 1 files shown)


karlowich (approved):
LGTM!

PR #388, opened 2023-11-16
Packaging

safl:
This is a quick stab at getting the packaging on the road, so to speak.
See https://github.com/OpenMPDK/xNVMe/issues/389 for a status on the packaging effort.

The files introduced here is a collection of the xNVMe packaging-effort, documenting the maintenance tasks, and the status of availability. The primary preparation work on this PR is the Debian packaging, in addition to the continous work on it, then the following tasks are intended:

- [x] Add package-verification via CI
  - Where applicable, e.g. running the package-creation process with verification-tools such as ``lintian``, that is, such that as much as possible is caught before packages are created for a release
- [x] This needs to be addressed: https://github.com/OpenMPDK/xNVMe/issues/401
  - An approach of encoding the version like: ``libnvme.so, libxnvme.so.0, libxnvme.so.0.7.3`` is now being pursued, which I assume comply with Debian and Redhat policy

All changes needed for packaging is done "upstream", that is, no patches inside e.g. the ``debian`` folder. This is done since the changes to comply with e.g. Debian policy so far only benefit the project/code-based and I am not expecting it to introduce in-compatible changes. If so, then we can explicitly address those.

Most of this has now gone in as preparation on ``next``, the remainder is the Debian packaging for which I await account-creation on https://salsa.debian.org/ as I have now learning that the proper process is to **not** keep the ``debian`` folder within the repository. When this is removed, then what remains:

- [x] Move the Debian packaging to [Salsa](https://salsa.debian.org/)
  - Currently moved here https://github.com/safl/xnvme-debpkg thus when the account on Salsa becomes available then things from the xNVMe repos side and CI is ready and just needs to change where it is retrieving the packaging from
- [x] CI scripts, testing the packaging scripts for Debian (clone salsa repos, pass it src-archive, build and lint)
- [x] Documentation, the ``README.rst``will be converted into a **packaging** section of the contributors documentation

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1058,3 +1058,46 @@ jobs:
         --form version="v${PROJECT_VERSION}" \
         --form description="xNVMe libraries and tools for NVMe" \
         "https://scan.coverity.com/builds?project=${PROJECT_NAME}"
+
+  packaging-debian:
+    needs: source-archive
+    runs-on: ubuntu-latest
+    container: ghcr.io/xnvme/xnvme-deps-debian-packaging:main
+    if: github.event_name == 'pull_request'
+
+    steps:
+    - name: Info
+      run: |
+        echo "Current branch is ${{ github.base_ref }}"
+        echo "Current tag is ${{ github.ref }}"
+
+    - name: Retrieve the xNVMe source archive
+      uses: actions/download-artifact@v3.0.1
+      with:
+        name: xnvme-src-archive
+
+    - name: Retrieve the xNVMe Debian package repository by branch
+      run: |
+        git clone https://salsa.debian.org/safl/xnvme.git debpkg
+        git -C debpkg checkout ${{ github.base_ref }}
+
+    - name: Prep
+      run: |
+        mv xnvme-src.tar.gz debpkg/input/xnvme-0.7.3.tar.gz
+
+    - name: Build
+      run: |
+        cd debpkg && make
+
+    - name: Upload debian packages
+      uses: actions/upload-artifact@v3.1.1
+      with:
+        name: packages-debian
+        path: |
+          debpkg/builddir/libxnvme*.deb
+          debpkg/builddir/xnvme-cli*.deb
+          debpkg/builddir/xnvme-full*.deb
+          debpkg/builddir/*.dsc
+          debpkg/builddir/*.buildinfo
+          debpkg/builddir/*.changes
+        if-no-files-found: error
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ builddir
 docs/build
 docs/autogen/enumerate_example
 docs/tutorial/dynamic_loading/enumerate_example
+compile_commands.json
 cscope.out
 tags
 cmake-build-debug/
--- a/Makefile
+++ b/Makefile
@@ -384,11 +384,12 @@ gen-man-pages:
 	@echo "## xNVMe: make gen-man-pages [DONE]"
 
 define tags-help
-# Helper-target to produce tags
+# Helper-target to produce tags and compile-commands
 endef
 .PHONY: tags
 tags:
 	@echo "## xNVMe: make tags"
+	@if [ -d "$(BUILD_DIR)" ]; then cp $(BUILD_DIR)/compile_commands.json .; fi;
 	@$(CTAGS) * --languages=C -h=".c.h" -R --exclude=$(BUILD_DIR) \
 		include \
 		lib \
--- a/docs/contributing/index.rst
+++ b/docs/contributing/index.rst
@@ -20,6 +20,7 @@ feedback on your `pull-request`_ or ask questions.
    contributing-toolbox.rst
    contributing-conventions.rst
    contributing-branches.rst
+   packaging/index.rst
    release-checklist.rst
 
 .. _Discord: https://discord.com/invite/XCbBX9DmKf
--- a/docs/contributing/packaging/aur/index.rst
+++ b/docs/contributing/packaging/aur/index.rst
@@ -1,0 +1,19 @@
+.. _sec-contributing-packaging-aur:
+
+Arch Linux (AUR - ``pacman``)
+-----------------------------
+
+To update the pkg::
+
+  # Clone the repository
+  git clone https://aur.archlinux.org/xnvme.git
+  cd xnvme
+
+Edit ``PKGBUILD`` -- update ``pkgver``, ``md5sums``, then::
+
+  updpkgsums
+  makepkg --printsrcinfo > .SRCINFO
+  git add PKGBUILD
+  git add .SRCINFO
+  git commit -s -m "bump to <major.minor.patch>"
+  git push origin
--- a/docs/contributing/packaging/brew/index.rst
+++ b/docs/contributing/packaging/brew/index.rst
@@ -1,0 +1,64 @@
+.. _sec-contributing-packaging-brew:
+
+macOS (``brew``)
+----------------
+
+Notes are provided on updating the package / brew-formula for **xNVMe**. At the
+bottom you will find notes on the initial creation of the formula. Do this:
+
+.. code-block:: bash
+
+    # Fork the **homebrew-core** repository
+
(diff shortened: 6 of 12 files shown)


lib/meson.build
```diff
@@ -96,6 +96,7 @@ xnvmelib_deps = [
 xnvmelib_shared = shared_library(
   'xnvme',
   xnvmelib_source,
+  soversion: meson.project_version().split('.')[0],
```
karlowich, line 99: If we are going to do it this way - we might as well put `version: meson.project_version()` this will give `libxnvme.so.0` but on linux it will also have `libxnvme.so.0.7.2`

.reuse/dep5
```diff
@@ -3,50 +3,18 @@ Upstream-Name: xNVMe
 Upstream-Contact: Simon A. F. Lund <os@safl.dk>
 Source: https://github.com/OpenMPDK/xNVMe
 
-Files: .github/codeql/codeql-config.yml .github/workflows/verify.yml .pre-commit-config.yaml meson_options.txt xnvme.spec.in
+Files: *
```
karlowich, line 6: Is this to say that all files have Samsung copyright and BSD-3-Clause license?
In that case you could probably remove the license from the lines below.

pkgs/brew/xnvme.rb
```diff
@@ -0,0 +1,69 @@
...
+      }
+    EOS
+
+    # Build the example using pkg-config for build-options
+    flags = shell_output("pkg-config xnvme --libs --cflags").strip
+    system ENV.cc, "test.c", "-o", "test", *flags.split
+
+    # Run it and check the output, this should produce the same output as the
+    # 'xnvme library-info' command, thus the output-assertion is the same
+    output = shell_output("./test info 1GB --be ramdisk")
```
karlowich, line 64: You are not using the arguments you give here

pkgs/deb/Makefile
```diff
@@ -0,0 +1,75 @@
+BUILDDIR=builddir
+URL=https://github.com/OpenMPDK/xNVMe/releases/download/v0.7.3/xnvme-0.7.3.tar.gz
+
+SRC_ROOT := $(BUILDDIR)/src
+TGZ_FILENAME := $(notdir $(URL))
+NAME := $(shell echo $(TGZ_FILENAME) | cut -d '-' -f 1)
+VERSION := $(shell echo $(TGZ_FILENAME) | cut -d '-' -f 2 | cut -d '.' -f 1-3)
```
karlowich, line 7: Perhaps it would be easier to set the version here and use that variable in the URL?

pkgs/deb/Makefile
```diff
@@ -0,0 +1,75 @@
+BUILDDIR=builddir
+URL=https://github.com/OpenMPDK/xNVMe/releases/download/v0.7.3/xnvme-0.7.3.tar.gz
+
+SRC_ROOT := $(BUILDDIR)/src
+TGZ_FILENAME := $(notdir $(URL))
+NAME := $(shell echo $(TGZ_FILENAME) | cut -d '-' -f 1)
```
karlowich, line 6: Is this ever going to be something other than `xnvme`?
Perhaps it would be easier to just hardcode

pkgs/deb/debian/libxnvme-dev.install
```diff
@@ -0,0 +1,4 @@
+usr/include/libxnvme*.h
+usr/lib/*/libxnvme.a
+usr/lib/*/libxnvme.so
```
karlowich, line 3: Does this need to be here? I see that libxnvme-dev depends on libxnvme0 - thus libxnvme-dev shouldn't be installing the `.so`

pkgs/deb/debian/control
```diff
@@ -0,0 +1,56 @@
...
+ libaio-dev,
+ liburing-dev,
+ meson (>= 0.58.0),
+Standards-Version: 4.6.2
+Homepage: https://xnvme.io/
+Vcs-Browser: https://github.com/OpenMPDK/xNVMe
+Vcs-Git: https://github.com/OpenMPDK/xNVMe.git
+Rules-Requires-Root: no
+
+Package: libxnvme0
```
karlowich, line 16: This package is called `xnvme` on all other distros, is there any reason to use a different name on debian?

karlowich (changes requested):
In general it looks good, I have left a couple of questions and nitpicks

lib/meson.build
safl: Yeah, but it does not fit the Debian conventions of "packages for shared libraries should be named the same as their library" e.g. ``libxnvme0`` then ``lintian`` will expect ``libxnvme.so.0``. If there is then also a ``libxnvme.so.0.7.3`` then ``lintian`` will complain:

https://mentors.debian.net/package/xnvme/

So, I went with only using ``SOVERSION`` encoding in the file-name and then mapping this to ``MAJOR-VER``.

pkgs/brew/xnvme.rb
safl: I don't think I follow?

The content of ``output`` is asserted. E.g. checking that expected output is produced from running the command.

pkgs/deb/debian/control
safl: Yes, to comply with Debian policy.

pkgs/deb/debian/libxnvme-dev.install
safl: Hmm, however, it might be what can enable the use of multiple naming conventions... gonna take it for a spin having "libxnvme.so.0" and "libxnvme.so.0.7.3"

lib/meson.build
safl: Guess i had done some other thing that led me to belief it was not policy-correct. Seems fine now.

.reuse/dep5
karlowich: I see. What I was proposing is removing the license lines from the code below, since it is the same license as the *. But perhaps `lintian` doesn't understand that.
```
Files: lib/xnvme_be_fbsd_async.c
Copyright: 2023 Fedor Uporov <fuporov.vstack@gmail.com>
License: BSD-3-Clause

Files: toolbox/third-party/windows/sys/queue.h
Copyright: 1991, 1993, The Regents of the University of California.
License: BSD-3-Clause

Files: toolbox/xnvme-driver.sh
Copyright: 2016 Intel Corporation
License: BSD-3-Clause
```

pkgs/brew/xnvme.rb
karlowich: The program you wrote, `test`,  doesn't take any arguments, but you provide `info 1GB --be ramdisk`

pkgs/deb/debian/control
karlowich: I see

.reuse/dep5
safl: Right ok, yeah, I don't think that will work.

pkgs/brew/xnvme.rb
safl: Ohh, you are right, I have made a note on https://github.com/OpenMPDK/xNVMe/issues/389 to fix this with the next release. The "xnvme.rb" in this PR is the version that first went into ``brew``:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/x/xnvme.rb

The intent is to just keep the file to show where it came from and future work to be done on the ``homebrew-core`` repository.

pkgs/deb/Makefile
safl: This got a make-over as the download functionality was never used, what is handy is invoking "make gen-src-archive" when fixing up the repository when packaging, thus changed it to do so instead.

pkgs/deb/Makefile
safl: This got a make-over as the download functionality was never used, what is handy is invoking "make gen-src-archive" when fixing up the repository when packaging, thus changed it to do so instead.

karlowich (commented):
This looks good! 
I do not have any comments, however I am a bit hesitant to approve, since this PR is huge and I have only gone through it once.
I will do a second pass later and approve.

PR #391, opened 2023-11-20
fix(rust): add loop-breakout when header is found

safl:
The search in include-headers did not break when the header was found, thus, it would only setup header_path correctly when it was located in the last iterated include_path.


karlowich (approved):
LGTM

PR #394, opened 2023-11-21
fix(nvm): change sdlba to slba for write_zeroes

karlowich:
According to the spec there is no such thing as sdlba for write zeroes, this is only used for copy


safl (approved):
Nice catch.

PR #395, opened 2023-11-21
Remove metadata buffer checks

ankit-sam:
These checks trigger I/O fail when metadata is enabled.


safl (approved):
Great, thanks.

karlowich (approved):
Looks good!

PR #398, opened 2023-11-22
update fdp tutorial for xnvme ioengine

ankit-sam:
1. Update fdp tutorial for xnvme ioengine.
2. Add a sample fio command and its output.


safl (approved):
Excellent!

PR #400, opened 2023-11-24
feat(cijoe/tests): test ioworker on fabrics

karlowich:
Previously, this was disabled because the performance of nvme/tcp was very slow - this has now been fixed, thus enabling the tests.


safl (approved):
Great!

Issue #226, opened 2023-01-18
Runtime instrumentation of dev


karlowich:
It is currently not possible to control the "mixin" of ``dev`` interfaces. It is not critical since each backend only has a single one of these interfaces. However, it should be cleaned up and instrumentation enabled.

*This issue is from the 'ISSUES' file, see #220*

karlowich:
Since it was decided that #402 wasn't needed, then I will close this as well.

PR #402, opened 2023-12-04
feat(be): add support for dev mixin 

karlowich:
This addresses #226


karlowich:
Closing this as there is currently no use-case, which makes it hard to justify the extensive change to the code-base

PR #405, opened 2023-12-07
Packaging preparation

safl:
This is a collection of changes for the preparation of packaging done here https://github.com/OpenMPDK/xNVMe/pull/388
Since these things are "good-to-go", whereas, the actual packaging is still on-going.


karlowich (approved):
Looks good!

PR #406, opened 2023-12-11
feat(tests/lblk): add test for LBA format

karlowich:


karlowich:
Closing this as testing the format command causes the Kernel to freeze (likely due to a race).
Besides this, the metadata functionality is being tested through the PI now.

PR #409, opened 2024-01-09
refactor: move identification and geo to device

safl:
- [x] Document the device assumptions
- [x] Add a test... verify that it indeed is capable of opening as ``XNVME_GEO_UNKNOWN``, having this test has caught several unconsidered edge-cases
- [x] Ensure that geometry type is checked
  - not unknown when something is expected... NVM when using e.g. lba_nbytes, ZNS when using e.g. ``nzones``.
- [x] Drop the "automatic" call to ``xnvme_dev_geo_geo()`` from the cli-library
   - Having this call prevents a cli-tool to auto-open without auto-identification/geo-derivation, this is the exact reason why we have moving it from ``xnvme_dev_open()`` to ``xnvme_dev_derive_geometry()`` and implicitly on ``xnvme_dev_get_geo()`` etc. however if the cli then goes as accesses it, then all that refactoring is rendered useless for cli-tools


lib/xnvme_dev.c
```diff
@@ -17,25 +17,17 @@ _zoned_geometry(struct xnvme_dev *dev)
...
 	struct xnvme_geo *geo = &dev->geo;
-	uint64_t nzones;
-	int err;
 
 	if (!zns->lbafe[0].zsze) {
 		XNVME_DEBUG("FAILED: !zns.lbafe[0].zsze");
 		return -EINVAL;
 	}
 
-	err = xnvme_znd_stat(dev, XNVME_SPEC_ZND_CMD_MGMT_RECV_SF_ALL, &nzones);
```
karlowich: If this method is not used here, then I think it can just be removed

include/xnvme_dev.h
```diff
@@ -22,7 +22,9 @@ struct xnvme_dev {
 	struct xnvme_be be;       ///< Backend interface
 	struct xnvme_ident ident; ///< Device identifier
 
-	uint8_t _pad[4];
+	uint8_t _pad[2];
+	uint8_t is_identified; ///< Device is identified
+	uint8_t is_derived;    ///< Device geometry is derived
```
karlowich, line 27: Perhaps these should be booleans - for the sake of clarity.

lib/xnvme_dev.c
```diff
@@ -426,48 +427,79 @@ xnvme_dev_pr(const struct xnvme_dev *dev, int opts)
 const struct xnvme_geo *
 xnvme_dev_get_geo(const struct xnvme_dev *dev)
 {
+	struct xnvme_dev *dropped = (struct xnvme_dev *)dev;
+
+	if (!dev->is_derived) {
+		dropped->is_derived = xnvme_dev_derive_geo(dropped);
```
karlowich, line 433: I get that this wouldn't be possible if `is_derived` was a bool, but perhaps that is for the better.
If this fails, then a "truthy" value will be stored in `is_derived` even though it is not actually derived.
In general I think the error handling in this commit, needs a second pass.

karlowich (changes requested):
I like the direction, some things needs a little more polish.

lib/xnvme_dev.c
safl: The function ``xnvme_znd_stat()`` is convenient helper-function provided by the xNVMe in the public API for users of ZNS capable devices.

include/xnvme_dev.h
safl: Agree, did this and...

lib/xnvme_dev.c
safl: ... renamed as well to better reflect what they actually represented... and **attempt** has been made...

docs/contributing/assumptions.rst
```diff
@@ -0,0 +1,106 @@
...
+  # Initialized by xnvme_dev_open() using helper xnvme_ident_from_uri(), then
+  # backends further manipulate the structure
+  dev.ident 
+
+  # These are setup by the backend as it is "openening" the device, here
+  # it might do simple things such as fstat() on Linux to determine whether the
+  # device-handle is a block-device or a char-device to indicate the ident.dtype
+  dev.ident.dtype
+
+  # The device-module assigns this in identification-helper _dev_idfy()
```
karlowich, line 84: > The device-module assigns this in identification-helper _dev_idfy()
The device-module assigns this in `the` identification-helper _dev_idfy()

docs/contributing/assumptions.rst
```diff
@@ -0,0 +1,106 @@
...
+  dev.is_derived # setup by xnvme_dev_get_*() access after calling xnvme_dev_derive_geometry()
+
+  # These are assignmed by the device-module at "identification-time"
+  dev.id.ctrlr
+  dev.id.ns
+  dev.idcss.ctrlr
+  dev.idcss.ns
+  dev.attempted_dev_idfy
+
+  # Populated by the device-module using the helper xnvme_dev_derive_geometry()
```
karlowich, line 102: > device-module

device module

karlowich (changes requested):
Looks good.
Provided a couple of comments

karlowich (approved):

PR #412, opened 2024-01-16
fix(gha): ignore status of Debian package build

safl:
The build of the Debian packaging is convenient to ensure that we do not introduce packaging regressions. However, in the current state we cannot introduce the required changes in the Salsa repository, in lockstep with changes on the xNVMe repository, which are needed to to adjust the packaging to the changes.

Thus, API-changes break the ci pipeline.

Thus, this change to keep the build of the package, but not fail upon errors.


mbrsamsung (approved):
Looks good

PR #413, opened 2024-01-16
feat(queue): add a default NOOP callback

safl:
The callback on the command-context initilization is set to NULL. When set to NULL, then it is the backend responsibility to check whether a valid function pointer is set upon "callback-time".

To avoid having to check this, then this change assigns a default NOOP callback.


mbrsamsung (approved):
Looks good, safl and I had a discussion of screen about the merits of explicit return values

PR #417, opened 2024-01-19
 Add new compatibility table to docs, changes from requirements.txt to poetry, fixes issues with sphinx theme not being included

mbrsamsung:
When adding a new section to the sphinx docs, some issues where detected with the theme stylesheet being included in the generated html file.

Add new compatibility section to docs
Changes from requirements.txt to poetry
Fixes issues with sphinx theme not being included


toolbox/pkgs/ubuntu-jammy.sh
safl, line 1: Why is this file touched?

docs/autogen/conf.py
safl, line 1: Excellent!

docs/autogen/.envrc
safl, line 1: If I do not have ``direnv`` installed, then this does not affect me, right?
I am OK with adding it, however, I would personally prefer using the helpers from the ``Makefile``.

safl (commented):
Excellent, a minor question inline.

docs/autogen/.envrc
mbrsamsung: Its just a configuration file, it will do nothing if you don't have direnv installed.

I see it as a developer tool that one can opt-in or not. Alternatively we could add the file to gitignore

toolbox/pkgs/ubuntu-jammy.sh
mbrsamsung: This is a mistake it will be removed from the commit

mbrsamsung: Its reverted, it was actually just a chmod +x on the file.

docs/autogen/conf.py
mbrsamsung: :+1:

safl (approved):
Thanks!

PR #421, opened 2024-01-26
fix(cijoe/configs): correct the linux.repository.remote

safl:
The remote was pointing to a non-existant repository. This fixes it by pointing it to https://github.com/torvalds/linux.git


karlowich (approved):
LGTM

PR #424, opened 2024-01-29
feat(mk): add helper producing documentation

safl:
For producing documentation, with output from command-examples included,
then different resources are needed such as a virtual machine with the
proper device configuration, and wall-clock time to execute all the
commands. A helper 'cijoe-do-docgen' takes care of this.

However, when reducing the scope to rendering the prose of the
documentation, then the build requirements is reduced to the setup of
a Python venv. and executing doxygen, breathe, a couple of scripts,
and sphinx-doc.

Such rendering is done by executing 'make' inside of 'docs/autogen', for
convenience, then this is lifted up into the Makefile in the root of the
repository. Such that 'make docs' will to a quick rendering, providing a
significantly simpler means to maintain and expand documentation.


mbrsamsung (approved):
Looks good

karlowich (approved):
lgtm

PR #432, opened 2024-02-13
fix(tests/copy): send correct amount of Source Range entries

karlowich:
Previously, we sent 128 Source Ranges even though we only filled out the
first. This PR fixes this.

Furthermore, the MSSRL is now used to limit the number of bytes per
Source Range.

### DIFF
--- a/tests/copy.c
+++ b/tests/copy.c
@@ -16,19 +16,20 @@
  *
  *  Constraints:
  *  Abide to mdts constraint
- *  Abite to COPY format zero max entries
+ *  Abide to COPY format zero max entries
  */
 static int
 test_copy(struct xnvme_cli *cli)
 {
 	struct xnvme_dev *dev = cli->args.dev;
 	const struct xnvme_geo *geo = xnvme_dev_get_geo(dev);
+	const struct xnvme_spec_idfy_ns *ns = xnvme_dev_get_ns(dev);
 	uint32_t nsid;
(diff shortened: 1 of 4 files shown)


mbrsamsung (approved):
Looks good to me

tests/copy.c
```diff
@@ -134,10 +134,6 @@ static struct xnvme_cli_sub g_subs[] = {
 			{XNVME_CLI_OPT_POSA_TITLE, XNVME_CLI_SKIP},
 			{XNVME_CLI_OPT_URI, XNVME_CLI_POSA},
 
-			{XNVME_CLI_OPT_NON_POSA_TITLE, XNVME_CLI_SKIP},
-			{XNVME_CLI_OPT_SLBA, XNVME_CLI_LOPT},
-			{XNVME_CLI_OPT_ELBA, XNVME_CLI_LOPT},
-
```
ankit-sam, line 140: Looks ok, but I am thinking does it makes sense to fix the `test_copy` function so that user can pass slba, sdlba and nlb. Maybe we can have some sort of check like this so that slba and sldba don't overlap
```
if (slba < sdlba)
    assert (slba + nlb <= sldba);
else
    assert (sldba + nlb <= slba);
```

karlowich: @ankit-sam I have tried to address this now

ankit-sam (approved):
Looks good

PR #436, opened 2024-02-22
fix(pkgs/gentoo): rename dev-util/meson to dev-build/meson

karlowich:
Seems the emerge package has been renamed.
The new one is here:
https://packages.gentoo.org/packages/dev-build/meson


mbrsamsung (approved):
Looks fine

PR #438, opened 2024-03-05
Refactor identification

karlowich:
I have cleaned up the identification part of the documentation, such that it is more clear which idfy structs are command set specific and which aren't.

There were multiple duplicated structs these have been cleaned up.


karlowich:
Accidentally pushed to the wrong remote - thus I have closed this and opened the correct PR.

PR #441, opened 2024-03-14
fix(core): correcting css struct size

Baekalfen:
Commit 59c6fe0b1ac2bd933464ba81389b080762bb8850 moved the css struct but replaced it with a struct of the wrong size. This reinstates the correct size.


karlowich (approved):
Nice catch! Thanks!

Baekalfen:
Deemed not related

PR #442, opened 2024-03-14
fix(cli,opts): remove option for css

karlowich:
The SPDK option for selecting command set is entirely unused in our code. This means that it has always been set to "one or more io command sets". This makes sense for our use case, and there seems to be nothing to gain, in limiting the IO to only NVM or even no IO (only admin).

As such, remove this unused opt and struct.

(Additionally, this opt follows an outdated convention and is generally just an eye-sore.)

### DIFF
--- a/include/libxnvme_cli.h
+++ b/include/libxnvme_cli.h
@@ -120,8 +120,6 @@ struct xnvme_cli_args {
 	uint32_t main_core;
 	const char *core_mask;
 
-	struct xnvme_opts_css css; ///< SPDK controller-setup: do command-set-selection
-
 	uint32_t use_cmb_sqs;
 	const char *adrfam;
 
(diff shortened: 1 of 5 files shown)


birkelund:
It would be prudent to verify CAP.CSS before setting CC.CSS. Not all controllers support "one or more command sets", in which case the 110b value in CC.CSS is reserved. Mostly it doesn't matter since such devices probably will just check that the least significant bit is 0, but you know...

karlowich:
@birkelund I get your point. Right now we support neither CAP nor CC - I think it would be more valuable to implement this properly instead of having the odd option here and there. So I think this is worth cleaning up, in an effort to simplify things slightly.
Besides, it only works for SPDK right now, which is not ideal either.

karlowich:
Closing this as it has been deemed not relevant

PR #443, opened 2024-03-15
feat(macos/driverkit): implement new DriverKit-based backend using MacVFN

Baekalfen:
This backend only works on macOS and depends on MacVFN being installed on the system. MacVFN uses Apple's DriverKit to hook a PCI device and by support of libvfn's code implements an NVMe driver.

MacVFN devices are currently enumerated by a counter in xNVMe, as bus-device-function is seemingly not available.


lib/xnvme_be_macos_driverkit_admin.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <xnvme_be.h>
+#include <xnvme_be_nosys.h>
+#ifdef XNVME_BE_MACOS_ENABLED
+#include <mach/mach_error.h>
+#include <CoreFoundation/CoreFoundation.h>
+
+#include <xnvme_dev.h>
+#include <xnvme_be_macos_driverkit.h>
+
+std::map<uint64_t, std::tuple<uint64_t, uint64_t>> address_mapping;
```
karlowich, line 18: I don't like this global variable, it seems like the kind of thing that should be inside the `xnvme_be_macos_driverkit_state`

karlowich (commented):
Some initial comments. I will have to take a closer look and test the code

safl (changes requested):
First off, thanks for diving into the world of DriverKit and user-space drivers on macOS, and especially, thanks for providing a xNVMe backend using your MacVFN project, cool stuff.

Here is some initial feedback. I have changed this PR to a Draft as there are some major things that need to be addressed:

- [x] There must be functional verification of the backend, in the same manner as with all the other backends
  - [x] Assuming the only means to test this is by setting up a physical Apple Silicon machine with prerequisites for the backend to function and the github-runner, well, then this must be done
  - [x] Add a cijoe-config defining a test-target (the physical Apple Silicon) with device identifiers, backend names etc.
  - [x] Adjust testcases according to the capabilities of the backend and the NVMe device used
  - [ ] Adjust the github workflow
- [ ] There must be documentation on how to install prerequisites, just as for all the dependencies
  - [ ] How to install MacVFN, configure macOS, auxiliary things needed, etc.

I have only skimmed the code, I have a lot of questions, but before spending more time on it, then we need to get this to a place where usage and requirements are documented and we can automate the functional verification of the backend.

Baekalfen:
> * [ ]  There must be functional verification of the backend, in the same manner as with all the other backends
>   
>   * [ ]  Assuming the only means to test this is by setting up a physical Apple Silicon machine with prerequisites for the backend to function and the github-runner, well, then this must be done

Ok, I'll wait until this has been procured.

lib/xnvme_be.c
```diff
@@ -383,6 +390,7 @@ xnvme_enumerate(const char *sys_uri, struct xnvme_opts *opts, xnvme_enumerate_cb
 			XNVME_DEBUG("INFO: skipping be: '%s'; !enabled", be.attr.name);
 			continue;
 		}
+		XNVME_DEBUG("INFO: found be: '%s'; enabled", be.attr.name);
```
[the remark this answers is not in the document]
Baekalfen: Yeah, probably not, I'll remove it. It's helpful though.

lib/xnvme_be_macos_driverkit.c
```diff
@@ -0,0 +1,80 @@
...
+		.async = &g_xnvme_be_macos_driverkit_async,
+		.check_support = xnvme_be_supported,
+	},
+	{
+		.mtype = XNVME_BE_ASYNC,
+		.name = "emu",
+		.descr = "Use emu pool for Asynchronous I/O",
+		.async = &g_xnvme_be_cbi_async_emu,
+		.check_support = xnvme_be_supported,
+	},
```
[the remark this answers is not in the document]
Baekalfen: I don't know. Lots of other backends have the emu interface too.

lib/xnvme_be_macos_driverkit_admin.cpp
```diff
@@ -0,0 +1,63 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <map>
```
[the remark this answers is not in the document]
Baekalfen: It's needed for `#include <xnvme_be_macos_driverkit.h>`

lib/xnvme_be_macos_driverkit_async.cpp
```diff
@@ -0,0 +1,208 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+
+#ifdef XNVME_BE_MACOS_ENABLED
+#include <map>
```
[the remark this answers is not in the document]
Baekalfen: It's needed for `#include <xnvme_be_macos_driverkit.h>`

lib/xnvme_be_macos_driverkit_admin.cpp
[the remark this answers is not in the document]
karlowich: In that case, shouldn't it be inside `xnvme_be_macos_driverkit.h` instead?

lib/xnvme_be_macos_driverkit_async.cpp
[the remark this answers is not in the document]
karlowich: In that case, shouldn't it be inside `xnvme_be_macos_driverkit.h` instead?

safl:
> > * [ ]   There must be functional verification of the backend, in the same manner as with all the other backends
> >   
> >   * [ ]   Assuming the only means to test this is by setting up a physical Apple Silicon machine with prerequisites for the backend to function and the github-runner, well, then this must be done
> 
> Ok, I'll wait until this has been procured.

Could you address the need for documentation and testing? With those things in place, then other can run it on their own HW while we are waiting for dedicated testing HW.

Baekalfen:
> Could you address the need for documentation and testing? With those things in place, then other can run it on their own HW while we are waiting for dedicated testing HW.

In the meanwhile, you can find the instructions on: https://github.com/SamsungDS/MacVFN

lib/xnvme_be_macos_driverkit_admin.cpp
Baekalfen: I think I did it because it's a C++ dependency. And `struct xnvme_be_macos_driverkit_state` is used from both cpp and c files.

I've made a workaround now.

lib/xnvme_be_macos_driverkit_admin.cpp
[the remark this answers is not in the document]
Baekalfen: It's tricky, because it cannot be included from within `extern "C" {`, as it's a C++ dependency. And `xnvme_be_macos_driverkit.h` is included from both cpp and c files.

I've made a workaround now.

Baekalfen:
> I have only skimmed the code, I have a lot of questions, but before spending more time on it, then we need to get this to a place where usage and requirements are documented and we can automate the functional verification of the backend.

I would rather have the feedback now, so I can start making adjustments, and not drag out the process or do things that have to be reverted anyway.

lib/xnvme_be_macos_driverkit.c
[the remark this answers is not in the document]
karlowich: Yes, it seems that we don't have a pattern for which do.
I personally don't see why you would need an emulated async I/O interface, when you have a "real" one.
But maybe that is a matter of taste.

lib/xnvme_be_macos_driverkit_admin.cpp
karlowich: That's better in my opinion

safl:
@NaddiNadja, we want to enable testing for this functionality using the MacBook in the local lab. Your help with this when you have some availability would be greatly appreciated!

To-Do for Setting Up the MacBook:

- [x] Get access, wipe it, and connect it in the local CI lab.
- [x] Install the MacVFN driver as described in this PR.
- [x] Install miscellaneous development tools as outlined in this PR.
- [x] Install the GitHub runner, or alternatively, install the runner on the host used for bench and latency testing, SSH-ing into the MacBook instead of running it directly.

With the above setup, it should be possible to:

- [x] Set up GitHub CI to run the verify job using the CIJOE configuration that describes the system.

safl:
@Baekalfen @birkelund the work that you are doing on:

- https://github.com/SamsungDS/libvfn/pull/23
- https://github.com/SamsungDS/libvfn/pull/10

Would this lead to the libvfn backend of xNVMe being able to handle macvfn, making this PR superfluous?

Baekalfen:
> @Baekalfen @birkelund the work that you are doing on:
> 
> * [Prep for DriverKit SamsungDS/libvfn#23](https://github.com/SamsungDS/libvfn/pull/23)
> * [Add macOS/DriverKit through MacVFN SamsungDS/libvfn#10](https://github.com/SamsungDS/libvfn/pull/10)
> 
> Would this lead to the libvfn backend of xNVMe being able to handle macvfn, making the backend introduced with this PR superfluous?

No, there will always be special handling of MacVFN, as everything is RPC calls.

safl:
> > @Baekalfen @birkelund the work that you are doing on:
> > 
> > * [Prep for DriverKit SamsungDS/libvfn#23](https://github.com/SamsungDS/libvfn/pull/23)
> > * [Add macOS/DriverKit through MacVFN SamsungDS/libvfn#10](https://github.com/SamsungDS/libvfn/pull/10)
> > 
> > Would this lead to the libvfn backend of xNVMe being able to handle macvfn, making the backend introduced with this PR superfluous?
> 
> No, there will always be special handling of MacVFN, as everything is RPC calls.

Great, thanks for clarifying!

lib/xnvme_be_macos_driverkit_mem.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+	// nbytes has to be a multiple of alignment. Therefore, we round up to the nearest
+	// multiple.
+	XNVME_DEBUG("Unaligned size: %lx", nbytes);
+	nbytes = (1 + ((nbytes - 1) / sz)) * (sz);
+	XNVME_DEBUG("Aligned size: %lx", nbytes);
+
+	void* ptr = xnvme_buf_virt_alloc(sz, nbytes);
+	size_t output_cnt = sizeof(token);
+	*(uint64_t*) ptr = nbytes; // Put actual size inside allocated memory. Driver on the other end doesn't correctly get this.
```
birkelund: Can't this be part of a struct used with `IOConnectCallStructMethod()` instead of passing the pointer directly?

I.e.,

```c
struct alloc_buffer_request {
  void *ptr;
  size_t len;
};

IOConnectCallStructMethod(..., NVME_ALLOC_BUFFER, req, 1, &token, &output_cnt);
```

As I understand it, the remote side (MacVFN) knows what structs to expect, so it can unpack it nicely. Or have I misunderstood the protocol?

birkelund: On the MacVFN side, I see that you are retrieving the argument directly as an `IOMemoryDescriptor`. What governs when what is passed is just an `OSData` and when it is an `IOMemoryDescriptor`? The documentation is a little vague on that point... ;)

birkelund: Mads explained offline. No, it is not trivial to just pass a nice "parameter structure" due to the way IOKit maps the arguments. Bottomline is that you just have to suck it up and pack it like above.

NaddiNadja:
@Baekalfen, I've created a github actions job that tests xNVMe on our macOS machine in this PR. Some of the tests fail with a segfault, so I need you to take a look and assess what needs to be done. 

The test results are uploaded as the artifacts here: https://github.com/xnvme/xnvme/actions/runs/11125964021?pr=443

Baekalfen:
> @Baekalfen, I've created a github actions job that tests xNVMe on our macOS machine in this PR. Some of the tests fail with a segfault, so I need you to take a look and assess what needs to be done.
> 
> The test results are uploaded as the artifacts here: https://github.com/xnvme/xnvme/actions/runs/11125964021?pr=443

Looks pretty good, well done!

* It fails on enum open. As I see it, it tries to open the device multiple times. This is not supported.
* It fails on vector operations. That's expected, as it isn't implemented.
* `lblk` fails because it tried to free a null-pointer. This has been fixed.
* Different log calls fail. Looks like an error propagated from hardware, and not the driver? I can see some log calls are working and some are not. What are the expected answers?

NaddiNadja:
> * It fails on enum open. As I see it, it tries to open the device multiple times. This is not supported.

Fixed

> * It fails on vector operations. That's expected, as it isn't implemented.

Woops, I thought I had gotten all the iovec tests - fixed too

Baekalfen:
@safl What's the status? Seems like @NaddiNadja has cleared all the checkmarks. Docs were also added.

safl:
> @safl What's the status? Seems like @NaddiNadja has cleared all the checkmarks. Docs were also added.

Right, thanks to @NaddiNadja for the amazing work getting this up and running.  
Now that it's operational, we can start looking at potential changes.  
One major concern is the introduction of C++, which I'd like to avoid.  
@karlowich, @birkelund, and I discussed a couple of possible approaches, these are, in order of preference:

1. **Provide a macvfn library (e.g., libmacvfn)**  
   This library would offer a C interface to interact with the user-space interface of macvfn.  
   - By doing so, the driverkit-backend won't be "special" and can implement the backend interface and introduce dependencies (libmacvfn) like the other xNVMe backends.

2. **Remove the C++-dependent code from the driverkit-backend**  
   - This could be achieved by moving the address-mapping out of the driverkit-backend and into macvfn.  
   - Is this feasible? It seems the address_mapping/look-ups could potentially be done on the driver side.

3. **Move the driverkit-backend into a subproject**  
   - The subproject would implement the xNVMe backend interface and use the C++ toolchain.  
   - It could reside in the xNVMe repository or be moved out entirely.  
   - xNVMe would link to it as a plain object without requiring a C++ toolchain.

We can discuss in greater detail offline, if that is preferable, or we can continue here.

Baekalfen:
> 1. **Provide a macvfn library (e.g., libmacvfn)**
>    This library would offer a C interface to interact with the user-space interface of macvfn.
>    
>    * By doing so, the driverkit-backend won't be "special" and can implement the backend interface and introduce dependencies (libmacvfn) like the other xNVMe backends.
> 2. **Remove the C++-dependent code from the driverkit-backend**
>    
>    * This could be achieved by moving the address-mapping out of the driverkit-backend and into macvfn.
>    * Is this feasible? It seems the address_mapping/look-ups could potentially be done on the driver side.
> 3. **Move the driverkit-backend into a subproject**
>    
>    * The subproject would implement the xNVMe backend interface and use the C++ toolchain.
>    * It could reside in the xNVMe repository or be moved out entirely.
>    * xNVMe would link to it as a plain object without requiring a C++ toolchain.
> 
> We can discuss in greater detail offline, if that is preferable, or we can continue here.

1. MacVFN is not really a library. It's more of an independent app/server as it has its own processes etc. and have a lifetime that supersedes its clients. The only redistributable part is the header file, which is self-contained. xNVMe is not linking against MacVFN when building. So making the single header-file into a library seems overkill.

2. We can avoid C++, as it was started because we handled [IOBufferMemoryDescriptor](https://developer.apple.com/documentation/driverkit/iobuffermemorydescriptor)s. But that has been changed since.

As for the address-mapping. It's not a part of MacVFN, as it's a unique problem for xNVMe, and MacVFN has no general need for it. The API is like this:

* Alloc buffer in xNVMe (`aligned_alloc`, get `void *`)
* Map buffer with MacVFN (get token)
* Change data in buffer
* Send NVMe write to MacVFN by supplying token and offset

This is part of xNVMe because it has an API that only exposes a `void *` to its users on `xnvme_buf_alloc`, and its users are allowed to provide an offset pointer inside a buffer for an NVMe command. The `void *` is not known or accessible from MacVFN, as it is in a different address space. The code in xNVMe therefore has to resolve the user-supplied `void *` to its base pointer and find the token for MacVFN. 

You didn't mention the `MacVFNShared.h` file. If there was no C++ in the backend code, would that then be the solution, to put `MacVFNShared.h` somewhere in the xNVMe repo?

3. Not really sure how this would work. This would be a new Git repo with an xNVMe-to-MacVFN wrapper?

safl:
Thanks for all the updates, @Baekalfen!

You mentioned offline that macvfn versioning isn’t planned for the near term. To provide some stability, could you instead add a tag in the macvfn repository? This would allow us to reference a specific "frozen" point in the macvfn repository within **xNVMe** documentation and CI, indicating that **xNVMe** version `vX.Y.Z` is tested against this tag in macvfn.

Could you also ensure that this tag remains unchanged and consistently available? Since versioning isn’t the objective, a tag like `for_xnvme_v0.9.6`—to denote compatibility with the next **xNVMe** release—could work well.

**update** fyi. @Baekalfen then @NaddiNadja took a look at the CI and it seems like something is wrong with macvfn. That is, the **ci** does what it is supposed to, but ``xnvme enum`` gives nothing when logging in manually on the test-machine. Might be easiest to debug it together.

Baekalfen:
> could you instead add a tag in the macvfn repository?

I'll add a tag

> Could you also ensure that this tag remains unchanged and consistently available?

Sure

> Since versioning isn’t the objective, a tag like `for_xnvme_v0.9.6`—to denote compatibility with the next **xNVMe** release—could work well.

I'll probably do `xnvme_v0.9.6`. How do we make sure that the version I tag will align with the one that it gets merged and released with?

> **update** fyi. @Baekalfen then @NaddiNadja took a look at the CI and it seems like something is wrong with macvfn.

Solved

safl:
> > could you instead add a tag in the macvfn repository?
> 
> I'll add a tag
> 
> > Could you also ensure that this tag remains unchanged and consistently available?
> 
> Sure
> 
> > Since versioning isn’t the objective, a tag like `for_xnvme_v0.9.6`—to denote compatibility with the next **xNVMe** release—could work well.
> 
> I'll probably do `xnvme_v0.9.6`. How do we make sure that the version I tag will align with the one that it gets merged and released with?

We ensure that, by putting the macvfn header available from your tag into ``toolbox/third-party/``, use the tagged-version in the CI. Then, if macvfn changes, then the following needs to be addressed in a PR to xNVMe:

- ``toolbox/third-party`` update to match the version from the new tag
- backend adjusted to macvfn changes -- if needed
- documentation -- describing the tag etc.
- ci -- workflow updates, change of tag

> 
> > **update** fyi. @Baekalfen then @NaddiNadja took a look at the CI and it seems like something is wrong with macvfn.
> 
> Solved

Baekalfen:
This branch and MacVFN updated. MacVFN tagged with `xnvme_v0.9.6` as specified.

@NaddiNadja there's something with the CI again.

Otherwise, I cannot see anything pending.

NaddiNadja:
@Baekalfen, yes you have overwritten all my changes, so it makes sense that it didn't work. I had the changes locally still, so I've added them again.

Baekalfen:
> @Baekalfen, yes you have overwritten all my changes, so it makes sense that it didn't work. I had the changes locally still, so I've added them again.

Sorry, hadn't noticed there was something to pull. When I auto-squashed all the fixups, I expected to `git push -f` anyway.

docs/backends/xnvme_be_macos.rst
```diff
@@ -0,0 +1,22 @@
+.. _sec-backends-macos:
+
+macOS
+=====
+
+The macOS/DriverKit backend in xNVMe `requires MacVFN to be installed <https://github.com/samsungds/macvfn>`_.
+
```
safl, line 7: Add a note here about the tag to use when installing macvfn.

include/xnvme_be_macos_driverkit.h
```diff
@@ -0,0 +1,90 @@
...
+#include <xnvme_be.h>
+
+#include "MacVFNShared.h"
+
+/**
+ * Internal representation of XNVME_BE_MACOS_DRIVERKIT state
+ *
+ * NOTE: When changing this struct, ensure compatibility with 'struct xnvme_be_cbi_state'
+ */
+struct buffer;
```
safl: Move the buffer up before the doc-string, since the docstring should be before the thing that it documents.

lib/xnvme_be_macos_driverkit_admin.c
```diff
@@ -0,0 +1,61 @@
...
+		assert(buf);
+		_buf_base_offset = ((uint64_t)dbuf) - buf->vaddr;
+	}
+	NvmeSubmitCmd cmd = {
+		.queue_id = 0,
+		.dbuf_nbytes = dbuf_nbytes,
+		.dbuf_offset = _buf_base_offset,
+		.dbuf_token = (uint64_t)dbuf - _buf_base_offset,
+	};
+	NvmeSubmitCmd cmd_return;
```
safl: Add a space once you are done declaring variables and unpacking variables.

lib/xnvme_be_macos_driverkit_admin.c
```diff
@@ -0,0 +1,61 @@
...
+
+int
+xnvme_be_macos_driverkit_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes,
+			       void *XNVME_UNUSED(mbuf), size_t XNVME_UNUSED(mbuf_nbytes))
+{
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)ctx->dev->be.state;
+
+	uint64_t _buf_base_offset = 0;
+	if (dbuf) {
```
safl, line 28: Move this below variable declaration. This would be preferable to return error instead of asserting. Something like:

```c
struct buffer *buf;

if (dbuf) {
  buf = buffer_find(state->buffers, (uint64_t)dbuf);
  if (!buf) {
   return -appropriate_error_code;
  }
}
```

lib/xnvme_be_macos_driverkit_admin.c
```diff
@@ -0,0 +1,61 @@
...
+		.dbuf_nbytes = dbuf_nbytes,
+		.dbuf_offset = _buf_base_offset,
+		.dbuf_token = (uint64_t)dbuf - _buf_base_offset,
+	};
+	NvmeSubmitCmd cmd_return;
+	memcpy(cmd.cmd, &ctx->cmd, sizeof(struct xnvme_spec_cmd));
+	cmd.backend_opaque = (uint64_t)ctx;
+
+	size_t output_cnt = sizeof(NvmeSubmitCmd);
+	kern_return_t ret =
```
safl: When declaring variables, then the initialization done must be:

* Variable unpacking, such as
  - ``struct xnvme_be_macos_driverkit_state *state = (struct xnvme_be_macos_driverkit_state *)ctx->dev->be.state``
* Static initialization, such as
  - ``uint64_t _buf_base_offset = 0;``
  
Things that require error-handling, come after this. Ordering at the top-scope should be:

* Variable declaration with initialization using argument unpacking
* Variable declaration with static initializer
* Variable declaration without initialization

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,196 @@
...
+		    "Incorrect size")
+
+int
+xnvme_be_macos_driverkit_queue_init(struct xnvme_queue *_queue, int XNVME_UNUSED(opts))
+{
+	struct xnvme_queue_macos_driverkit *q = (struct xnvme_queue_macos_driverkit *)_queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)_queue->base.dev->be.state;
+
+	int qpid;
```
safl, line 37: Add a space here.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,196 @@
...
+	qpid = _xnvme_be_driverkit_create_ioqpair(state, q->base.capacity, 0);
+	if (qpid < 0) {
+		XNVME_DEBUG("FAILED: nvme_create_ioqpair(); err: %d", qpid);
+		return qpid;
+	}
+	q->id = qpid;
+
+	mach_vm_size_t size = 0;
```
safl: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,196 @@
...
+{
+	int queue_ret;
+	struct xnvme_queue_macos_driverkit *q = (struct xnvme_queue_macos_driverkit *)queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)queue->base.dev->be.state;
+	uint64_t qid[2] = {q->id, q->id}; // Submission queue, completion queue
+
+	kern_return_t ret =
+		IOConnectCallScalarMethod(state->device_connection, NVME_POKE, qid, 2, NULL, 0);
+	if (ret != kIOReturnSuccess) {
```
safl, line 105: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,196 @@
...
+{
+	struct xnvme_queue_macos_driverkit *q =
+		(struct xnvme_queue_macos_driverkit *)ctx->async.queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)ctx->dev->be.state;
+
+	NvmeSubmitCmd cmd = {0};
+	int queue_ret;
+	uint64_t _buf_base_offset;
+	cmd.queue_id = state->qid_sync;
```
safl, line 140: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,218 @@
...
+
+#include <xnvme_dev.h>
+
+int
+xnvme_be_macos_driverkit_enumerate(const char *sys_uri, struct xnvme_opts *opts,
+				   xnvme_enumerate_cb cb_func, void *cb_args)
+{
+	XNVME_DEBUG("xnvme_be_macos_driverkit_enumerate()");
+
+	if (sys_uri) {
```
safl, line 31: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,218 @@
...
+int
+_xnvme_be_driverkit_create_ioqpair(struct xnvme_be_macos_driverkit_state *state, int qd, int flags)
+{
+	NvmeQueue ioqpair_return;
+	kern_return_t ret;
+	size_t output_cnt;
+	unsigned int qid = __builtin_ffsll(state->qidmap);
+	// int err;
+
+	if (qid == 0) {
```
safl, line 101: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,93 @@
...
+{
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)dev->be.state;
+	if (!buf) {
+		return;
+	}
+	struct buffer *_buf = buffer_find(state->buffers, (uint64_t)buf);
+	assert(_buf);
+
+	XNVME_DEBUG("buf token: %llx", (uint64_t)_buf->vaddr);
```
safl: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,218 @@
...
+
+int
+xnvme_be_macos_driverkit_dev_open(struct xnvme_dev *dev)
+{
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)dev->be.state;
+	int qpid;
+
+	kern_return_t err;
+	io_service_t service;
```
safl, line 170: See notes on declaration / initialization order and error-handling.

lib/xnvme_be_macos_driverkit_admin.c
```diff
@@ -0,0 +1,61 @@
...
+int
+xnvme_be_macos_driverkit_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes,
+			       void *XNVME_UNUSED(mbuf), size_t XNVME_UNUSED(mbuf_nbytes))
+{
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)ctx->dev->be.state;
+
+	uint64_t _buf_base_offset = 0;
+	if (dbuf) {
+		struct buffer *buf = buffer_find(state->buffers, (uint64_t)dbuf);
```
safl: In general, when something requires error-handling, then move that down to after variable declaration.
E.g. declare ``struct buffer *buf`` at the top of the block-scope, and after declaration you initialize the variables that need error-handling. e.g.:

```c
if (!buf) {
...
} 
```

lib/xnvme_be_macos_driverkit_sync.c
```diff
@@ -0,0 +1,85 @@
...
+#ifdef XNVME_BE_MACOS_ENABLED
+#include <xnvme_be_macos_driverkit.h>
+#include <xnvme_dev.h>
+
+static int
+cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes, void *mbuf, size_t mbuf_nbytes)
+{
+	NvmeSubmitCmd cmd = {0};
+	NvmeSubmitCmd cmd_return = {0};
+	uint64_t _buf_base_offset;
```
safl, line 23: See notes on declaration / initialization order and error-handling.

safl (changes requested):
This is looking great, @Baekalfen! Thanks for making those changes—we're almost there! :)
I've left some inline comments with a couple of minor points and a general pattern on variable declaration/initialization order and error-handling that I hope you can address. These are the last items I have—otherwise, LGTM!

docs/backends/xnvme_be_macos.rst
Baekalfen: It's in the MacVFN repo. No need to maintain the same instructions in two places.

include/xnvme_be_macos_driverkit.h
Baekalfen: Done

lib/xnvme_be_macos_driverkit_admin.c
Baekalfen: Done.

FYI, in GitHub you can click and drag to highlight the specific lines. I initially didn't get the context.

lib/xnvme_be_macos_driverkit_admin.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_admin.c
Baekalfen: Outdated after changes.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_admin.c
Baekalfen: The reason is that `buf` doesn't exist after this scope. I've changed it to what you describe.

lib/xnvme_be_macos_driverkit_sync.c
Baekalfen: Done.

safl:
@NaddiNadja Can you help with the ci error?

Baekalfen:
> @NaddiNadja Can you help with the ci error?

Looks like a random error, don't think it's PR-related. I don't have the permissions, but maybe start by re-running the test.

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,96 @@
...
+		return;
+	}
+	_buf = buffer_find(state->buffers, (uint64_t)buf);
+	assert(_buf);
+
+	XNVME_DEBUG("buf token: %llx", (uint64_t)_buf->vaddr);
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_DEALLOC_BUFFER,
+					&_buf->vaddr, sizeof(_buf->vaddr), NULL, &_output_cnt);
+	assert(ret == kIOReturnSuccess);
+	// buffer_remove(state->buffers, (uint64_t) buf);
```
karlowich: If this is not needed let's remove it

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,96 @@
...
+	ptr = xnvme_buf_virt_alloc(sz, nbytes);
+	buf = (uint64_t *)ptr;
+	buf[0] = nbytes;        // Put actual size inside allocated memory. Driver on the other end
+				// doesn't correctly get this.
+	buf[1] = (uint64_t)ptr; // Our token for the memory reference
+
+	XNVME_DEBUG("buf token: %llx", (uint64_t)ptr);
+	kern_return_t ret = IOConnectCallStructMethod(state->device_connection, NVME_ALLOC_BUFFER,
+						      ptr, nbytes, NULL, 0);
+	assert(ret == kIOReturnSuccess);
```
karlowich: A general question:
Why do we need these asserts? Is "normal" error handling like in the rest of the other backends not sufficient?

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,202 @@
...
+	NvmeSubmitCmd cmd = {0};
+	int queue_ret;
+	struct buffer *buf;
+	uint64_t _buf_base_offset;
+
+	cmd.queue_id = state->qid_sync;
+
+	switch (ctx->cmd.common.opcode) {
+	case XNVME_SPEC_NVM_OPC_READ:
+		break;
```
karlowich, line 144: In my opinion there's no need for the empty cases in the switch statement

cijoe/tests/logic/test_xnvme_tests_enum.py
```diff
@@ -7,6 +7,8 @@
 def test_open(cijoe, device, be_opts, cli_args):
     if be_opts["be"] == "ramdisk":
         pytest.skip(reason="[be=ramdisk] does not support enumeration")
+    if be_opts["be"] == "driverkit":
+        pytest.skip(reason="[be=ramdisk] does not support repeatedly enumerating")
```
karlowich: The skip reason here needs to be corrected `[be=driverkit]` instead of `[be=ramdisk]`

cijoe/tests/logic/test_xnvme_tests_enum.py
```diff
@@ -23,6 +25,8 @@ def test_multi(cijoe, device, be_opts, cli_args):
         pytest.skip(reason="[be=spdk] does not support repeatedly enumerating")
     if be_opts["be"] == "ramdisk":
         pytest.skip(reason="[be=ramdisk] does not support enumeration")
+    if be_opts["be"] == "driverkit":
+        pytest.skip(reason="[be=ramdisk] does not support repeatedly enumerating")
```
karlowich: Same here

karlowich (changes requested):

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Which part? The assert or the comment?

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: The `assert` just safe-guards that we immediately stop, if the RPC call happens to fail, although it's unlikely. There's no predictable way to recover anyway.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Maybe, maybe not. It informs the next developer that it's a conscious decision to pass these through unchanged.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Changed to document buffer_remove

cijoe/tests/logic/test_xnvme_tests_enum.py
Baekalfen: Fixed.

cijoe/tests/logic/test_xnvme_tests_enum.py
Baekalfen: Fixed.

lib/xnvme_be_macos_driverkit_mem.c
karlowich: Great

lib/xnvme_be_macos_driverkit_async.c
karlowich: Sure, but `ctx->cmd.common.opcode` has sufficiently many cases that it doesn't make sense to include the ones that don't require special handling.
That said, either way is fine.

Baekalfen: Then I'll prefer to keep it. To me, it serves a purpose.

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,221 @@
...
+			return -errno;
+		}
+		if (cb_func(dev, cb_args)) {
+			xnvme_dev_close(dev);
+		}
+	}
+
+	IOObjectRelease(iterator);
+
+	return -ENOSYS;
```
safl: Is the function missing a return 0; at the end? Without it, it seems it can’t signal success as intended?

lib/xnvme_be_macos_driverkit_mem.c
safl: xNVMe does not use ``assert`` to terminate execution, nor should it; handling such cases is left to the  library user.

While asserts can be helpful for debugging, meson is currently not set up to define ``NDEBUG`` to disable assertions in release builds.

Therefore, please replace instances of ``assert`` with ``XNVME_DEBUG``

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,97 @@
...
+		(struct xnvme_be_macos_driverkit_state *)dev->be.state;
+	size_t _output_cnt = 0;
+	struct buffer *_buf;
+	kern_return_t ret;
+
+	if (!buf) {
+		return;
+	}
+	_buf = buffer_find(state->buffers, (uint64_t)buf);
+	assert(_buf);
```
safl: Asserts should not be used here, replace these with ``XNVME_DEBUG``.

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,97 @@
...
+
+	XNVME_DEBUG("buf token: %llx", (uint64_t)_buf->vaddr);
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_DEALLOC_BUFFER,
+					&_buf->vaddr, sizeof(_buf->vaddr), NULL, &_output_cnt);
+	assert(ret == kIOReturnSuccess);
```
safl: Asserts should not be used here, replace these with ``XNVME_DEBUG``.

safl (changes requested):
Left a comment inline about a potential logic-error and then a general request on changing the use of ``assert``.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Yeah, that's a mistake. Weird no one checks for it. Fixed.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Fixed.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Fixed.

Baekalfen:
@safl I can't reply to your comment for some reason?

> xNVMe does not use assert to terminate execution, nor should it; handling such cases is left to the library user.

I've removed all asserts from my code.

Although I understand the sentiment, the user has no way of handling these errors, as it's never supposed to happen, but if it does, you're in an inconsistent error state.

Also, you are in fact using `assert` in other places:

```
lib/xnvme_be_cbi_async_posix.c:82:	assert(req != NULL);
lib/xnvme_be_cbi_async_posix.c:149:	assert(req != NULL);
lib/xnvme_be_fbsd_async.c:178:	assert(req != NULL);
lib/xnvme_be_fbsd_async.c:257:	assert(req != NULL);
lib/xnvme_be_windows_async_iocp.c:153:	assert(req != NULL);
lib/xnvme_be_windows_async_iocp_th.c:177:	assert(req != NULL);
lib/xnvme_be_windows_async_iocp_th.c:244:	assert(req != NULL);
```

safl:
> @safl I can't reply to your comment for some reason?
> 
> > xNVMe does not use assert to terminate execution, nor should it; handling such cases is left to the library user.
> 
> I've removed all asserts from my code.
> 
> Although I understand the sentiment, the user has no way of handling these errors, as it's never supposed to happen, but if it does, you're in an inconsistent error state.

If you call ``assert`` then the user has very no / very poor means of handling the error. If an error is returned, the user can do what they need to do, to handle the error.

> Also, you are in fact using `assert` in other places:
> 
> ```
> lib/xnvme_be_cbi_async_posix.c:82:	assert(req != NULL);
> lib/xnvme_be_cbi_async_posix.c:149:	assert(req != NULL);
> lib/xnvme_be_fbsd_async.c:178:	assert(req != NULL);
> lib/xnvme_be_fbsd_async.c:257:	assert(req != NULL);
> lib/xnvme_be_windows_async_iocp.c:153:	assert(req != NULL);
> lib/xnvme_be_windows_async_iocp_th.c:177:	assert(req != NULL);
> lib/xnvme_be_windows_async_iocp_th.c:244:	assert(req != NULL);
> ```

The codebase isn't perfect, and as you can see, these instances are used for debugging and should be replaced with ``XNVME_DEBUG``. Regardless, the point remains: xNVMe does not crash on error but leaves error handling to the library user.

safl:
@Baekalfen I found a couple of additional things, instead of commenting on the code, then I have pushed two fixes. It is about a convention in the xNVMe code of:

* All returns on error paths have ``XNVME_DEBUG("FAILED: ...")``

* Informational usually has ``XNVME_DEBUG("INFO: ..")``

* When using ``XNVME_DEBUG`` then threre is no need for a trailing ``\n``

I have not been able to test the changes that I have made since I do not have a mac readily availabe, so I will do a couple of iterations on them here.

There are a couple of others things which I will provide as comments.

lib/xnvme_be_macos_driverkit_admin.c
```diff
@@ -0,0 +1,70 @@
...
+	cmd.backend_opaque = (uint64_t)ctx;
+
+	output_cnt = sizeof(NvmeSubmitCmd);
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_ONESHOT, &cmd,
+					sizeof(NvmeSubmitCmd), &cmd_return, &output_cnt);
+	if (ret != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallStructMethod(NVME_ONESHOT); "
+			    "0x%08x, '%s'",
+			    ret, mach_error_string(ret));
+		return ret;
```
safl: The return value here is ``On success, 0 is returned. On error, negative errno is returned.``. 

Does ``ret`` map 1:1 to ``-errno``? If not, then since there is no ``errno`` to propagate, then instead use a fitting constant from errno.h.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+int
+xnvme_be_macos_driverkit_queue_term(struct xnvme_queue *_queue)
+{
+	struct xnvme_queue_macos_driverkit *q = (struct xnvme_queue_macos_driverkit *)_queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)q->base.dev->be.state;
+
+	_xnvme_be_driverkit_delete_ioqpair(state, q->id);
```
safl: The error from ``_xnvme_be_driverkit_delete_ioqpair()`` should be propagated here.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,212 @@
...
+	int queue_ret;
+	NvmeSubmitCmd nvme_cmd;
+	kern_return_t ret;
+	struct xnvme_cmd_ctx *ctx;
+
+	ret = IOConnectCallScalarMethod(state->device_connection, NVME_POKE, qid, 2, NULL, 0);
+	if (ret != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallScalarMethod(); ret(0x%x), msg(%s)", ret,
+			    mach_error_string(ret));
+		return -1;
```
safl: The return value should abide by: ``On success, number of completions processed, may be 0. On error, negative errno is returned.`` Since there is no ``errno`` to propagate, then instead use a fitting constant from ``errno.h``. Using ``return -1`` is not appropriate since that is the same as ``return -EPERM;``, which does not seem applicable here.

Do you know of a generally available mapping of ``kern_return_t`` to ``errno``?

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,230 @@
...
+		XNVME_DEBUG("FAILED: IOServiceGetMatchingService()");
+		return -ENODEV;
+	}
+
+	err = IOServiceOpen(service, mach_task_self_, MACVFN_CLIENT_VERSION, &connection);
+	if (err != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOServiceOpen(MACVFN_CLIENT_VERSION); connection(%u),"
+			    "ret(0x%08x), '%s'",
+			    connection, err, mach_error_string(err));
+		return 1;
```
safl: The return value should abide by: ``On success, 0 is returned. On error, negative errno is returned.`` Since there is no ``errno`` to propagate, then instead use a fitting constant from ``errno.h``.

Do you know of a generally available mapping of ``kern_return_t`` to ``errno``?

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,230 @@
...
+
+	state->device_service = service;
+	state->device_connection = connection;
+
+	err = IOConnectCallScalarMethod(state->device_connection, NVME_INIT, NULL, 0, NULL, 0);
+	if (err != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallScalarMethod(NVME_INIT);"
+			    "ret(0x%08x), '%s'",
+			    err, mach_error_string(err));
+		return 1;
```
safl: The return value must abide by: ``On success, 0 is returned. On error, negative errno is returned.`` Since there is no ``errno`` to propagate, then instead use a fitting constant from ``errno.h``.

Do you know of a generally available mapping of ``kern_return_t`` to ``errno``?

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,115 @@
...
+				   uint64_t *XNVME_UNUSED(phys))
+{
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)dev->be.state;
+	long sz = sysconf(_SC_PAGESIZE);
+	void *ptr;
+	uint64_t *buf;
+
+	if (sz == -1) {
+		XNVME_DEBUG("FAILED: sysconf(), errno: %d", errno);
```
safl, line 28: The return must abide by: ``On success, a pointer to the allocated memory is returned. On error, NULL is returned and errno set to indicate the error.``

Thus, set ``errno`` to a fitting constant from ``errno.h``.

lib/xnvme_be_macos_driverkit_mem.c
```diff
@@ -0,0 +1,115 @@
...
+
+	XNVME_DEBUG("INFO: buf token: %llx", (uint64_t)_buf->vaddr);
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_DEALLOC_BUFFER,
+					&_buf->vaddr, sizeof(_buf->vaddr), NULL, &_output_cnt);
+	if (ret != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallStructMethod(NVME_DEALLOC_BUFFER); "
+			    "ret(0x%08x), '%s'",
+			    ret, mach_error_string(ret));
+
+		errno = EIO;
```
safl: No need to set ``errno`` here as there is no error-handling for the user on ``xnvme_buf_free()``.

lib/xnvme_be_macos_driverkit_sync.c
```diff
@@ -0,0 +1,98 @@
...
+	}
+	memcpy(cmd.cmd, &ctx->cmd, sizeof(struct xnvme_spec_cmd));
+
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_ONESHOT, &cmd,
+					sizeof(NvmeSubmitCmd), &cmd_return, &output_cnt);
+	if (ret != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallStructMethod(NVME_ONESHOT); "
+			    "ret(0x%08x), '%s'",
+			    ret, mach_error_string(ret));
+		return ret;
```
safl: The return value should abide by: ``On success, 0 is returned. On error, negative errno is returned.`` Since there is no ``errno`` to propagate, then instead use a fitting constant from ``errno.h``.

Do you know of a generally available mapping of ``kern_return_t`` to ``errno``?

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,230 @@
...
+		.id = (uint64_t)qid,
+	};
+	kern_return_t ret;
+
+	ret = IOConnectCallStructMethod(state->device_connection, NVME_DELETE_QUEUE_PAIR,
+					&ioqpair_delete, sizeof(NvmeQueue), NULL, &output_cnt);
+	if (ret != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallStructMethod(NVME_DELETE_QUEUE_PAIR); "
+			    "ret(0x%08x), '%s'",
+			    ret, mach_error_string(ret));
```
safl, line 151: Is this missing a ``return`` with error?

safl (changes requested):
Added comments / requests inline for the handling of return-values

lib/xnvme_be_macos_driverkit_admin.c
Baekalfen: `kern_return_t` and errno do not map to each other. And there's seemingly little-to-no overlap in the codes they have. `EIO` seems like the most fitting errno, even though it's very vague.

Changed to `-EIO`.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Changed to `-EIO`. There's no general mapping, and seemingly no other good errors to return.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Changed to `-EIO`.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Changed to `-EIO`.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: `sysconf` sets the errno, and the backend returns it. This is also exactly how `cbi_mem_posix` does it.

lib/xnvme_be_macos_driverkit_mem.c
Baekalfen: Removed. And removed the one on line 77 as well.

lib/xnvme_be_macos_driverkit_sync.c
Baekalfen: Changed to `-EIO`.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Added return.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,216 @@
...
+xnvme_be_macos_driverkit_queue_term(struct xnvme_queue *_queue)
+{
+	struct xnvme_queue_macos_driverkit *q = (struct xnvme_queue_macos_driverkit *)_queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)q->base.dev->be.state;
+	int ret;
+
+	ret = _xnvme_be_driverkit_delete_ioqpair(state, q->id);
+	if (ret) {
+		return ret;
```
safl, line 83: Add error-path debug-message: ``XNVME_DEBUG("FAILED: _xnvme_be_driverkit_delete_ioqpair();``

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,216 @@
...
+	while ((queue_ret = queue_dequeue(q->cq_queue, &nvme_cmd)) == 0) {
+		ctx = (struct xnvme_cmd_ctx *)nvme_cmd.backend_opaque;
+		memcpy(&ctx->cpl, nvme_cmd.cpl, sizeof(struct xnvme_spec_cpl));
+		ctx->async.cb(ctx, ctx->async.cb_arg);
+
+		reaped += 1;
+		queue->base.outstanding -= 1;
+	}
+
+	if (queue_ret == QUEUE_ERROR) {
```
safl, line 119: Add error-path debug-message: ``XNVME_DEBUG("FAILED: queue_ret == QUEUE_ERROR");``

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,216 @@
...
+		_buf_base_offset = ((uint64_t)dbuf) - buf->vaddr;
+		cmd.dbuf_nbytes = dbuf_nbytes;
+		cmd.dbuf_offset = _buf_base_offset;
+		cmd.dbuf_token = buf->vaddr;
+		XNVME_DEBUG("INFO: buf token: %llx: %llx, %llx", (uint64_t)buf->vaddr,
+			    (uint64_t)dbuf, cmd.dbuf_offset);
+	}
+	if (mbuf) {
+		buf = buffer_find(state->buffers, (uint64_t)mbuf);
+		if (!buf) {
```
safl, line 175: add error-path debug-message: ``XNVME_DEBUG("FAILED: buffer_find()");``

safl (changes requested):
A couple of minor things.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,219 @@
...
+{
+	struct xnvme_queue_macos_driverkit *q = (struct xnvme_queue_macos_driverkit *)_queue;
+	struct xnvme_be_macos_driverkit_state *state =
+		(struct xnvme_be_macos_driverkit_state *)q->base.dev->be.state;
+	int ret;
+
+	ret = _xnvme_be_driverkit_delete_ioqpair(state, q->id);
+	if (ret) {
+		XNVME_DEBUG("FAILED: _xnvme_be_driverkit_delete_ioqpair()");
+		return ret;
```
safl, line 83: Should be ``return -EIO;``, since the value in ``ret`` is not applicable for ``errno``,  right?

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,219 @@
...
+		cmd.mbuf_offset = _buf_base_offset;
+		cmd.mbuf_token = buf->vaddr;
+		XNVME_DEBUG("INFO: buf token: %llx: %llx, %llx", (uint64_t)buf->vaddr,
+			    (uint64_t)mbuf, cmd.mbuf_offset);
+	}
+	memcpy(cmd.cmd, &ctx->cmd, sizeof(struct xnvme_spec_cmd));
+	cmd.backend_opaque = (uint64_t)ctx;
+
+	queue_ret = queue_enqueue(q->sq_queue, &cmd);
+	if (queue_ret == QUEUE_FULL) {
```
safl, line 190: Should have ``XNVME_DEBUG("FAILED: ...")`` on the error-path here.

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,219 @@
...
+		return -EBUSY;
+	} else if (queue_ret) {
```
safl, line 192: Should have ``XNVME_DEBUG("FAILED: ...")`` on the error-path here.

lib/xnvme_be_macos_driverkit_dev.c
```diff
@@ -0,0 +1,231 @@
...
+	state->device_connection = connection;
+
+	err = IOConnectCallScalarMethod(state->device_connection, NVME_INIT, NULL, 0, NULL, 0);
+	if (err != kIOReturnSuccess) {
+		XNVME_DEBUG("FAILED: IOConnectCallScalarMethod(NVME_INIT);"
+			    "ret(0x%08x), '%s'",
+			    err, mach_error_string(err));
+		return -EIO;
+	}
+	XNVME_DEBUG("NVME_INIT done\n");
```
safl: Should be: ``XNVME_DEBUG("INFO: NVME_INIT done\n");``

lib/xnvme_be_macos_driverkit_async.c
```diff
@@ -0,0 +1,219 @@
...
+
+	case XNVME_SPEC_FS_OPC_WRITE:
+		ctx->cmd.nvm.slba = ctx->cmd.nvm.slba >> ctx->dev->geo.ssw;
+		ctx->cmd.common.opcode = XNVME_SPEC_NVM_OPC_WRITE;
+		break;
+	}
+
+	if (dbuf) {
+		buf = buffer_find(state->buffers, (uint64_t)dbuf);
+		if (!buf) {
```
safl, line 162: Should have ``XNVME_DEBUG("FAILED: ...")`` on the error-path here.

safl (changes requested):
Another couple of minor things.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: No, because `_xnvme_be_driverkit_delete_ioqpair` is returning `-EIO` or `0`.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: On queue full? It's hardly an error path

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_dev.c
Baekalfen: Done, but without the newline

lib/xnvme_be_macos_driverkit_async.c
safl: ok, great.

lib/xnvme_be_macos_driverkit_async.c
Baekalfen: Done.

lib/xnvme_be_macos_driverkit_async.c
safl: Yeah, you are right. It would just lead to an excessive amount of debug-output even in the common case.

safl (approved):
LGTM!

karlowich (approved):
Nice

safl:
Awesome work @Baekalfen , thanks!

PR #444, opened 2024-03-15
Clean up API documentation

karlowich:
This is the first stab at making the xNVMe documentation easier to read and use.
I have rewritten the sections on the C API and the C API examples.

This also requires some changes to the autogeneration and examples all of which is included in the PR.


bonii (approved):

ankit-sam (approved):
Looks good

PR #445, opened 2024-03-18
fix assertion failure with xnvme enum

ankit-sam:
During xnvme enumeration for spdk backend trtype is not set for TCP, RDMA, FC and CUSTOM transport. If assertions are enabled in spdk, this leads to `spdk_nvme_trid_populate_transport()` throw below specified error.

`DBG:xnvme_be_spdk_dev.c:xnvme_be_spdk_enumerate-687: INFO: trtype: TCP, # 2 of 4 DBG:xnvme_be_spdk_dev.c:_xnvme_be_spdk_ident_to_trid-300: FAILED: uri: , trtype: TCP DBG:xnvme_be_spdk_dev.c:xnvme_be_spdk_enumerate-693: SKIP: !_xnvme_be_spdk_ident_to_trid() nvme.c:1035:spdk_nvme_trid_populate_transport: *ERROR*: no available transports xnvme: nvme.c:1036: spdk_nvme_trid_populate_transport: Assertion `0' failed. Aborted (core dumped)`

Set trtype to avoid assertion 0 failed error.

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -288,6 +288,11 @@ _xnvme_be_spdk_ident_to_trid(const struct xnvme_ident *ident, struct xnvme_opts 
 
 	memset(trid, 0, sizeof(*trid));
 
+	// For TCP, RDMA, FC and CUSTOM transport types, prevent assertion 0,
+	// failure in spdk_nvme_trid_populate_transport().
+	// TODO: See if there is a better way to fix this.
+	trid->trtype = trtype;
+
 	switch (trtype) {
 	case SPDK_NVME_TRANSPORT_PCIE:
 		sprintf(trid_str, "trtype:%s traddr:%s", spdk_nvme_transport_id_trtype_str(trtype),


ankit-sam:
@karlowich I am able to reproduce this issue. What we need to do is just allocate some hugepages (without unbinding nvme devices from kernel and binding to uio/vfio). This way we have only hugepages (requirement for DPDK initialization). Now we start iterating through all the SPDK transport and for TCP, RDMA, FC and CUSTOM transport the trtype is not set. This results in assertion 0 failure in `spdk_nvme_trid_populate_transport()`

lib/xnvme_be_spdk_dev.c
```diff
@@ -299,6 +299,9 @@ _xnvme_be_spdk_ident_to_trid(const struct xnvme_ident *ident, struct xnvme_opts
 		if (sscanf(ident->uri, "%[^:]:%d", addr, &port) != 2) {
 			XNVME_DEBUG("FAILED: uri: %s, trtype: %s", ident->uri,
 				    spdk_nvme_transport_id_trtype_str(trtype));
+			// Prevent assertion 0, in spdk_nvme_trid_populate_transport()
+			// TODO: find a better way to fix this
+			trid->trtype = trtype;
```
karlowich, line 304: This looks fine to me, but I'm wondering if it makes sense to move this up, so we always set `trid->trtype`. What do you think?

ankit-sam: Correct, it makes more sense.

karlowich (approved):
Great!

PR #453, opened 2024-04-08
Advance perf report

karlowich:
This PR fixes some minor issues, but most importantly lays the ground work for populating the test setup section.

### DIFF
--- a/cijoe/scripts/bench_reporter.py
+++ b/cijoe/scripts/bench_reporter.py
@@ -53,7 +53,11 @@ def main(args, cijoe, step):
     # Read the cover-template, populate it, then store it in the artifacts directory
     with (templates_path / "cover.jinja2.tmpl").open() as template:
         with cover_path.open("w") as cover:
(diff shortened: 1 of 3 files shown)


cijoe/scripts/bench_reporter.py
```diff
@@ -14,8 +14,156 @@
...
+
+    for state_path in sysinfo_path.glob("*.state"):
+        with state_path.open("r") as state:
+            for line in (line.strip() for line in state):
+                output_path = sysinfo_path / state_path.name.replace(
+                    ".state", ".output"
+                )
+                right, left = line.strip().split(":")
+                if "cmd" in right:
+                    if "hostname" in left:
```
mbrsamsung, line 65: It could be an idea to define your strings as constant, as an example HOSTNAME = "hostname"

karlowich: I will do this for the ones that are repeated

PR #454, opened 2024-04-09
fix(docs): update compatibility table

NaddiNadja:


docs/compatibility/compatibility_table.rst
```diff
@@ -21,10 +21,10 @@ xNVMe compatibility
...
     * - Write
-      - ✗
-    * - Wrote Uncorrectable
+      - ✓
+    * - Write Uncorrectable
       - ✗
     * - Write Zeroes
-      - ✗
+      - ✓
     * - Read
```
karlowich, line 29: I think we should move read up, so the list is in alphabetical order.

karlowich (changes requested):
One nitpick, otherwise good

karlowich (approved):

PR #455, opened 2024-04-09
feat(meson): added unit tests

NaddiNadja:
The unit tests can be run by compiling with the meson option 'tests' set to true and then running `meson test` in the build directory.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -3,32 +3,75 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 tests_prefix = 'xnvme_tests_'
-tests_source = [
-  'async_intf.c',
-  'buf.c',
-  'cli.c',
-  'compare.c',
-  'copy.c',
-  'delay_identification.c',
-  'enum.c',
-  'ioworker.c',
-  'kvs.c',
-  'lblk.c',
-  'scc.c',
-  'xnvme_cli.c',
-  'xnvme_file.c',
-  'znd_append.c',
-  'znd_explicit_open.c',
-  'znd_state.c',
-  'znd_zrwa.c',
-]
+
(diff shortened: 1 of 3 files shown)


tests/meson.build
```diff
@@ -3,37 +3,82 @@
...
+  ],
+  'compare.c': [
+    ['no args', ['compare', '1GB']],
+    ['slba=0 nlb=7', ['compare', '1GB', '--slba', '0', '--nlb', '7']],
+  ],
+  'copy.c': [
+    ['no args', ['copy', '1GB']],
+    ['slba=0 sdlba=8', ['copy', '1GB', '--slba', '0', '--nlb', '7', '--sdlba', '8']],
+    ['slba=8 sdlba=0', ['copy', '1GB', '--slba', '8', '--nlb', '7', '--sdlba', '0']],  ],
+  'delay_identification.c': [],
```
karlowich, line 31: We don't need to skip this.
It works with the arguments: `['open',  '1GB']`

karlowich (changes requested):
This is great!
In addition to the comment I left, I think we should add a `make test`.
You can take inspiration from `make install` in the `Makefile`

karlowich:
Additionally, it would be great to have the same functionality for the executables in `xnvme/tools/`

karlowich (approved):
Looks good!

ankit-sam (approved):
Looks good, but just wondering if this should include all the tests (say from xnvme.c we are missing a bunch like, set & get features, dsm, log pages etc).

karlowich:
@ankit-sam We are only running the tests that work with the ramdisk. The idea is when we add more functionality to the ramdisk, we can test more things

PR #456, opened 2024-04-09
fix reclaim unit handle status output format

ankit-sam:
Output reclaim unit handle status descriptors to either the requested limit or the device supported value, whichever is smaller.


karlowich (approved):
Looks good

Issue #272, opened 2023-03-23
Correctly handle the XNVMEC_LFLG type arguments in xnvmec_assing_arg


Joelgranados:
in xnvmec_assing_arg.
With XNVMEC_LFLG type arguments `arg` is 0x0 (because there is no accompanying arg) and `num` is always 0 because the code that sets it is skipped in line 1207.
Every time you are in a case of a XNVMEC_LFLG type arg you should have logic that handles it being in the arguments.

IMO the refactoring should be changing all the XNVMEC_LFLG type args from `args->ARG = arg ? num : 1;` to `args->ARG_NAME = 1;` or whatever should happen when the arg is present in the command line.

karlowich:
Since #457 has been merged, I will close this.

PR #457, opened 2024-04-10
fix(xnvme/cli): handle LFLG type arguments in xnvme_cli_assign_arg

NaddiNadja:
When `opt_type == XNVME_CLI_LFLG`, `arg` will always be 0x0, so checking for this variable was redundant. This fix only removes the check for `arg` for `xnvme_cli_opt` values that are used as the `XNVME_CLI_LOPT` type.

This solves #272.


karlowich (approved):
Looks good

PR #464, opened 2024-04-24
fix(tests/map): write out macro

karlowich:
clang-format-14 (CI) and clang-format-18 (local) seems to only disagree
about this macro. To avoid the error write out the macro.

I also believe that the written out macro is more readable.


ankit-sam (approved):
looks good

Issue #465, opened 2024-04-26
xnvme identify controller erroneously sets nsid


Baekalfen:
When calling `xnvme idfy --cns 1`, the `nsid` gets internally set automatically. For a Samsung 980 Pro drive, this will make the admin command fail if the `nsid` is not zero.

When calling `xnvme idfy-ctrlr`, the `nsid` is unconditionally set to `0`, which makes the admin command succeed.

karlowich:
Since #466 has been merged, I will close this

PR #466, opened 2024-04-26
fix(tools/xnvme/idfy): only add nsid if needed

karlowich:
Including the nsid if it is not needed, eg. cns=1 (idfy-ctrlr), can in some cases cause the command to fail.

Closes: #465


karlowich:
@Baekalfen Can you verify, that this resolves #465 ?

Baekalfen (approved):
It works 🚀

safl (approved):
Nice improvement.

Issue #462, opened 2024-04-18
Support keep alive in spdk tcp backend


LiadOz:
According to [NVMe TCP Transport Specification 2021.06.02 ](https://nvmexpress.org/wp-content/uploads/NVM-Express-TCP-Transport-Specification-2021.06.02-Ratified-1.pdf)
```
The NVMe/TCP Transport requires the use of the Keep Alive feature (refer to section 7.12 in the NVMe
Base Specification). The NVMe/TCP Transport does not impose any limitations on the minimum and
maximum Keep Alive Timeout value. The minimum should be set large enough to account for any transient
fabric interconnect failures between the host and controller.
TCP level keep alive functionality is not prohibited but it is recommended that the TCP level keep alive
timeout is set to a higher value than the NVMe Keep Alive Timeout to avoid conflicting policies.
```
We have encountered the need for keep alive timeout in our application in a case of a sudden shutdown. The shutdown caused the host to abort the commands, but the target has kept them until it was brought back a few minutes later, after that time commands that were sent before the shutdown were completed without the application knowing about them causing data integrity issues. Had a keep alive been active, the target would have aborted those commands.

In `lib/xnvme_be_spdk_dev.c` the keep_alive_timeout_ms option is disabled for transport types TCP and RDMA, and it seems that this had existed since the spdk backend was created.
Allowing the user to specify a keep alive timeout is not enough, here is how that keep alive should be used according to spdk:
```c
// include/spdk/nvme.h
struct spdk_nvme_ctrlr_opts {
    // ...

    /**
     * Keep alive timeout in milliseconds (0 = disabled).
     *
     * The NVMe library will set the Keep Alive Timer feature to this value and automatically
     * send Keep Alive commands as needed.  The library user must call
     * spdk_nvme_ctrlr_process_admin_completions() periodically to ensure Keep Alive commands
     * are sent.
     */

    uint32_t keep_alive_timeout_ms;
    // ...
}
```
So to fully support the feature, either xNVMe needs to create a thread that polls `spdk_nvme_ctrlr_process_admin_completions` or give the user the ability to call that function on their own.

karlowich:
Hi @LiadOz, thanks for putting this on our radar. We will have to spend some time coming up with the optimal solution.
For now, I think you should be able to achieve the desired outcome by sending the Keep Alive command "manually" through `xnvme_cmd_pass_admin()`. Can you try this out?

LiadOz:
[...]

PR #467, opened 2024-04-30
feat(spdk): add keep_alive_timeout to opts

LiadOz:
Relates to #462
This only adds the option to specify keep alive timeout, if this option is used, the user must send the KEEP_ALIVE command on their own.


karlowich (approved):
Looks good

karlowich (commented):
Hi @LiadOz, I changed your PR to point to `next` instead and rebased it on `next`.

karlowich (commented):
@LiadOz after further inspection, it seems that we overwrite the variable in the switch statement below your code (line 383). Perhaps you should move your change inside the switch statement instead?

LiadOz:
@karlowich fixed

karlowich:
@LiadOz Did it work before you fixed it?

LiadOz:
It worked because I have two working environments, which had different versions of this code, the current version here is the one I tested.

karlowich (approved):
Great!

PR #474, opened 2024-05-14
Revisit how we handle the shared library in Meson

karlowich:
This adds control over our exported symbols and leverages meson functionality to control whether a user wants both the static and shared libraries or just one of them.


safl (approved):
This looks **very** good to me.

The "CI" in the commit message could be lower-case but that does not really matter, just for another time.

mbrsamsung (approved):
Looks good to me

PR #477, opened 2024-05-17
Documentation overhaul: add user-selectable versions and current site

safl:
* Add an option to dest.py to select which instance of the website / documentation to add to the site-root and set as "current" in versions.json

  - Previously, this would always be 'main', however, this implies that the website can only be updated with releases of xNVMe, this is not practical. Thus this to provide pushes to e.g. the 'website' branch to update the site-root.
  - Care should be taken to then have such a 'docs' branch be based on top of 'main' such that changes in e.g. api or otherwise does not trickle into the docs.

* Extend the autogen/dest.py script to emit versions.json for the PyData theme to consume

* Adjust conf.py; set html-theme options for the PyData config to consume the generated versions.json

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -800,21 +800,25 @@ jobs:
   #
   # * push of branches and tags
   # * manual 'workflow_dispatch'
+  # * pull-request changes, thus, we can catch whether a PR breaks documentation
   #
-  # When triggered on push to branch, then the documentation is deployed to:
+  # When triggered by 'push' or 'workflow_dispatch', then the documentation is
+  # deployed to:
   #
-  #   githubpages/docs/<branch>
+  #   githubpages/docs/<github.ref>
   #
-  # For example: githubpages/docs/next
+  # A couple of examples:
   #
-  # When triggered on push of tag, then the documentation is deployed to:
+  # * githubpages/docs/next
+  # * githubpages/docs/current
+  # * githubpages/docs/foo
+  # * githubpages/docs/v1.2.3
+  # * githubpages/docs/v1.2.4
   #
-  #   githubpages/docs/<tag>
-  #   githubpages/docs/latest
+  # Thus, upon push/dispatch then the archive/history is updated
+  # The root of the site is always synced with githubpages/docs/current in case
+  # there are no changes to "docs/current", then there is nothing to commit
   #
-  # For example: githubpages/docs/v0.5.0
-  # Note that latest is always populated when pushing as tag. It is assumed
-  # that tags are used exclusively for releases
   docgen:
     needs: [source-archive-with-subprojects, source-format-check, build-python]
     runs-on: [self-hosted, linux, verify]
@@ -910,12 +914,16 @@ jobs:
       run: |
         git config --global --add safe.directory $(pwd)/site
 
-    - name: Add docgen to site
+    - name: Extract the website
       if: ((github.event_name == 'push') || ((github.event_name == 'workflow_dispatch') && (github.event.inputs.job == 'docgen')))
       run: |
         cp cijoe/docgen-${{ matrix.guest.os }}-${{ matrix.guest.ver }}/artifacts/docs.tar.gz .
         tar xzf docs.tar.gz
-        ./docs/autogen/dest.py --docs html --site site --ref "${{ github.ref }}"
+
+    - name: Update site repository
+      if: ((github.event_name == 'push') || ((github.event_name == 'workflow_dispatch') && (github.event.inputs.job == 'docgen')))
+      run: |
+        ./docs/autogen/dest.py --docs html --site site --ref "${{ github.ref }}" --current current
 
     - name: Push site-changes
       if: ((github.event_name == 'push') || ((github.event_name == 'workflow_dispatch') && (github.event.inputs.job == 'docgen')))
--- a/docs/autogen/dest.py
+++ b/docs/autogen/dest.py
@@ -46,12 +46,20 @@
 
     * Removing everything in ``<args.site>`` that is not in KEEPLIST
     * Copy ``<args.docs>`` to ``<args.site>/docs/<args.ref>``
-    * Copy ``<args.site>/docs/main/.`` to ``<args.site>/.``
+    * Copy ``<args.site>/docs/<args.current>/.`` to ``<args.site>/.``
+    * Emit a PyData versions.json docs in ``<args.site>``
+      - Note that the script organizes instances of the documentation that are not
+        exposed in the PyData versions.json
+      - The instances of the documentation that are exposed in the PyData versions.json
+        are named as a version-tag (vMAJOR.MINOR.PATCH), or;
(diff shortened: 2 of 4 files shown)


docs/autogen/dest.py
```diff
@@ -105,6 +114,43 @@ def remove_except(path, keep):
...
 
 
+def gen_versions(path: Path, current: str):
+    """Returns a dict compatible with the Pydata versions-format"""
+
+    def version_key(name):
+        match = re.match(r"v(\d+)\.(\d+)\.(\d+)", name)
+        if match:
+            return (1, -int(match.group(1)), -int(match.group(2)), -int(match.group(3)))
+        return (0, name)
```
karlowich, line 124: The return type of this is weird, either you return a tuple with four numbers or a number and a string.
Additionally, negating the numbers is unclear. I would prefer if the call to `sorted()` had `reverse=True` instead.

docs/autogen/dest.py
```diff
@@ -105,6 +114,43 @@ def remove_except(path, keep):
...
+    ]
+
+    for item in sorted(path.iterdir(), key=lambda x: version_key(x.name)):
+        versions.append(
+            {
+                "name": f"{item.name}",
+                "version": f"{item.name}",
+                "url": f"/docs/{item.name}",
+            }
+        )
```
karlowich, line 142: Instead of having to write `.name` four times you could do a list comprehension inside the sorted call.
Something like: `[x.name for x in path.iterdir()]`.
That would also allow you to get rid of the lambda: `key=version_key`

docs/autogen/dest.py
```diff
@@ -105,6 +114,43 @@ def remove_except(path, keep):
...
+        if match:
+            return (1, -int(match.group(1)), -int(match.group(2)), -int(match.group(3)))
+        return (0, name)
+
+    versions = [
+        {
+            "name": "current",
+            "version": f"{current}",
+            "url": "https://xnvme.org/",
+            "preferred": True,
```
karlowich, line 131: I would prefer if the "current" wasn't repeated. If I read this correctly we will have:
* current
* 0.7.4
* 0.7.3
* etc...

In my opinion it would be nicer to have:
* 0.7.4 (current)
* 0.7.3
* etc...

karlowich (changes requested):
A couple of comments

docs/autogen/dest.py
safl: It is done to define an ordering that is not just alphanumerical. I want to separate branch names and version tags. Branch-names comes before version-tags and version-tags are in descending order.

I do prefer to have this logic inside the key-function, since that is what it is there for, to define ordering. By adding ``reverse=True`` then that would affect the ordering of the branch names as well, which it should not.

docs/autogen/dest.py
safl: Would this be preferable:
```python
    for vname in (
        x.name for x in sorted(path.iterdir(), key=lambda x: version_key(x.name))
    ):
        versions.append(
            {
                "name": f"{vname}",
                "version": f"{vname}",
                "url": f"/docs/{vname}",
            }
        )
```

I don't see how skipping ``key=version_key`` is possible without losing the ordering?

safl: Or something like:

```python
    for dname in sorted(
        (d.name for d in path.iterdir()), key=lambda x: version_key(x.name)
    ):
        versions.append(
            {
                "name": f"{dname}",
                "version": f"{dname}",
                "url": f"/docs/{dname}",
            }
        )
```

karlowich: This is what I mean:
```python
for name in sorted((d.name for d in path.iterdir()), key=version_key):
        versions.append(
            {
                "name": f"{name}",
                "version": f"{name}",
                "url": f"/docs/{name}",
            }
        )
```

safl: Nice, adjusted it to this.

docs/autogen/dest.py
safl: I have added a doc-string describing how the function sorts, which might help with the non-obviousness of the code.

docs/autogen/dest.py
safl: Currently, the items in the list are simply branch and tag-names based on the convention that, that is what they were named when they were placed in the ``docs/`` folder.

To do something like you mention here then this needs to be done in a different manner, where for the branch-names, then the script would need to parse version-information from the HTML in ``docs/<branch_name>`` for all generated docs in such folders.

How about we just keep it simple as:

```
branches
tags
```

?

.github/workflows/verify.yml
```diff
@@ -793,21 +793,25 @@ jobs:
...
+  # * githubpages/docs/current
+  # * githubpages/docs/foo
+  # * githubpages/docs/v1.2.3
+  # * githubpages/docs/v1.2.4
   #
-  #   githubpages/docs/<tag>
-  #   githubpages/docs/latest
+  # Thus, upon push/dispatch then the archive/history is updated
+  # The root of the site is always synced with githubpages/docs/current in case
+  # there are no changes to "docs/current" well, then there is nothing to commit
```
karlowich, line 813:
```suggestion
  # there are no changes to "docs/current", then there is nothing to commit
```

docs/autogen/dest.py
```diff
@@ -46,12 +46,20 @@
 
     * Removing everything in ``<args.site>`` that is not in KEEPLIST
     * Copy ``<args.docs>`` to ``<args.site>/docs/<args.ref>``
-    * Copy ``<args.site>/docs/main/.`` to ``<args.site>/.``
+    * Copy ``<args.site>/docs/<args.current>/.`` to ``<args.site>/.``
+    * Emit a PyData versions.json docs in ``<args.site>``
+      - Do note that the script organizes instances of the documentation that are not
```
karlowich, line 51:
```suggestion
      - Note that the script organizes instances of the documentation that are not
```

docs/autogen/dest.py
```diff
@@ -46,12 +46,20 @@
...
+        exposed in the PyData versions.json
+      - The instances of the documentation that are exposed in the PyData versions.json
+        contains are named as a version-tag (vMAJOR.MINOR.PATCH), or the names;
```
karlowich, line 54:
```suggestion
        are named as a version-tag (vMAJOR.MINOR.PATCH), or;
```

docs/contributing/contributing-branches.rst
```diff
@@ -14,20 +14,24 @@ Branches
 ``next`` Staging / integration
    This is where GitHUB pull-requests should point to.
 
+``current`` Website updates
+   This is based of ``main`` and only contains changes to the website /
```
karlowich, line 18:
```suggestion
   This is based off ``main`` and only contains changes to the website /
```

karlowich (changes requested):
Nitpicks

safl:
> Nitpicks

Thanks! Added your suggestions and squashed them in.

karlowich (approved):
Nice

PR #482, opened 2024-05-24
documentation overhaul: version selection with visual indicators

safl:
* Drop the ``--current`` argument to dest.py, the current / preferred version is now always the latest version
* The version of xNVMe is added to the theme options ``html_theme_options.switcher.version_match`` to enable correct behavior or the multi-version switcher
* Added visual indicators when looking at an old version and inform when looking at the development version via the ``html_theme_options.show_version_warning_banner``
* Update the documentation describing what the branch named **docs** is for, this was previously the task of **current**
  * Since the purpose of the branch is to host changes to the website for the most recent release, then i find ``docs`` to be a better suited name

**NOTE**: This is currently failing due to unrelated failure of HW for benchmarking

### DIFF
--- a/docs/autogen/dest.py
+++ b/docs/autogen/dest.py
@@ -14,8 +14,9 @@
 
     The intent of the organization is two-fold:
 
-    * Provide a preview of website/documentation of the current state of the 'next'
-      branch, as well as other in-development branches
+    * Provide a preview of website/documentation of the 'next' branch / release, other
+      in-development branches are not available via version-switcher, however,
+      they can be pointed to manually e.g. https://xnvme.io/en/feature-a
 
     * Keep a bit of history on the project documentation available online, e.g. for
       those not on the latest release.
@@ -26,15 +27,16 @@
         refs/tags/<tag_name>
 
     The organization is as follows, for all refs, then the folder ``<args.docs>`` is
-    renamed to ``<args.ref>`` and is placed in ``<args.site>/docs/<args.ref>``, here are
+    renamed to ``<args.ref>`` and is placed in ``<args.site>/en/<args.ref>``, here are
     a couple of examples::
 
-        <args.site>/docs/main
-        <args.site>/docs/next
-        <args.site>/docs/v1.2.3
+        <args.site>/en/main
+        <args.site>/en/docs
+        <args.site>/en/next
+        <args.site>/en/v1.2.3
 
-    This is the archive/history part of the organization. The latest version of the
-    website and documentation lives in the root of ``<args.site>``:
+    Handling of "docs" and version-tags (vX.Y.Z) is special, these are copied to
+    the ``<args.site>``:
 
         <args.site>/.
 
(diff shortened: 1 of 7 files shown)


docs/contributing/release-checklist.rst
```diff
@@ -47,7 +47,7 @@ release are integrated on ``next`` and all tests are passing. Then:
   - Setup a PR of ``next`` onto ``main``
   - Review the PR
   - Wait for tests to finalize / pass
-  - Double-check the generated docs at https://xnvme.io/docs/next
+  - Double-check the generated docs at https://xnvme.io/en/next
```
karlowich, line 50: This confuses me, are we looking to have multiple languages?

docs/autogen/dest.py
```diff
@@ -29,12 +30,13 @@
     renamed to ``<args.ref>`` and is placed in ``<args.site>/docs/<args.ref>``, here are
```
karlowich, line 30: This line is not correct if we switch to `/en/`

karlowich (commented):
A couple of comments

docs/contributing/release-checklist.rst
safl: The previous usage of the ``docs`` folder were fitting when the generated sphinx-doc only lived inside of ``docs``. After changing the sphinx-doc then the folder is for different version of the entire site.Thus, switching to the commonly used convention for multi-version sphinx-docs / read-the-docs.

docs/autogen/dest.py
safl: nice, catch. thanks.

docs/contributing/release-checklist.rst
karlowich: Do we need a prefix? Couldn't we just have `xnvme.io/next`?

safl: The convention is to have the language in there.

safl: Having said that, then it would be possible but require a bit more work in the ``dest.py`` script to manage this and then of course potential conflicts between branch-names and top-level section names. So, for simplicity I would like to just follow the ``/en/<branch,tag>`` convention.

docs/autogen/test_multi_version.sh
```diff
@@ -0,0 +1,24 @@
...
+  cp -r builddir/html $HOME/git/xnvme.github.io/en/$VER
+done
+
+# Copy the build directory to the destination, when testing this locally then
+# change the --url to "localhost"
+./dest.py \
+  --docs builddir/html \
+  --site $HOME/git/xnvme.github.io \
+  --ref refs/heads/docs \
+  --url "https://xnvme.io"
```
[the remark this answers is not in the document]
safl: fixed.

mbrsamsung (approved):
lgtm

karlowich (approved):

PR #483, opened 2024-05-24
documentation overhaul: quick pass cleanup

safl:
Initial removal of stale documentation and refactor / move of sections.

Depends on: https://github.com/xnvme/xnvme/pull/482 for fixing the docgen error


docs/backends/xnvme_info_block.out
```diff
@@ -0,0 +1,43 @@
+
+
+
+
+
+
```
karlowich, line 6: Is this whitespace intended?

docs/backends/xnvme_info_zoned.out
```diff
@@ -0,0 +1,43 @@
+
+
+
+
+
+
```
karlowich, line 6: Same here

karlowich (commented):
A couple of comments

docs/backends/xnvme_info_block.out
safl: It is the raw output from the command in the ``.cmd`` file. If the commands output empty lines, then there will be empty lines here.

docs/backends/xnvme_info_zoned.out
safl: It is the raw output from the command in the ``.cmd`` file. If the commands output empty lines, then there will be empty lines here.

mbrsamsung (approved):
lgtm

PR #484, opened 2024-05-27
feat(be/wpdk): drop support for WPDK (SPDK on Windows)

safl:
A lot of effort has gone into getting WPDK up and running; however,
it has not reached a point where it is possible to automate functional
verification.

Thus, we are dropping support for now. If we get the request, then we
have a good starting point to reintroduce it. However, until then, the
focus on Windows support will pivot back to maintaining support for the
Windows IO interfaces of IOCP, driver IOCTLs, and  Windows IORING.

This is done in preparation for updating the documentation and toolchain
and automating the performance evaluation of xNVMe on Windows.

* docs: the description of WSL and the need for installation of net-uio
  and the other DPDK drivers on Windows are removed as they are no
  longer required for xNVMe on Windows. For xNVMe on Windows we should
  now have reduced dependencies to msys2 / mingw.

* subprojects: the "wpdk" and the special "spdk-win" are removed as they
  are no longer needed

* meson: WDPK specific depedencies are removed as they are no longer
  needed

* github-workflow: the installation of WSL is removed as the dependency
  on WSL is for WPDK/SPDK not for building xNVMe

This does not affect any of the backend code, since the toolchain,
drivers, and the "spdk-win" and "wpdk" subprojects enabled the use of
the xNVMe SPDK backend on Windows.


meson.build
```diff
@@ -142,7 +142,7 @@ conf_data.set('XNVME_BE_MACOS_ENABLED', corefoundation_dep.found() and iokit_dep
 
 conf_data.set('XNVME_BE_FBSD_ENABLED', is_freebsd)
 
-conf_data.set('XNVME_BE_SPDK_ENABLED', (is_linux or is_freebsd or is_windows) and get_option('with-spdk'))
+conf_data.set('XNVME_BE_SPDK_ENABLED', (is_linux or is_freebsd) and get_option('with-spdk'))
 conf_data.set('XNVME_BE_SPDK_TRANSPORT_PCIE_ENABLED', (is_linux or is_freebsd or is_windows) and get_option('with-spdk'))
```
karlowich, line 146: The `is_windows` is also unneeded here

meson.build
```diff
@@ -229,14 +212,6 @@ spdk_dep = dependency(
   '_spdk',
   required: (is_windows ? false : true) and conf_data.get('XNVME_BE_SPDK_ENABLED')
```
karlowich, line 213: This can be cleaned up

docs/getting_started/system.rst
```diff
@@ -437,49 +437,3 @@ Yielding the output:
...
-
-Refer to the DPDK links listed below to resolve runtime dependencies:
-
-* :xref-dpdk-guide-lock-pages:`Grant Lock pages in memory Privilege<>`
-
-* :xref-dpdk-guide-install-kmods:`Download DPDK kmods-driver<>`
-
-* :xref-dpdk-guide-install-virt2phys:`Install virt2phys<>`
-
-* :xref-dpdk-guide-install-netuio:`Install NetUIO<>`
```
karlowich, line 484: These should also be removed from the `conf.py` in `docs/autogen`

karlowich (changes requested):

meson.build
safl: nice! Thanks :)

docs/getting_started/system.rst
safl: Right, thanks.

karlowich (approved):

karlowich:
Can you run `toolbox/pkgs/emitter/emit.py` to generate an up to date script for windows as well?

karlowich:
@safl I found some `#ifndef XNVME_BE_WINDOWS_ENABLED` statements in the SPDK backend, make sure you remove those as well

safl:
> @safl I found some `#ifndef XNVME_BE_WINDOWS_ENABLED` statements in the SPDK backend, make sure you remove those as well

awesome, thanks! Fixed.

mbrsamsung (approved):
LGTM

toolbox/pkgs/windows-2022.ps1
karlowich, line 1: If you make this change, you should change the way the emitter works too.
I imagine changes are required in both `toolbox/pkgs/emitter/deps.yaml` and the relevant templates in `toolbox/pkgs/emitter/templates`

karlowich (changes requested):
One comment, otherwise it is looking good

toolbox/pkgs/windows-2022.ps1
safl: Nice catch, thanks. fixed

mbrsamsung (approved):

PR #485, opened 2024-05-27
CI: Add missing devices to configs 

### LINKED ISSUE #473: Run userspace tests on FreeBSD
Currently our freebsd doesn't define any userspace devices. 
This means that even though SPDK is installed and supported, we don't test it. 
This should be addressed.

karlowich:
Closes #473


safl (dismissed):
Awesome.

PR #487, opened 2024-05-28
Support for latency benchmarking on FreeBSD

NaddiNadja:


cijoe/configs/perf-lat-freebsd.toml
```diff
@@ -0,0 +1,71 @@
...
+port = 22
+username = "root"
+password = "root"
+
+[os]
+name = "freebsd"
+version = "13"
+
+[xnvme.repository]
+remote = "https://github.com/OpenMPDK/xNVMe.git"
```
safl, line 15: Replace the old org. link:
https://github.com/OpenMPDK/xNVMe.git
with
https://github.com/xnvme/xnvme.git

safl (changes requested):
This is looking great, three things:

* A minor thing noted inline
* A general note that the commit-message should be wrapped at around 72 for readability, e.g. if you look here: https://github.com/xnvme/xnvme/pull/487/commits then the short-description is truncated. Also, wrap the long-description, in other words... wrap the entire commit-message at 72 chars.
* Something should probably be done to allow the benchmark jobs be to run in parallel, looks like then are currently waiting for each other to finish? I guess, this might be a simple case of some missing labels on the installed runners.

cijoe/workflows/benchmark-latency.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+steps:
+
+- name: benchmark
+  uses: fio_latency
+  with:
+    repetitions: 1
+    io_sizes: ['4096']
+    io_depths: [1, 2, 4, 8, 16]
+    ndevices: 8
```
safl, line 24: Are these used? As I recall, then only a single device it utilized for the latency-eval.

safl (approved):
This is **excellent**! I have a question, though it's non-blocking.

We have a lot of work to do on the report and scripts. However, since this PR focuses on setting up the plumbing for FreeBSD, let's address improvements to the report, scripts, and workflow in a separate PR.

cijoe/workflows/benchmark-latency.yaml
NaddiNadja: No, it looks like`ndevices` is a leftover from reusing the structure of the benchmark workflow to setup the perf-lat workflow.

safl: ok, thx, let's clean that up in a separate PR.

mbrsamsung (approved):
LGTM

safl (approved):
LGTM

PR #488, opened 2024-05-28
fix(docs/autogen): always populate site-root

safl:
The organization script always removes everything but the items in the KEEPLIST. Thus, the content of the site-root is removed. Afterwards, either the latest version, or 'docs' is added to the site-root.

Due to a logic-error, then the site-root was not re-populated as intended. This fixes it.

Whenever there are pushes to ``next``, after merge of a PR, then the website will currently break.


karlowich (approved):
Lgtm

PR #498, opened 2024-06-12
xNVMe CLI: Clean up sanitize

karlowich:
Additionally, this also cleans up the `zoned.c` which was tangled with sanitize in `xnvme.c`

### DIFF
--- a/tools/xnvme.c
+++ b/tools/xnvme.c
@@ -799,7 +799,7 @@ sub_sanitize(struct xnvme_cli *cli)
 {
 	struct xnvme_dev *dev = cli->args.dev;
 	struct xnvme_cmd_ctx ctx = xnvme_cmd_ctx_from_dev(dev);
-	uint8_t sanact = cli->args.action;
+	uint8_t sanact = cli->args.sanact;
 	uint8_t ause = cli->args.ause;
 	uint32_t ovrpat = cli->args.ovrpat;
 	uint8_t owpass = cli->args.owpass;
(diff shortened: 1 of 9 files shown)


tools/xnvme.c
```diff
@@ -1479,13 +1479,20 @@ static struct xnvme_cli_sub g_subs[] = {
 	},
 	{
 		"sanitize",
-		"Sanitize...",
-		"Sanitize...",
+		"Start a Sanitize Operation",
```
safl, line 1482: Very nice fixes here.

safl (changes requested):
* The commit: ``refactor(zoned): rename action to zsa`` is confusing because you are fixing what seems to be a bug (assignment to ``args->action`` when handling "ZRMS"). And then you refactor ``ZRMS`` to ``ZSA`` in the cli + tool, while also, unrelated to this, refactor ``action`` to ``zsa`` in the API. To me this would be better to split into:

  - ``refactor(cli,tools/zoned): rename 'zrms' to 'zsa' and fix use of cli->action``
  - ``refactor(api/znd): rename 'action' to 'zsa'``

* The commit: ``refactor(sanitize): change flags to be booleans`` touches api, cli, and tools, i would prefer that this is reflected in the short-message, e.g.

  - ``refactor(api/nvm:adm:cli,tools/sanitize): change flags to be booleans``
  - Additionally, then these data-type changes, shouldn't they be done in ``libxnvme_cli.h`` as well? Otherwise, then you are assigning a ``bool`` to ``uint32_t``.

safl (approved):
lgtm

PR #503, opened 2024-06-17
fix(dev): use helper function for geo-access in xnvme_dev_pr()

safl:
Previously, the pretty-printers xnvme_dev_pr() and xnvme_dev_fpr() directly accessed dev->geo instead of using the xnvme_dev_get_geo() helper function. This bypassed the delayed-identification machinery, causing the device to be printed as "UNKNOWN".

This commit corrects the issue by using the helper function for geo-access.

### DIFF
--- a/lib/xnvme_dev.c
+++ b/lib/xnvme_dev.c
@@ -423,6 +423,7 @@ xnvme_dev_derive_geo(struct xnvme_dev *dev)
 int
 xnvme_dev_fpr(FILE *stream, const struct xnvme_dev *dev, int opts)
 {
+	const struct xnvme_geo *geo;
 	int wrtn = 0;
 
 	switch (opts) {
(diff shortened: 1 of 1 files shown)


karlowich:
@safl Do we want to use the helper everywhere? A quick `grep -r "dev->geo"` shows that we access `geo` from `dev` quite often.

safl:
@karlowich  No, not everywhere, since that would lead to circular calls, e.g. during setup of device / derivation of geometry. There might be other places where it should be done, however, I just stumbled over this one.

karlowich:
@safl Interestingly, the failing test says that what you're fixing is the intended behaviour

karlowich:
@safl Thinking about it, I think the behaviour assumed by the test is reasonable. In my opinion it weird that printing the device would have "side-effects"

safl:
Right... I had sort of overlooked this... and yes calling something "trival" like ``xnvme_dev_pr()`` should not have indirect side-effects like this... closing this again.

PR #505, opened 2024-07-01
feat(toolbox): add conda package definition

safl:
This is done to meet a request for xNVMe as a Conda package, see https://github.com/xnvme/xnvme/issues/506 for tracking of additional features to add to the content of the Conda package.

- [x] Switch Conda recipes to off-tree: https://github.com/xnvme/xnvme-conda

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1255,3 +1255,48 @@ jobs:
           debpkg/builddir/*.dsc
           debpkg/builddir/*.buildinfo
           debpkg/builddir/*.changes
+
+  packaging-conda:
+    needs: source-archive
+    if: github.event_name == 'pull_request'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    steps:
+    - name: Set up Miniconda
+      uses: conda-incubator/setup-miniconda@v3
+      with:
+        auto-update-conda: true
+        miniconda-version: "latest"
+    - name: Info
+      run: |
+        echo "Current branch is ${{ github.base_ref }}"
+        echo "Current tag is ${{ github.ref }}"      
+
+    - name: Retrieve the xNVMe source archive
+      uses: actions/download-artifact@v4.1.1
+      with:
+        name: xnvme-src-archive
+
+    - name: Retrieve the xNVMe Conda package repository and prep. src.
+      run: |
+        git clone https://github.com/xnvme/xnvme-conda.git condapkg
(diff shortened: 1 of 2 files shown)


.github/workflows/verify.yml
```diff
@@ -1255,3 +1255,63 @@ jobs:
...
+        cd xnvme/toolbox/conda
+        conda activate xnvme_conda_build
+        conda-build . --output-folder /tmp/conda_pkg
+        find /tmp/conda_pkg
+
+    - name: Find the built conda package
+      id: find-package
+      run: |
+        PACKAGE_PATH=$(find /tmp/conda_pkg -name "*.tar.bz2" | head -n 1)
+        echo "PACKAGE_PATH=$PACKAGE_PATH" >> $GITHUB_ENV
```
karlowich, line 1310: Do we need this step?

karlowich (commented):
Just a single comment

.github/workflows/verify.yml
mbrsamsung: What do you mean?

safl: > Do we need this step?

Yes, to produce a path for "upload-artifact", making the built-artifact available.

karlowich:
>  I'll move the Conda-recipes into a repository of its own
I agree with this approach

.github/workflows/verify.yml
safl: > Maybe I'm not reading the code correctly - but it seems to set an environment variable that we never use

hah, you are right, I had forgotten that I went with the entire build folder to get the auxiliary artifacts as well. So, nice catch!

docs/contributing/packaging/index.rst
```diff
@@ -19,6 +19,10 @@ see the :xref-github-xnvme-issues:`GitHUB Issue Tracker<389>`.
 
   - See: https://formulae.brew.sh/formula/xnvme
 
+* Conda
+
+  - See: h
```
karlowich, line 24: Something is missing here

safl:
Switching this back to DRAFT as the conda recipe itself is now in being staged for conda-forge:
https://github.com/conda-forge/staged-recipes/pull/26854

Thus, this PR here will probably change into just being maintenance documentation.

safl:
Closing this, since it might be a while and by then I expect the Documentation overhaul will be finalized and thus simpler to update the documentation then. Status / progress on the Conda packaging is tracked here:

- https://github.com/xnvme/xnvme/issues/389
- https://github.com/xnvme/xnvme/issues/506

PR #511, opened 2024-07-11
feat(meson): change spdk to be optional

### LINKED ISSUE #508: meson: change ``with-spdk`` from boolean to feature with default value ``auto``
Currently, when grabbing the **xNVMe** source and building it, all dependencies except **SPDK** are handled gracefully by disabling **features** whose dependencies are not met. However, for **SPDK**, the build fails and leaves the build directory in a broken state.

This issue typically occurs when building **xNVMe** on a system where the dependencies for building **SPDK** are not met. In such cases, the user must explicitly disable the **SPDK** backend.

A more sensible behavior would be to change the meson-option ``with-spdk`` from type ``boolean`` to ``feature`` and set its default value to ``auto``. This change should adjust the behavior to skip building the **SPDK** subproject when its dependencies are not satisfied. While I might be overlooking something, this adjustment should lead to a more reliable build process.

There might be more to it, but definitely worth exploring.

karlowich:
Closes #508

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -262,7 +262,7 @@ jobs:
       matrix:
         container:
         # Using an alpine container which is the same as upstream but with bash
-        # alpine: SPDK not supported, --with-spdk=false
+        # alpine: SPDK not supported, --with-spdk=disabled
         - {os: 'alpine', dh: 'refenv/alpine-bash', ver: 'latest'}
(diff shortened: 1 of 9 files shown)


NaddiNadja (approved):
lgtm

NaddiNadja (approved):
lgtm

safl (approved):
Excellent!

PR #512, opened 2024-07-15
fix(cijoe/qemu): prolong timeout for guest.is_up()

NaddiNadja:
verify freebsd script keeps failing without this


karlowich:
This is included here: #513

PR #513, opened 2024-07-15
Fix FreeBSD build errors

karlowich:
Currently, the FreeBSD build fails because it tries to install pipx using the wrong pkg-name. This fixes the issue


NaddiNadja (approved):
great!

safl (approved):
Great!

PR #522, opened 2024-08-14
Run latency benchmarking on Windows

NaddiNadja:
Run latency benchmarking on Windows, comparing the three backends `ioring`, `iocp` and `iocp_th` from the xNVMe engine to the fio engine `windowsaio`.

[latency-windows.pdf](https://github.com/user-attachments/files/16611980/latency-windows.pdf)

### DIFF
--- a/cijoe/configs/perf-lat-windows.toml
+++ b/cijoe/configs/perf-lat-windows.toml
@@ -1,0 +1,52 @@
+[cijoe.transport.ssh]
+hostname = "perf-lat-win.home.arpa"
+port = 22
+username = "root"
+password = "root"
+
+[os]
+name = "windows"
+prettyname = "Windows"
+version = "11"
+
+[cijoe.run]
+shell = "bash"
+
+[xnvme.repository]
+remote = "https://github.com/xnvme/xnvme.git"
+path = "C:/Users/root/git/xnvme"
+branch = "next"
+
+[xnvme.build]
+type = "debug"
+
+[fio]
+bin = "C:/Users/root/git/fio/fio.exe"
+
+[fio.repository]
+remote = "https://github.com/axboe/fio.git"
+path = "C:/Users/root/git/fio"
+tag = "fio-3.36"
+
+[fio.engines]
+
+null = {type = "builtin", group = "null", device = "bdev"}
+windowsaio = {type = "builtin", group = "windowsaio", device = "bdev"}
+
+xnvme_null = {type = "builtin", group = "null", be = "windows", async = "nil", device = "bdev"}
+xnvme_ioring = {type = "builtin", group = "windowsaio", be = "windows", async = "io_ring", device = "bdev"}
+xnvme_iocp = {type = "builtin", group = "windowsaio", be = "windows", async = "iocp", device = "bdev"}
+xnvme_iocpth = {type = "builtin", group = "windowsaio", be = "windows", async = "iocp_th", device = "bdev"}
+  
+[[devices]] 
+key = "bdev"
+uri = "\\\\.\\PhysicalDrive1" # backslashes need to be escaped
+nsid = 1
+labels = ["dev", "bdev", "nvm"]
+driver_attachment = "kernel"
+
+[[duts]]
+pcie = "0000:03:00.0"
+os = "0n1"
+sn = "PHBT718003QR016D"
+label = "INTELM1016G"
--- a/cijoe/scripts/fio_latency.py
+++ b/cijoe/scripts/fio_latency.py
@@ -61,6 +61,10 @@ class Engines(Enum):
     NIL = "nil"
     SPDK_NVME = "spdk_nvme"
     SPDK_BDEV = "spdk_bdev"
+    WINDOWSAIO = "windowsaio"
+    XNVME_IORING = "xnvme_ioring"
+    XNVME_IOCP = "xnvme_iocp"
+    XNVME_IOCPTH = "xnvme_iocpth"
     XNVME_IO_URING = "xnvme_io_uring"
     XNVME_IO_URING_CMD = "xnvme_io_uring_cmd"
     XNVME_KQUEUE = "xnvme_kqueue"
@@ -138,7 +142,7 @@ class XnvmeKernelEngine(KernelEngine):
         return f"{self.name}_{self.async_}"
 
     def extra_args(self) -> Dict[str, str]:
-        return {"-xnvme_be": f"{self.be}", "-xnvme_async": f"{self.async_}"}
+        return {"--xnvme_be": f"{self.be}", "--xnvme_async": f"{self.async_}"}
 
 
 @dataclass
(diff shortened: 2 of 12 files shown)


cijoe/configs/perf-lat-windows.toml
```diff
@@ -0,0 +1,53 @@
...
+remote = "https://github.com/axboe/fio.git"
+path = "C:/Users/root/git/fio"
+tag = "fio-3.36"
+
+[fio.engines]
+
+null = {type = "builtin", group = "null", device = "bdev"}
+windowsaio = {type = "builtin", group = "windowsaio", device = "bdev"}
+
+# xnvme_nil doesn't exist on windows yet
```
karlowich, line 36: This comment seems to be outdated

cijoe/configs/perf-lat-windows.toml
```diff
@@ -0,0 +1,53 @@
...
+xnvme_null = {type = "builtin", group = "null", be = "windows", async = "nil", device = "bdev"} # Kernel
+xnvme_ioring = {type = "builtin", group = "windowsaio", be = "windows", async = "io_ring", device = "bdev"} # Kernel
+xnvme_iocp = {type = "builtin", group = "windowsaio", be = "windows", async = "iocp", device = "bdev"} # Kernel
+xnvme_iocpth = {type = "builtin", group = "windowsaio", be = "windows", async = "iocp_th", device = "bdev"} # Kernel
```
karlowich, line 40: We only have kernel interfaces on Windows, no need for the comments

cijoe/scripts/fio_latency.py
```diff
@@ -116,12 +120,20 @@ def prepare(self) -> None:
 
 class KernelEngine(Engine):
     def prepare(self) -> None:
-        self.cijoe.run(" ".join(["xnvme-driver", "reset"]))
+        shell = (
+            self.cijoe.config.options.get("cijoe", {}).get("run", {}).get("shell", "sh")
+        )
+        if shell != "pwsh":
+            self.cijoe.run(" ".join(["xnvme-driver", "reset"]))
```
karlowich, line 127: This brings up an interesting point.
`xnvme-driver` is really only relevant on Linux and FreeBSD. Maybe we shouldn't even install it on macOS and Windows.
If we eventually get userspace drivers on either macOS or Windows, then `xnvme-driver` should probably be expanded to support attaching/detaching the devices on those as well.

cijoe/scripts/fio_prep.py
```diff
@@ -19,26 +19,37 @@ def main(args, cijoe, step):
...
-    prefix = cijoe.config.options.get("fio", {}).get("build", {}).get("prefix", {})
-    if prefix:
-        commands += [
-            (f"./configure --prefix={prefix}", repos["path"]),
+    shell = cijoe.config.options.get("cijoe", {}).get("run", {}).get("shell", "sh")
+
+    make = "gmake" if shell == "csh" else "make"
+
+    if shell == "pwsh":
+        # 2>$null will ignore the error if repo exists
```
karlowich, line 27: If you always start by removing it, then I guess it would never exist.
What happens if you try to remove it and it doesn't exist?
Probably the `2>$null` should be moved to the `remove-item`.

cijoe/scripts/xnvme_build.py
```diff
@@ -23,14 +23,24 @@ def main(args, cijoe, step):
...
-        "meson setup builddir",
-        "meson compile -C builddir",
-        "cat builddir/meson-logs/meson-log.txt",
-    ]
+    shell = cijoe.config.options.get("cijoe", {}).get("run", {}).get("shell", "sh")
+
+    if shell == "pwsh":
+        commands = [
+            "git rev-parse --short HEAD",
+            "if (test-path builddir) { remove-item builddir -verbose -recurse }",
```
karlowich, line 31: The if statement could also be added to the fio_prep

cijoe/scripts/xnvme_install.py
```diff
@@ -21,16 +21,18 @@ def main(args, cijoe, step):
...
+        ["cd builddir", "meson compile"],
+        ["cd builddir", "meson install"],
+        ["cd builddir", "meson --internal uninstall"],
+        ["cd builddir", "meson install"],
+        ["cat builddir/meson-logs/meson-log.txt"],
     ]
     for cmd in commands:
-        err, _ = cijoe.run(cmd, cwd=xnvme_source)
+        sequence_operator = "; " if shell == "pwsh" else " && "
+        err, _ = cijoe.run(sequence_operator.join(cmd), cwd=xnvme_source)
```
karlowich, line 35: This could be simplified by using the `cwd`, instead of `cd builddir`

cijoe/workflows/benchmark-provisioning-windows.yaml
```diff
@@ -0,0 +1,33 @@
+---
+doc: |
+  Environment prep for Windows
+  =================================================================
+
+  Environment preparation consists of
+
+  * Retrieving and updating repositories for xnvme, fio, and SPDK
+  * Configure, build and install xnvme, fio, and SPDK
```
karlowich, line 9: We don't do anything with SPDK

karlowich (changes requested):
Nice!
A couple of comments

cijoe/scripts/fio_latency.py
NaddiNadja: the problem is exactly that `xnvme-driver` is not installed on windows (because it is not relevant), and therefore it gives the following error, making the whole workflow fail.
`The term 'xnvme-driver' is not recognized as the name of a cmdlet, function, script file, or operable program.`

so i had to make sure it is not run on windows to begin with

karlowich (approved):
Nice!

cijoe/scripts/sysinfo_report.py
```diff
@@ -215,7 +215,7 @@ def get_motherboard_cpu() -> Tuple[int, Tuple[str, str]]:
         cpu = ""
 
         err, state = cijoe.run(
-            f'get-computerinfo | findstr /B /C:"{CS_PROCESSORS}" /B /C:"{CS_MANUFACTURER}" /B /C:"{CS_MODEL}"'
+            f'powershell "get-computerinfo | findstr /B /C:\\"{CS_PROCESSORS}\\" /B /C:\\"{CS_MANUFACTURER}\\" /B /C:\\"{CS_MODEL}\\""'
```
safl, line 218: This is fine by me, however, would it be nicer if all the command-output was collected and then processed by the script? For example changing this to:

```python
err, state = cijoe.run('powershell "get-computerinfo"')
```

And then extract from ``state.output()`` the data needed? In other words, replacing the pipes with Python logic.

safl (commented):
Excellent! I don't see any blockers, this is excellent, I have a minor question.

cijoe/scripts/sysinfo_report.py
NaddiNadja: You are right, as we are already looping through the output lines, the pipe is actually unnecessary. I was worried about performance because some of the output can be very long, but I added a check to stop the loop when the desired values have been found and did some checks on performance, and it seems unchanged. :)

safl: Great! Btw. seems like the CI started complaining?

NaddiNadja: hehe woops, that's on me, I was using the windows machine to test some local changes which made it unable to rebase :)

karlowich (approved):

safl:
Does this still depend on changes to cijoe?

karlowich:
@safl the dependency on https://github.com/refenv/cijoe/pull/69 has been dropped with the comment:
> As we have switched to using bash as default on our Windows machine, this is not needed anymore.

safl:
> @safl the dependency on [refenv/cijoe#69](https://github.com/refenv/cijoe/pull/69) has been dropped with the comment:
> 
> > As we have switched to using bash as default on our Windows machine, this is not needed anymore.

The PR message still state it as a dependency, thus I wanted to confirm.

safl (approved):
lgtm.

PR #527, opened 2024-08-28
Remove outdated/irrelevant (and fix simple) todos

karlowich:
Todos should be GitHub issues and not live in the code, this is a first stab at removing some existing todos


NaddiNadja (approved):
assuming that the removed todos were outdated, this looks good to me :)

safl (approved):
LGTM.

PR #530, opened 2024-09-05
fix(build): drop rt dependency

### LINKED ISSUE #529: Drop linking with rt?
In a quick stab at cross-compiling for aarch64 I stumbled over on ``meson.build:151:12: ERROR: C shared or static library 'rt' not found``. Linking explicitly with ``rt`` is somethat that was previously done for compatibility with older systems, it should not be needed anymore.
So, try removing it and see what the CI catches.

karlowich:
closes #529

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -88,7 +88,6 @@ xnvmelib_deps = [
   aio_dep,
   spdk_dep,
   thread_dep,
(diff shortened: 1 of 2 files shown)


karlowich:
Closing this as it breaks `cbi_posix`

PR #533, opened 2024-09-28
feat(gha): run verify and docgen on github-hosted runners

safl:


mbrsamsung (approved):
LGTM

.github/workflows/verify.yml
```diff
@@ -943,7 +945,7 @@ jobs:
   #
   docgen:
     needs: [source-archive-with-subprojects, source-format-check, build-python]
-    runs-on: [self-hosted, linux, verify]
```
NaddiNadja, line 946: Why is the one on line 644 `[self-hosted, linux, x64, verify]`, but this one "only" `[self-hosted, linux, verify]`?

safl: I believe the 'runs-on' label for the verify-job was updated when the arch-runner was added to avoid scheduling it there. The docgen job should have also been set to x86 at that time.

Regardless, it's sufficient to schedule it on 'runs: ubuntu-latest' now, as all GitHub-hosted runners are x86, except for macOS, which these jobs won't use.

NaddiNadja (approved):
LGTM

karlowich (approved):
lgtm

PR #535, opened 2024-10-03
fixes for testing on real devices

NaddiNadja:
Our tests have not been run on real machines before, and it seems that qemu has been a bit "too" forgiving in some cases, resulting in tests passing even though they either (a) should not or (b) would not on all devices.

This PR should be the first step in the direction of being able to build and test xNVMe on real machines.


karlowich (approved):
Nice!

safl (approved):
Great!

PR #536, opened 2024-10-03
fix(docs): update diagram of machines used in CI

NaddiNadja:


karlowich (changes requested):
Looks good!

One thing:
Under HOSTING-QEMU-IMG we now also have FreeBSD 14 in addition to FreeBSD 13 and Debian Bullseye

safl (approved):
Excellent!

karlowich (approved):
Nice

PR #538, opened 2024-10-15
refactor(cijoe/bench): move shared functionality to imported script

NaddiNadja:
Bump cijoe to latest version and refactor shared functionality into imported script.

Our two workflows "latency" and "bench" that each generate graphs and reports, shared functionality in the plotting and reporting path of the workflows. This funcationliaty could not be put into a shared script before cijoe support the functionality of importing self-made scripts, which was added in v0.9.35.

### DIFF
--- a/cijoe/scripts/plotter.py
+++ b/cijoe/scripts/plotter.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python
+"""
+Helper functions for plotting benchmark graphs
+==============================================
+
+
+
+.. note::
+    This uses matplotlib and numpy for plotting
+"""
+import logging as log
+from pathlib import Path
+from typing import Dict, List
+
+import matplotlib.pyplot as plt
+from cijoe.core.resources import dict_from_yamlfile
+
+OUTPUT_NORMALIZED_FILENAME = "benchmark-output-normalized.json"
+PLOT_SCALE = 1.5
+BAR_WIDTH = 0.2
+
+Data = Dict[str, Dict[str, List[float | int] | str]]
+
+
+def data_as_a_function_of(data, x="iodepth", y="iops", filter=lambda _: True) -> Data:
+    """Organize and label data with 'y' points as a function of 'x' points"""
+
+    samples = {}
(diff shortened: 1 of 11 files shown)


NaddiNadja:
This is waiting for the next version of cijoe (v0.9.40+), where we've fixed that calling `powershell "...."` hangs forever.

safl (dismissed):
LGTM!

cijoe/scripts/plotter.py
```diff
@@ -0,0 +1,235 @@
...
+"""
+import logging as log
+from pathlib import Path
+from typing import Dict, List
+
+import matplotlib.pyplot as plt
+from cijoe.core.resources import dict_from_yamlfile
+
+OUTPUT_NORMALIZED_FILENAME = "benchmark-output-normalized.json"
+PLOT_SCALE = 1.5
```
karlowich, line 19: It seems that two scripts import this, but the original value is not 1.5 for both. `latency_plotter` originally had 1.3.
Is this a problem?

cijoe/scripts/reporter.py
```diff
@@ -1,64 +1,71 @@
-#!/usr/bin/env python3
-"""
-Report templates
-============================
-
-Create the general templates for the xNVMe reports.
-
-"""
```
karlowich: Is there a need to remove the the doc-string?

cijoe/scripts/plotter.py
NaddiNadja: No, it only defines how much "air" there is above the bars in the plot. 1.5 looks fine, in my opinion

NaddiNadja:
@safl and @karlowich , can you take a look again? The only things changed is that I added docs to the `cijoe/scripts/reporter.py` script, and that I've added the `--monitor` argument for all our cijoe runs in gha :)

karlowich (dismissed):
LGTM

safl (approved):
LGTM!

karlowich (approved):

PR #539, opened 2024-10-15
Provision using tgz in latency and benchmark jobs

NaddiNadja:
Previously, we provisioned using the github repository, which meant that new PRs with the "bench" tag did not benchmark the xnvme source contained in the PR, but instead the source on branch "next" in the repository.


cijoe/workflows/benchmark-provisioning-windows.yaml
karlowich, line 1: Do we still need the benchmark-provisioning files? Also their doc-strings are now outdated

NaddiNadja: We need the linux and freebsd one, because they do additional provisioning beside xnvme, and therefore we also need the windows one, even though it seems silly, but it's because the GHA step just says "run provisioning for {{os}}-{{version}}", so we need one for each of the three machines we run latency on. 

I guess we could do some "if it exists ...", but imo, it's better to have an explicitly empty file, which states that no extra provisioning is needed for Windows rather than only executing the provisioning if it exists. What if I accidentally delete a provisioning workflow for e.g. Linux, and then it looks like this step is OK in the GHA job, but it fails on the next one bc the machine was not provisioned properly.

karlowich: Okay, that's fine.
Let's update their doc-strings to describe what their actually doing

karlowich (approved):
Nice!

safl (approved):
Excellent, thanks!

PR #541, opened 2024-10-16
feat(build/spdk): bump to v24.09

karlowich:
This allows us to drop our out-of-tree patches.
I decided to drop everything related to the patches as a way of signifying that we shouldn't be adding new patches, and instead upstream any changes we need.


safl (changes requested):
Excellent with the version-bump and removal of patches that are now available upstream.

The infrastructure for applying out-of-tree patches to the custom SPDK version that xNVMe builds is very convenient, let's keep that.

karlowich:
I see your point, I will keep the patch scaffolding for ease of development in case we need to test potential changes in the future

safl (approved):
lgtm

NaddiNadja (approved):
lgtm :)

PR #542, opened 2024-10-21
vfio phy mapping

jwdevantier:
This PR introduces two fixes to the vfio (libvfn) backend.

The PR ensures that `xnvme_be_vfio_buf_alloc`, actually sets the PHY pointer when the physical address is also needed. This is when the user calls `xnvme_buf_phys_alloc` where the physical address is also needed.

This PR also fixes an issue where `xnvme_be_vfio_buf_vtophys` interpreted the return-code of `iommu_translate_vaddr`  in the wrong way (it returns 1/true on success).

### DIFF
--- a/lib/xnvme_be_vfio_mem.c
+++ b/lib/xnvme_be_vfio_mem.c
@@ -10,6 +10,25 @@
 #include <xnvme_be_vfio.h>
 #include <xnvme_dev.h>
 
+void
+xnvme_be_vfio_buf_free(const struct xnvme_dev *dev, void *buf)
+{
+	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
+	struct iommu_ctx *ctx = state->ctrl->pci.dev.ctx;
+	size_t len;
+
+	XNVME_DEBUG("xnvme_be_vfio_buf_free(%p, %p)", dev, buf);
+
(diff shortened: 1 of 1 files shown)


safl (approved):
This looks excellent. Perhaps @birkelund  has input on something I might not have considered on the libvfn side. Otherwise, plain awesome.

lib/xnvme_be_vfio_mem.c
```diff
@@ -68,7 +68,7 @@ xnvme_be_vfio_buf_vtophys(const struct xnvme_dev *dev, void *buf, uint64_t *phys
 	struct xnvme_be_vfio_state *state = (void *)dev->be.state;
 	struct iommu_ctx *ctx = state->ctrl->pci.dev.ctx;
 
-	if (iommu_translate_vaddr(ctx, buf, phys)) {
```
karlowich, line 71: Nice catch!

karlowich (approved):
LGTM

karlowich:
It would be great if we had something testing these code paths

safl:
> It would be great if we had something testing these code paths

Made an issue for it here: https://github.com/xnvme/xnvme/issues/545

safl:
Thanks @jwdevantier !

PR #549, opened 2024-11-14
fix(cijoe): add `./` as prefix to local file for `pipx inject`

NaddiNadja:
Without this prefix, the command fails on our physical Linux machine.


safl (commented):
LGTM. Added bench-run, to check whether it affects it.

karlowich (approved):
lgtm

safl (approved):
lgtm

Issue #227, opened 2023-01-18
Improve communication of backend capabilities


karlowich:
The CLI tool ``xnvme library-info`` prints the backends "enabled" with the xNVMe library. However, it is missing information about all the different **mixins**. That is, informing which asynchronous, synchronous, admin, and enumeration interfaces which can be utilized. This should be added and a section in the documentation describing how runtime-instrumentation works and how to probe xNVMe for the capabilities.

*This is from the 'ISSUES' file, see #220*

PR #551, opened 2024-11-15
refactor(libconf)!: replace extern xnvme_libconf with struct and getter

safl:
The library configuration previously relied on shared access to a global variable (xnvme_libconf) exposed directly via the API. This behavior has been replaced with a dedicated data structure and a getter function, allowing users to query the library for an instance holding the configuration entries.

The pretty-printers have been updated to take this structure as an argument, aligning their behavior with other pretty-printers. This refactor addresses technical debt and prepares the codebase for upcoming API and FFI improvements.

Additionally, this change resolves an issue where gen_flags() was matching too broadly.


NaddiNadja (dismissed):
lgtm :)

lib/xnvme_libconf.c
```diff
@@ -28,37 +35,29 @@ xnvme_libconf_fpr(FILE *stream, enum xnvme_pr opts)
...
 	wrtn += fprintf(stream, "\n");
-	for (int i = 0; xnvme_libconf[i]; ++i) {
-		fprintf(stream, "  - '%s'\n", xnvme_libconf[i]);
+	wrtn += fprintf(stream, "  entries:\n");
+	for (int i = 0; libconf->entries[i]; ++i) {
+		fprintf(stream, "  - '%s'\n", libconf->entries[i]);
 	}
 
 	// TODO: this is kinda hackish... a better means for backends to provide
 	// various introspective information should be provided
```
karlowich: Let's remove this TODO. We already have an issue: #227

lib/xnvme_libconf_entries.c
```diff
@@ -1,11 +1,8 @@
-// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
-//
-// SPDX-License-Identifier: BSD-3-Clause
-
 ///< This file is autogenerated by 'toolbox/xnvme_libconf.py'
```
karlowich, line 1: Since the file is autogenerated, I propose that we remove it from version control

toolbox/xnvme_libconf.py
```diff
@@ -171,13 +172,40 @@ def update(args):
...
+#endif
+#ifdef NVME_IOCTL_IO64_CMD_VEC
+    "3p: NVME_IOCTL_IO64_CMD_VEC",
+#endif
+#ifdef NVME_IOCTL_ADMIN64_CMD
+    "3p: NVME_IOCTL_ADMIN64_CMD",
+#endif
+    """
+
+    """
```
karlowich: This `"""` looks to be here by mistake.

karlowich: Ah, I see this manually coded in `xnvme_libconf.c` - what is the reason for this?

lib/meson.build
```diff
@@ -70,7 +70,6 @@ xnvmelib_source = [
   'xnvme_kvs.c',
   'xnvme_lba.c',
   'xnvme_libconf.c',
-  'xnvme_libconf_entries.c',
```
karlowich, line 73: I don't understand why this is being removed, don't we need it?

karlowich (changes requested):

lib/xnvme_libconf_entries.c
safl: To achieve that, additional work is required to create the file when needed, which is beyond the scope of this PR.

lib/xnvme_libconf.c
safl: Right, ok. It makes a note an another issue with the library-config implementation, I have added that as an addendum to https://github.com/xnvme/xnvme/issues/227 and removed the TODO.

toolbox/xnvme_libconf.py
safl: This is the same as the addendum on https://github.com/xnvme/xnvme/issues/227

It should be produced here, and not in the pretty-printer. However, it cannot do so trivially.

Removed the commented out block as it now tracked on the issue #227

lib/meson.build
safl: It is imported by ``xnvme_libconf.c``.
Previously then the file provided the implementation backing the ``extern struct``. However, now that this is removed and instead retrieved via a getter, then ``xnvme_libconf_entries.c`` is not translated, instead ``xnvme_libconf.c`` is translated which includes ``xnvme_libconf_entries.c``.

lib/xnvme_libconf_entries.c
karlowich: I see - then it is fine

lib/meson.build
karlowich: Great

karlowich (approved):
lgtm

PR #552, opened 2024-11-18
fix(lib): format characters

safl:


lib/xnvme_be_windows_async_ioring.c
```diff
@@ -36,7 +36,7 @@ xnvme_be_windows_ioring_init(struct xnvme_queue *q, int XNVME_UNUSED(opts))
 
 	if (!handle) {
 		err = GetLastError();
-		XNVME_DEBUG("Failed creating IO ring handle: 0x%x\n", err);
+		XNVME_DEBUG("Failed creating IO ring handle: %" PRId32, err);
```
NaddiNadja, line 39: There was a line-break `\n` at the end of this string before (and in many of the other places `PRId32` has been inserted) - is it OK that it's gone now?

NaddiNadja (commented):
just a single thing I'm unsure of :)

lib/xnvme_be_windows_async_ioring.c
safl: Yeah, they should not be there to begin with since ``XNVME_DEBUG`` already outputs a newline.

NaddiNadja (approved):
lgtm :)

karlowich (approved):
lgtm

PR #554, opened 2024-11-19
fix(api/ident): remove unused macro definitions

safl:
These definitions were originally used when xNVMe employed a "schema" to encode options into the device identifier string, e.g.:

  "/dev/nvme0n1?nsid=2&async=io_uring"

This approach was replaced by `xnvme_opts`. However, these definitions have lingered since then and are now being removed as part of the API cleanup in preparation for code generation.

### DIFF
--- a/include/libxnvme_ident.h
+++ b/include/libxnvme_ident.h
@@ -8,14 +8,6 @@
 
 #define XNVME_IDENT_URI_LEN 384
 
-#define XNVME_IDENT_SCHM_LEN 5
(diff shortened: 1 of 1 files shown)


NaddiNadja (commented):
lgtm :)

karlowich (approved):
lgtm

safl:
> lgtm :)

If you submit this as a review instead of a comment, your feedback will count towards the two required reviews.

PR #556, opened 2024-11-26
fix(toolbox/pkgs): add automake to debian bookworm dependencies

NaddiNadja:
The SPDK configuration script fails when configuring ISA-L without automake because `aclocal` is not found.

----

I am unsure whether this is necessary in other pkgs scripts?


safl (approved):
LGTM. Similar notes could be added if needed when qcow images are built for other distributions or operating systems.

Issue #518, opened 2024-07-26
Add sanitilize log cli and pretty-printers


safl:
To add support for sanitize-log then the following needs to be done:

# Tasks

- [x] Add data-structure to ``libxnvme_spec.h``
- [x] Add a pretty-printer for the log-page structure in ``lib/xnvme_spec.c``
- [x] Add a sub-command to the ``xnvme`` cli-tool that sends get-log-page for the sanitize command status in ``tools/xnvme.c`` using ``xnvme_prep_adm_log()`` and the pretty-printer added above

# For guidance

The **API** function (``xnvme_prep_adm_log()``) takes care of the command-construction, submission, and completion, examples of its use are available in the cli-tools ``xnvme``, below are a couple of examples:

- S.M.A.R.T. / Health Information (https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/tools/xnvme.c#L232)
- Error Information: https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/tools/xnvme.c#L273
- FDP Config (https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/tools/xnvme.c#L324)

Examples of "pretty-printers", these live together with the layout definitions from the spec, here is an example:

* Data-structure definition
https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/include/libxnvme_spec.h#L195
* Pretty-printers
  - https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/lib/xnvme_spec.c#L23
  - https://github.com/xnvme/xnvme/blob/5fbaf6a0960a567fdb377b2aeb82b8838d674f3b/lib/xnvme_spec.c#L81

karlowich:
#561 has been merged, thus closing this.

PR #526, opened 2024-08-15
Added log-sanitize option

vikramkhurana-vigoursoft:
log-sanitize option to monitor the status of sanitize command

PR #561, opened 2024-12-16
feat(spec,tools): add log-sanitize

### LINKED ISSUE #525: Adding Sanitize Log CLI and Pretty Printers
Adding a log-sanitize CLI to query the status of the initiated Sanitize command. 
It allows the requester to know when the sanitize operation completes and the estimated time values based on the type of request.

karlowich:
Adaptation of #526 and closes #525

Thanks @vikramkhurana-vigoursoft

### DIFF
--- a/cijoe/tests/tools/test_xnvme.py
+++ b/cijoe/tests/tools/test_xnvme.py
@@ -54,7 +54,7 @@ def test_idfy_ctrlr(cijoe, device, be_opts, cli_args):
 def test_idfy_cs(cijoe, device, be_opts, cli_args):
     if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
         pytest.skip(reason="[admin=block] does not implement idfy-cs")
-    if "ramdisk" in device["labels"]:
+    if be_opts["be"] == "ramdisk":
         pytest.skip(reason="[be=ramdisk] does not implement idfy-cs")
 
     err, _ = cijoe.run(f"xnvme idfy-cs {cli_args}")
@@ -68,7 +68,7 @@ def test_format(cijoe, device, be_opts, cli_args):
     elif be_opts["be"] == "spdk" and "fabrics" in device["labels"]:
         pytest.skip(reason="spdk fabrics does not support format")
 
-    if "ramdisk" in device["labels"]:
+    if be_opts["be"] == "ramdisk":
         pytest.skip(reason="[be=ramdisk] does not implement format")
 
     err, _ = cijoe.run(f"xnvme format {cli_args}")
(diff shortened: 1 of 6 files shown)


karlowich:
Closes #518

cijoe/tests/tools/test_xnvme.py
```diff
@@ -79,10 +79,18 @@ def test_format(cijoe, device, be_opts, cli_args):
...
+
+    pytest.skip(reason="QEMU NVMe devices doesn't support sanitize")
+
+    err, _ = cijoe.run(f"xnvme sanitize --sanact 0x2 {cli_args}")
+
+    assert not err
 
-    pytest.skip(reason="TODO: always fails. Investigate.")
+    # Get log sanitize status log page
+    err, _ = cijoe.run(f"xnvme log-sanitize {cli_args} --nsid {device['nsid']}")
```
NaddiNadja: Is test_sanitize still always skipped? And what if the test is run on non-QEMU devices?

karlowich: Right, perhaps it would be better for the test to need a device with label `sanitize`.

mbrsamsung (dismissed):
lgtm

cijoe/tests/tools/test_xnvme.py
```diff
@@ -75,14 +75,20 @@ def test_format(cijoe, device, be_opts, cli_args):
     assert not err
 
 
-@xnvme_parametrize(labels=["nvm"], opts=["be", "admin"])
+@xnvme_parametrize(labels=["nvm", "sanitize"], opts=["be", "admin"])
 def test_sanitize(cijoe, device, be_opts, cli_args):
     if be_opts["be"] == "linux" and be_opts["admin"] in ["block"]:
         pytest.skip(reason="[admin=block] does not implement sanitize")
+    if "ramdisk" in device["labels"]:
+        pytest.skip(reason="[be=ramdisk] does not implement sanitize")
```
NaddiNadja: Maybe stupid question, but why is the reason "[be=ramdisk] (...)", when this is not the condition we are looking for? Should the if-statement be `if be_opts["be"] == "ramdisk"`, or should the reason maybe be changed? 
(and also, is ramdisk both as a label *and* a backend?)

NaddiNadja (commented):
single question

cijoe/tests/tools/test_xnvme.py
karlowich: You are right, this has been fixed now

NaddiNadja (approved):
nice lgtm :)

safl (approved):
lgtm, made a note of a potential issue here: https://github.com/xnvme/xnvme/issues/578
But no reason to keep this back.

Issue #480, opened 2024-05-21
Re-deploy verification runners


safl:
Documentation states that runners live in ``/scratch/ghr/openmpdk/xnvme/runners`` this is a poorly named path as the project is no longer in the OpenMPDK github. Thus, eventually, then this should be "cleaned" or runners re-deployed. Note that this is a very minor thing, and this is just to make a note of the discrepancy as the documentation will be changed to no longer include "openmpdk".

PR #550, opened 2024-11-14
feat(ci/gha): add tests on physical linux machine

NaddiNadja:
The "verify-macos" job is converted to a "verify-physical" job that has a matrix for all our test machines (atm 2 machines).

PR #562, opened 2024-12-19
fix(docs/CI): remove outdated info about runners

karlowich:
This addresses #480.

We no longer have a `bench-amd` machine, thus removing this.

The verify job is now using hosted runners; so, we don't need a section about how to setup the self-hosted runners.


NaddiNadja:
As part of #550, I'm updating this whole section to match our current setup, so I don't think this is necessary :)

safl:
> As part of #550, I'm updating this whole section to match our current setup, so I don't think this is necessary :)

Closing it then.

PR #566, opened 2024-12-21
fix(be/spdk): don't detach in-use enumerated controllers

### LINKED ISSUE #565: SPKD backend unconditionally detaches from controller during enumeration even if XNVME_ENUMERATE_DEV_KEEP_OPEN is specified
My understanding is that when `XNVME_ENUMERATE_DEV_KEEP_OPEN` is returned from the callback passed to  `xnvme_enumerate()` , the device is kept open so that it can be used directly without having to re-open it using a URI. This is particularly useful if trying to identify a device by model and serial number rather than bus location.

However, while the SPDK backend will nominally leave the device open, it will unconditionally detach the underlying controller before returning from `xnvme_enumerate()`. This leaves the device in a broken and unusable state.

I believe there are two problems in `enumerate_attach_cb()` in [xnvme_be_spdk_dev.c](../blob/main/lib/xnvme_be_spdk_dev.c):
1) It will always decrement the refcount on the controller using `_cref_deref()` even if device closure has been skipped.
2) In order to avoid detaching the controller in the middle of enumeration, the second argument to `_cref_deref()` is false. Then, once the device enumeration loop for the controller is complete, there is an unconditional call to `spdk_nvme_detach()`.

The relevant code:
```c
		dev = xnvme_dev_open(ident.uri, &opts);
		if (!dev) {
			XNVME_DEBUG("FAILED: xnvme_dev_open(), err: %d", errno);
			return;
		}
		if (ectx->enumerate_cb(dev, ectx->cb_args)) {
			xnvme_dev_close(dev);
		}

		err = _cref_deref(ctrlr, false);
		if (err) {
			XNVME_DEBUG("FAILED: _cref_deref():, err: %d", err);
		}
	}

	if (spdk_nvme_detach(ctrlr)) {
		XNVME_DEBUG("FAILED: spdk_nvme_detach()");
	}
```

I believe two changes are warranted:
1) `_cref_deref()` should only be called if the device is closed.
2) After the enumeration loop for the controller is complete, there should be a way to detect if the reference count is non-zero, and only detach the controller if the reference count is zero.

jadoc:
If `XNVME_ENUMERATE_DEV_KEEP_OPEN` is returned from the callback during enumeration, the SPDK controller underlying the open device must not be detached. Only deref the controller if the device is closed. Conditionally detach controllers only if their reference count is zero.

It is now valid for a controller to have a reference count of zero and not be cleaned up yet. Entries in the reference count array may not be reused if the controller pointer is non-null. A controller with a zero reference count may be still be incremented later. A new cleanup routine allows deferred detachment of all controllers with a zero refcount.

This change is an attempt to fix #565.

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -26,23 +26,31 @@
 static struct xnvme_be_spdk_ctrlr_ref g_cref[XNVME_BE_SPDK_CREFS_LEN];
 
 /**
- * look for a cref for the given given ident
+ * Increment the refcount for the ctrlr with the given ident. If ctrlr is NULL, only lookup based
+ * on ident.
  *
- * @return On success, a pointer to the cref is returned. When none exists, NULL is returned.
+ * @return On success, a pointer to the ctrlr is returned. On error, NULL is returned.
  */
 static struct spdk_nvme_ctrlr *
-_cref_lookup(struct xnvme_ident *id)
+_cref_ref(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
 {
+	int free_pos = -1;
 	for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
-		if (strncmp(g_cref[i].uri, id->uri, XNVME_IDENT_URI_LEN - 1)) {
+		if (!g_cref[i].ctrlr) {
+			if (free_pos < 0) {
+				free_pos = i;
+			}
 			continue;
 		}
-		if (!g_cref[i].ctrlr) {
-			XNVME_DEBUG("FAILED: corrupted ctrlr in cref");
+		if (strncmp(g_cref[i].uri, ident->uri, XNVME_IDENT_URI_LEN - 1)) {
+			continue;
+		}
+		if (g_cref[i].refcount < 0) {
+			XNVME_DEBUG("FAILED: corrupted refcount");
 			return NULL;
 		}
-		if (g_cref[i].refcount < 1) {
-			XNVME_DEBUG("FAILED: corrupted refcount");
+		if (ctrlr != NULL && ctrlr != g_cref[i].ctrlr) {
+			XNVME_DEBUG("FAILED: multiple ctrlr with same ident");
 			return NULL;
 		}
 
@@ -51,38 +59,28 @@ _cref_lookup(struct xnvme_ident *id)
 		return g_cref[i].ctrlr;
 	}
 
-	return NULL;
+	if (!ctrlr) {
+		XNVME_DEBUG("FAILED: no matching ctrlr and no ctrlr provided");
+		return NULL;
+	}
+
+	if (free_pos < 0) {
+		XNVME_DEBUG("FAILED: out of slots");
+		return NULL;
+	}
+
+	g_cref[free_pos].ctrlr = ctrlr;
+	g_cref[free_pos].refcount += 1;
+	strncpy(g_cref[free_pos].uri, ident->uri, XNVME_IDENT_URI_LEN);
+
+	return g_cref[free_pos].ctrlr;
 }
 
 /**
- * Insert the given controller
+ * Decrement the refcount for the given ctrlr.
  *
- * @return On success, 0 is returned. On error, negated ``errno`` is returned to
- * indicate the error.
+ * @return On success, zero returned. On error, -errno is returned.
  */
-static int
-_cref_insert(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
-{
-	if (!ctrlr) {
-		XNVME_DEBUG("FAILED: !ctrlr");
-		return -EINVAL;
-	}
-
-	for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
-		if (g_cref[i].refcount) {
-			continue;
-		}
-
-		g_cref[i].ctrlr = ctrlr;
-		g_cref[i].refcount += 1;
-		strncpy(g_cref[i].uri, ident->uri, XNVME_IDENT_URI_LEN);
-
-		return 0;
-	}
-
-	return -ENOMEM;
-}
-
 int
 _cref_deref(struct spdk_nvme_ctrlr *ctrlr, bool detach_on_no_refcount)
 {
@@ -103,10 +101,10 @@ _cref_deref(struct spdk_nvme_ctrlr *ctrlr, bool detach_on_no_refcount)
 
 		g_cref[i].refcount -= 1;
 
-		if (g_cref[i].refcount == 0) {
-			if (detach_on_no_refcount) {
-				XNVME_DEBUG("INFO: refcount: %d => detaching", g_cref[i].refcount);
-				spdk_nvme_detach(ctrlr);
+		if (g_cref[i].refcount == 0 && detach_on_no_refcount) {
+			XNVME_DEBUG("INFO: refcount: %d => detaching", g_cref[i].refcount);
+			if (spdk_nvme_detach(ctrlr)) {
+				XNVME_DEBUG("FAILED: spdk_nvme_detach()");
 			}
 			memset(&g_cref[i], 0, sizeof g_cref[i]);
 		}
@@ -116,6 +114,31 @@ _cref_deref(struct spdk_nvme_ctrlr *ctrlr, bool detach_on_no_refcount)
 
 	XNVME_DEBUG("FAILED: no tracking for %p", (void *)ctrlr);
 	return -EINVAL;
+}
+
+int
+_cref_cleanup()
+{
+	for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
+		if (!g_cref[i].ctrlr) {
+			continue;
+		}
+
+		if (g_cref[i].refcount < 0) {
+			XNVME_DEBUG("FAILED: invalid refcount: %d", g_cref[i].refcount);
+			continue;
+		}
+
+		if (g_cref[i].refcount == 0) {
+			XNVME_DEBUG("INFO: refcount: %d => detaching", g_cref[i].refcount);
+			if (spdk_nvme_detach(g_cref[i].ctrlr)) {
+				XNVME_DEBUG("FAILED: spdk_nvme_detach()");
+			}
+			memset(&g_cref[i], 0, sizeof g_cref[i]);
+		}
+	}
+
+	return 0;
 }
 
 static int g_xnvme_be_spdk_transport[] = {
@@ -453,7 +476,6 @@ attach_cb(void *cb_ctx, const struct spdk_nvme_transport_id *trid, struct spdk_n
 	struct xnvme_opts *opts = &dev->opts;
 	struct xnvme_be_spdk_state *state = (void *)dev->be.state;
 	struct spdk_nvme_ns *ns = NULL;
-	int err;
 
 	XNVME_DEBUG("INFO: nsid: %d", opts->nsid);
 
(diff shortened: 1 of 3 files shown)


karlowich:
Hi @jadoc, thank you very much for your contribution!
Overall it looks very good, however there seems to be a bug when enumerating devices with multiple namespaces causing the same ctrlr to be detached more than once.
```bash
...
  - {uri: '0000:03:00.0', dtype: 0x2, nsid: 0x1, csi: 0x0, subnqn: 'nqn.2019-08.org.qemu:deadbeef'}
# DBG:xnvme_be.c:xnvme_be_factory-276: INFO: checking be: 'spdk'
# DBG:xnvme_be.c:xnvme_be_factory-307: INFO: obtained backend instance be: 'spdk'
# DBG:xnvme_be_spdk_dev.c:_spdk_env_init-245: INFO: already initialized
# DBG:xnvme_be_spdk_dev.c:_spdk_env_init-270: INFO: spdk_env_is_initialized: 1
# DBG:xnvme_be_spdk_dev.c:xnvme_be_spdk_state_init-807: INFO: found dev->ident.uri: '0000:03:00.0' via cref_lookup()
# DBG:xnvme_be_spdk_dev.c:xnvme_be_spdk_state_init-851: INFO: re-using previously attached controller
# DBG:xnvme_be_spdk_dev.c:xnvme_be_spdk_state_init-901: INFO: open() : OK
# DBG:xnvme_be.c:xnvme_be_factory-312: INFO: obtained device handle
# DBG:xnvme_dev.c:_dev_idfy_csi-171: INFO: looks like csi(ZNS)
  - {uri: '0000:03:00.0', dtype: 0x2, nsid: 0x2, csi: 0x2, subnqn: 'nqn.2019-08.org.qemu:deadbeef'}
# DBG:xnvme_be_spdk_dev.c:_cref_deref-107: INFO: refcount: 0 => detaching
# DBG:xnvme_be_spdk_dev.c:_cref_cleanup-136: INFO: refcount: 0 => detaching
xnvme: nvme.c:61: nvme_ctrlr_detach_async: Assertion `ref_count > 0' failed.
Aborted
```
As seen in the debug output above `spdk_nvme_detach()` is called both from `_cref_deref()` and `_cref_cleanup()`.

jadoc:
I don't think there's an error in the new logic per se, but I inadvertently unmasked an existing bug. The current enumeration logic calls `_cref_insert()` once per namespace on the same controller, which I don't believe is safe. This has been leaving junk in `g_cref`. Only now that I'm walking `g_cref` with `_cref_cleanup()` is the corruption becoming apparent.

I think the answer is to replace both `_cref_insert()` and `_cref_lookup()` with a common `_cref_ref()` that safely does the right thing depending on whether the controller currently exists in the table.

karlowich:
@jadoc Yes, it appears you are right - the logic was broken before your changes.
I agree that what we need is some logic to either increment the refcount if the ctrlr exists OR insert it, instead of potentially inserting the same controller multiple times.
As for merging the two functions, I cannot decide if it is nicer or not.

karlowich:
Changing `_cref_insert()` to the below fixes the assertion error:
```C
static int
_cref_insert(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
{
        if (!ctrlr) {
                XNVME_DEBUG("FAILED: !ctrlr");
                return -EINVAL;
        }

        for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
                if (g_cref[i].ctrlr) {
                        if (g_cref[i].ctrlr != ctrlr) {
                                continue;
                        }
                        g_cref[i].refcount += 1;
                        return 0;
                }

                g_cref[i].ctrlr = ctrlr;
                g_cref[i].refcount += 1;
                strncpy(g_cref[i].uri, ident->uri, XNVME_IDENT_URI_LEN);

                return 0;
        }

        return -ENOMEM;
}
```

jadoc:
@karlowich
> Changing `_cref_insert()` to the below fixes the assertion error

Your proposed fix doesn't work in the case where a previous entry in the list has been freed. We must compare every entry in the list before concluding that the controller isn't already in the table. This is the best we can do unless this were replaced with a hash table. [^resizeable]

I have proposed an implementation of `_cref_ref()` that safely combines both the insert and lookup use cases. Let me know what you think.

There should probably be an explicit tests for the following. I'm not familiar with the test framework used by the project, so I'll have to figure out how much effort that will be.
1) Keeping devices open after enumeration.
2) Closing old devices before opening new ones.


[^resizeable]: Actually, using a resizable hash table like [uthash](/troydhanson/uthash) is probably a good idea regardless.

lib/xnvme_be_spdk_dev.c
```diff
@@ -26,63 +26,61 @@
...
  */
 static struct spdk_nvme_ctrlr *
-_cref_lookup(struct xnvme_ident *id)
+_cref_ref(struct xnvme_ident *ident, struct spdk_nvme_ctrlr *ctrlr)
 {
+	int free_pos = -1;
 	for (int i = 0; i < XNVME_BE_SPDK_CREFS_LEN; ++i) {
-		if (strncmp(g_cref[i].uri, id->uri, XNVME_IDENT_URI_LEN - 1)) {
+		if (!g_cref[i].ctrlr) {
+			if (free_pos < 0 && !g_cref[i].ctrlr) {
```
karlowich:
```suggestion
			if (free_pos < 0) {
```
`!g_cref[i].ctrlr` will always be true if we hit this path

karlowich (changes requested):
This is great!
One minor nitpick.

karlowich:
@jadoc Thanks again for your contribution!

> I'm not familiar with the test framework used by the project

Our tests are primarily C programs residing in `xnvme/tests`, I think the functionality we need to test fits in `xnvme/tests/enum`.
The tests are orchestrated from python in `xnvme/cijoe/tests`, i.e. `xnvme/cijoe/tests/logic/test_xnvme_tests_enum.py`

jadoc:
@karlowich 
> Our tests are primarily C programs residing in `xnvme/tests`, I think the functionality we need to test fits in `xnvme/tests/enum`. The tests are orchestrated from python in `xnvme/cijoe/tests`, i.e. `xnvme/cijoe/tests/logic/test_xnvme_tests_enum.py`

The problem I was having was with CIJOE running QEMU. My QEMU installation doesn't support NVMe key-value command set. I don't know if this is a matter of building the latest version of QEMU from source or if a special patch set is necessary to make it work. Alternatively, maybe I can disable the KV devices and just run the specific tests in question, but those enumeration tests may expect the KV devices to be there.

karlowich (dismissed):

karlowich:
@jadoc We're using the `for-xnvme` branch from this [fork](https://github.com/SamsungDS/qemu).

jadoc:
I tried adding tests, but it looks like keeping devices open is broken for linux and vfio backends as well. I guess I'll have to mark those as skip for now.

karlowich:
@jadoc This is great!
It looks like the part failing on linux is not related to keeping the devices open, the error is: 
`errno: -25, msg: 'Inappropriate ioctl for device'`.
This looks like maybe the flush is the problem, depending on which I/O interface is chosen this might not be supported.
I suggest sending an identify command instead of the flush.

For the vfio backend enumeration is not supported, so we expect it to fail.

tests/enum.c
```diff
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <libxnvme.h>
 
+#define MAX_DEVS 100
```
safl, line 8: nice.

tests/enum.c
```diff
@@ -136,6 +137,76 @@ test_enum_open(struct xnvme_cli *cli)
...
+			continue;
+		}
+		++found;
+		if (cli->args.verbose) {
+			xnvme_dev_pr(devs[i], XNVME_PR_DEF);
+		}
+
+		// Send a simple flush just to make sure the device really is open.
+		ctx = xnvme_cmd_ctx_from_dev(devs[i]);
+		ctx.cmd.common.opcode = XNVME_SPEC_NVM_OPC_FLUSH;
```
safl: The flush command will fail on NVMe devices that don't support the NVM command set (e.g., KV or computation storage devices). As @karlowich suggested, consider using an admin command instead. This ensures the check is independent of the device's command set support during testing.

One downside is that you need to allocate a buffer for the payload. However, you can mitigate this by allocating the buffer before the loop, reusing it throughout, and deallocating it afterward.

For the command itself, you can use a helper like ``xnvme_adm_idfy_ctrlr()``, which constructs a mandatory identify controller command.

safl (commented):
Thanks for this, @jadoc! This is an excellent contribution.

I’ve added an inline suggestion with additional details to @karlowich’s proposal on using an identify command. Other than that, please proceed with signing off the commits.

Again, thank you!

tests/enum.c
jadoc: > The flush command will fail on NVMe devices that don't support the NVM command set (e.g., KV or computation storage devices).

That's interesting because in my reading of the NVMe base specification version 2.1, section 7 states that flush is valid for all command sets, and section 3.1.3.4 states that support for flush is mandatory for all I/O controllers. Also, if it were a matter of underlying device support, I would expect the command to fail with the SPDK backend, but flush apparently succeeds for all devices with SPDK.

The difference in behavior between the Linux backend and the SPDK backend is that enumeration on SPDK only returns devices of type `XNVME_DEV_TYPE_NVME_NAMESPACE` while Linux returns a mixture of  `XNVME_DEV_TYPE_NVME_CONTROLLER` and `XNVME_DEV_TYPE_NVME_NAMESPACE`. If I locally tweak the test to skip over the `XNVME_DEV_TYPE_NVME_CONTROLLER` devices, it passes. It appears that `XNVME_DEV_TYPE_NVME_CONTROLLER` corresponds to "Administrative controller" as defined in section 1.5.5 of the base spec.

Anyway, I'll switch to using an admin identify controller command.

jadoc: I decided not to allocate the buffer outside the loop, as `xnvme_buf_alloc()` allows for device-specific behavior.

If the tests pass, I'll rebase into a single commit.

safl: > > The flush command will fail on NVMe devices that don't support the NVM command set (e.g., KV or computation storage devices).
> 
> That's interesting because in my reading of the NVMe base specification version 2.1, section 7 states that flush is valid for all command sets, and section 3.1.3.4 states that support for flush is mandatory for all I/O controllers. Also, if it were a matter of underlying device support, I would expect the command to fail with the SPDK backend, but flush apparently succeeds for all devices with SPDK.
> 

You are absolutely right, my memory betrayed me.

karlowich:
@jadoc The only thing that fails is the vfio backend.
I think adding a skip is appropriate.
```PYTHON
if be_opts["be"] in ["vfio"]:
        pytest.skip(reason=f"[be={be_opts['be']}] enumeration not supported")
```

jadoc:
@karlowich
> For the vfio backend enumeration is not supported, so we expect it to fail.

Something doesn't add up. I see other enumeration tests pass for vfio. I think there may actually be a bug with keeping vfio devices open. Anyway, I won't attempt to address it here.

karlowich:
@jadoc It is true that some tests look like they enumerate successfully with vfio, this is a bug in `xnvme_enumerate()`.
I will fix that in a separate PR.

karlowich (dismissed):
Great!

jadoc:
@karlowich
> @jadoc It is true that some tests look like they enumerate successfully with vfio, this is a bug in `xnvme_enumerate()`. I will fix that in a separate PR.

Ah, I think the problem is in the tests. They don't actually verify that a non-zero number of devices were enumerated, which seems pretty important. I'm unable to debug the vfio stuff myself properly because my test rig doesn't have an IOMMU and I'm having trouble getting QEMU to emulate one, at least when combined with my CPU.

karlowich:
@jadoc I have fixed the bug here: #571 
Also, we prefer if you rebase on next instead of the merge commit.

jadoc:
@karlowich 
> we prefer if you rebase on next instead of the merge commit.

Done.

karlowich (approved):

safl (approved):
Thanks @jadoc !

PR #568, opened 2024-12-28
fix(cijoe/enum): remove be opt from backend test

karlowich:
The test doesn't use the flag, so no reason to include it


safl (approved):
lgtm.

Issue #319, opened 2023-08-08
Completion-handling via eventfd


safl:
API Changes for eventfd completion

- [ ] Experimental API to handle eventfd
  - [ ] Helper-function returning eventfd from queue ```int xnvme_queue_get_eventfd(struct xnvme_queue *queue);```
  - [ ] Some other heplers needed?

Example Application

- [ ] Cli-tool using event-driven I/O to read a single lba
  - In the same manner as the cli programs in ``examples/*.c``
  - Possibly similar to ``examples/xnvme_single_async.c``

Initial backend implementation in:

- [ ] The vfio-backend (libvfn)
- [ ] The io_uring backend (io_uring)
- [ ] The io_uring backend command (io_uring_cmd)

karlowich:
@safl Can we close this? or do we wait for tests using libblkio?

safl:
> [@safl](https://github.com/safl) Can we close this? or do we wait for tests using libblkio?

It is still open on libblkio https://gitlab.com/libblkio/libblkio/-/merge_requests/168
So, lets keep it open for now as well.

PR #570, opened 2025-01-09
feat(uring & libaio): add get completion_fd

karlowich:
Add support for registering an eventfd to use as completion_fd.
Related to #319

### DIFF
--- a/lib/xnvme_be_linux_async_liburing.c
+++ b/lib/xnvme_be_linux_async_liburing.c
@@ -12,6 +12,8 @@
 #include <pthread.h>
 #include <errno.h>
 #include <liburing.h>
+#include <sys/eventfd.h>
+#include <unistd.h>
 #include <xnvme_queue.h>
 #include <xnvme_dev.h>
 #include <xnvme_be_linux_liburing.h>
@@ -77,6 +79,8 @@ xnvme_be_linux_liburing_init(struct xnvme_queue *q, int opts)
 		XNVME_DEBUG("FAILED: lock(g_sqpoll_wq.mutex), err: %d", err);
 		return -err;
 	}
+
+	queue->efd = -1;
 
 	queue->batching = 1;
 	if (getenv("XNVME_QUEUE_BATCHING_OFF")) {
@@ -177,6 +181,13 @@ xnvme_be_linux_liburing_term(struct xnvme_queue *q)
 	if (queue->poll_sq) {
 		io_uring_unregister_files(&queue->ring);
 	}
+
+	if (queue->efd != -1) {
+		io_uring_unregister_eventfd(&queue->ring);
+		close(queue->efd);
+		queue->efd = -1;
+	}
+
 	io_uring_queue_exit(&queue->ring);
 
 	if (queue->poll_sq && g_sqpoll_wq.is_initialized && (!(--g_sqpoll_wq.refcount))) {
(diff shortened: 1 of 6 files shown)


lib/xnvme_be_linux_async_liburing.c
```diff
@@ -392,6 +402,37 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
...
+
+	q->efd = efd;
+
+	if (io_uring_register_eventfd(&q->ring, efd)) {
+		XNVME_DEBUG("FAILED: failed to register eventfd");
+		close(efd);
+		return -errno;
+	}
+
+	q->batching = 0;
```
safl, line 436: If the user has submitted one or more commands without calling ``poke()``, there will be unsubmitted commands left in the queue. To prevent this, consider making this part more defensive.

For example, it could check for outstanding I/O and either submit the commands or return an error like ``EBUSY`` to indicate the queue must be cleared before calling ``get_completion_fd()``. I would lean towards the latter.

karlowich: The outstanding commands would be submitted on the next call to `...cmd_io()`/`...cmd_iov()`.
Is that not good enough?
To me, returning `-EBUSY` when trying to get a `completion_fd` doesn't map intuitively to "there are outstanding commands".

safl: They would, if it were ever invoked. However, it is likely that someone might jump into the API and perform a code experiment, such as invoking ``cmd_io()`` one or more times, then calling ``xnvme_queue_get_completion_fd()`` to retrieve an eventfd to start waiting for those command submissions to complete. This would simply result in a hang, with no information provided.

To avoid this, I recommend using the error value, which is also utilized to indicate: "the queue is doing something else, so you cannot set up or retrieve an eventfd at this time." By adding this information to the docstring of ``xnvme_queue_get_completion_fd()``, users would be guided toward the correct approach.

NaddiNadja (dismissed):
lgtm :)

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -392,6 +403,47 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
...
+{
+	struct xnvme_queue_liburing *q = (struct xnvme_queue_liburing *)queue;
+	int efd, err;
+
+	if (q->efd != -1) {
+		return q->efd;
+	}
+
+	if (q->batching) {
+		XNVME_DEBUG("Completion FD requested while batching is on, submitting any "
```
safl: I still don't think xnvme_queue_get_completion_fd() should be submitting I/O. It would be preferable to add a guard in xnvme_queue.c that returns EBUSY, to avoid redundancy and ensure consistent behavior across all backends.

safl (changes requested):

lib/xnvme_be_linux_async_liburing.c
safl: > I still don't think xnvme_queue_get_completion_fd() should be submitting I/O. It would be preferable to add a guard in xnvme_queue.c that returns EBUSY, to avoid redundancy and ensure consistent behavior across all backends.

Adding it to xnvme_queue.c might not be the way to go. However, adding a guard is still preferable.

safl: Since the user must handle "spurious" events, an alternative approach would be to write to the eventfd before returning it.

safl (approved):
Great!

NaddiNadja (approved):
lgtm :)

PR #571, opened 2025-01-13
fix(enum): fix silent failure

karlowich:
`xnvme_enumerate()` would always return a success, even if enumeration failed.
Now only `err=-ENOSYS` is suppressed, and only if a backend is not specified.


jadoc:
Please also consider including karlowich/xnvme#1 for future test robustness.

karlowich:
Thanks @jadoc!

safl (approved):
lgtm

mbrsamsung (approved):
lgtm

PR #577, opened 2025-02-06
feat(cijoe/fio): add fio_fancy from cijoe

NaddiNadja:
The fio wrapper will be removed from cijoe in a later version, so we need remove this dependency.


safl (approved):
Excellent!

karlowich (approved):
lgtm

PR #580, opened 2025-02-10
feat(tests/fio): add `fio()` function from cijoe

NaddiNadja:
The fio wrapper will be completely removed from cijoe, so we need to import it here. This should probably have been part of #577


karlowich (approved):
Nice!

safl (approved):
Lgtm

PR #583, opened 2025-02-14
fix(cijoe/latency): move run argument to aux file

NaddiNadja:
In a newer version of cijoe, only primitive types or lists of primitive types will be allowed as values in the step.with dict. Therefore, we move the 'runs' argument to an auxiliary file in order to get ready for this.


safl (approved):
Great! Adding bench label, just in case :)

karlowich (approved):
lgtm!

PR #585, opened 2025-03-03
fix(python/bindings): revert to v2.3.4

NaddiNadja:
This is the version before the most recent version, which might have broken our CI. https://pypi.org/project/ctypeslib2/#history


karlowich (approved):

safl (approved):
Excellent, thanks!

PR #588, opened 2025-03-26
fix(cli): default log-health to controller

### LINKED ISSUE #587: Improve ``xnvme log-health``
By default, ``xnvme log-health`` uses the NSID of the device handle when the user does not provide one.  
As a result, it performs the optional command behavior of ``log-health-namespace`` instead of ``log-health-controller``.

It would be more appropriate to:

- Use ``0xFFFFFFFF`` as the default NSID when none is provided.

This ensures the mandatory command behavior is executed by default.

Baekalfen:
log-health of a namespace is optional to implement for the drive, so change the default to log-health of the controller, which is mandatory. This will improve usability, as the default is expected to succeed, and align differences in handling on Windows, Linux and macOS.

Fixes #587

### DIFF
--- a/tools/xnvme.c
+++ b/tools/xnvme.c
@@ -253,7 +253,8 @@ sub_log_health(struct xnvme_cli *cli)
 	int err;
 
 	if (!cli->given[XNVME_CLI_OPT_NSID]) {
-		nsid = xnvme_dev_get_nsid(cli->args.dev);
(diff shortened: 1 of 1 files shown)


safl (approved):
LGTM, thanks @Baekalfen

karlowich (approved):
lgtm

PR #590, opened 2025-03-27
feat(cijoe): bump to v0.9.52

NaddiNadja:
Firstly, this changes all instances of `cijoe.config.options.get(....)` to `cijoe.getconf(...)`, which is just a leftover that we've never gotten around to fix.

Secondly, the upgrade from cijoe v0.9.44 to v0.9.52 comes with a lot of refactorisation, which therefore makes the second commit quite large, and it would not be possible to do in smaller chunks, as these changes do not work individually.

- Remove `--workflow` argument, as this is no longer part of the cijoe cli. The workflow has been changed to a positional argument.
- Refactor config- and workflow files for new Qemu module setup.
- Add argparse arguments to scripts instead of step arguments and remove the step parameter

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -406,12 +406,10 @@ jobs:
     - name: CIJOE, install fio
       if: (!contains('focal', matrix.container.ver))
       run: |
-        cd cijoe && cijoe \
+        cd cijoe && cijoe fio_prep \
         --monitor \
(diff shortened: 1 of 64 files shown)


safl (dismissed):
Amazing work -- thank you!

karlowich (approved):
Great!

safl (approved):
Great!

PR #591, opened 2025-04-01
feat(tests/ioworker): add option for nlb 

karlowich:
`xnvme_tests_ioworker` could only do single-LBA I/O. This adds `--nlb` to
`verify` and `verify-sync` so the per-I/O size is configurable, and reworks the
SLBA bookkeeping to support it.

- **`--nlb`**: per-I/O geometry becomes `io.naddr = vec_cnt * (nlb + 1)`,
  `io.nbytes = io.naddr * lba_nbytes`. Defaults to `0`, i.e. previous
  behaviour. Each iovec segment now covers `nlb + 1` LBAs.

- **Drop the shared `work->cur` cursor**: replaced by `work->buf` plus
  per-command SLBA tracking via `ctx->cmd.nvm.slba`. Each async worker owns one
  context and a disjoint slice — worker `i` starts at `range.slba + i *
  io.naddr` and strides by `nworkers * io.naddr`. The sync path advances by
  `io.naddr` per iteration. Buffer offsets are now uniformly
  `buf + slba * lba_nbytes`.

- **Zero the full read buffer**: `rbuf` is diffed over `range.nbytes` but was
  only zero-filled for `io.nbytes`. Also fixes the `xnvme_buf_alloc()` debug
  labels.

- **cijoe**: every existing case also runs at `--nlb 1` and `--nlb 7`. For the
  iovec cases, small `nlb` keeps the `spdk` + `nosgl` failure expectation
  (sub-page segments need an SGL); large `nlb` is page-sized and expected to
  pass everywhere.

### DIFF
--- a/cijoe/tests/logic/test_xnvme_tests_ioworker.py
+++ b/cijoe/tests/logic/test_xnvme_tests_ioworker.py
@@ -8,10 +8,26 @@ def test_verify(cijoe, device, be_opts, cli_args):
     err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args}")
     assert not err
 
+    # small nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 1")
+    assert not err
+
+    # large nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 7")
+    assert not err
+
 
 @xnvme_parametrize(labels=["dev"], opts=["be", "sync", "admin"])
 def test_verify_sync(cijoe, device, be_opts, cli_args):
     err, _ = cijoe.run(f"xnvme_tests_ioworker verify-sync {cli_args}")
+    assert not err
+
+    # small nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 1")
+    assert not err
+
+    # large nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 7")
     assert not err
 
 
@@ -33,6 +49,18 @@ def test_verify_iovec(cijoe, device, be_opts, cli_args):
     else:
         assert not err
 
+    # small nlb: sub-page segments require an SGL, rejected on nosgl
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 1 --vec-cnt 4")
+
+    if be_opts["be"] == "spdk" and "nosgl" in device["labels"]:
+        assert err
+    else:
+        assert not err
+
+    # large nlb: page-sized segments are PRP-expressible
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 7 --vec-cnt 4")
+    assert not err
+
 
 @xnvme_parametrize(labels=["dev"], opts=["be", "sync", "admin"])
 def test_verify_sync_iovec(cijoe, device, be_opts, cli_args):
@@ -50,16 +78,52 @@ def test_verify_sync_iovec(cijoe, device, be_opts, cli_args):
     else:
         assert not err
 
+    # small nlb: sub-page segments require an SGL, rejected on nosgl
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 1 --vec-cnt 4"
+    )
+
+    if be_opts["be"] == "spdk" and "nosgl" in device["labels"]:
+        assert err
+    else:
+        assert not err
+
+    # large nlb: page-sized segments are PRP-expressible
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 7 --vec-cnt 4"
+    )
+    assert not err
+
 
 @xnvme_parametrize(labels=["bdev"], opts=["be", "sync", "async", "admin"])
 def test_verify_direct(cijoe, device, be_opts, cli_args):
     err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --direct 1")
     assert not err
 
+    # small nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 1 --direct 1")
+    assert not err
+
+    # large nlb
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify {cli_args} --nlb 7  --direct 1")
+    assert not err
+
 
 @xnvme_parametrize(labels=["bdev"], opts=["be", "sync", "admin"])
 def test_verify_sync_direct(cijoe, device, be_opts, cli_args):
     err, _ = cijoe.run(f"xnvme_tests_ioworker verify-sync {cli_args} --direct 1")
+    assert not err
+
+    # small nlb
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 1 --direct 1"
+    )
+    assert not err
+
+    # large nlb
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 7  --direct 1"
+    )
     assert not err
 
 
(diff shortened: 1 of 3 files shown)


safl:
Rebase this before running tests again, there is an issue with Windows causing the job to run for about 5 hours.

karlowich:
Found 2 issues:

1. **Async path not advancing slba**: `_submit()` never updates `work->cur.slba` after vector loop, while `_submit_sync` does at line ~187. This causes subsequent I/Os to use stale SLBA values leading to PRP misalignment errors. Fix: add `work->cur.slba += work->io.naddr;` after the vector loop.

2. **Buffer offset mismatch** (line ~144): Non-vectored path uses `work->geo->lba_nbytes` instead of `work->io.nbytes`. May cause incorrect offsets when nlb > 0.

---
*Review by Karlowich OpenCode - automated review, may contain errors*

tests/ioworker.c
```diff
@@ -112,10 +114,10 @@ _submit(struct iowork *work, struct ioworker *worker, struct xnvme_cmd_ctx *ctx)
 	int err;
 	if (work->vectored) {
 		for (int j = 0; j < work->vec_cnt; ++j) {
-			work->cur.slba = ctx->cmd.nvm.slba + j;
+			work->cur.slba = ctx->cmd.nvm.slba + j * (work->io.nlb + 1);
 			worker->vec[j].iov_base =
 				work->cur.data + work->cur.slba * work->geo->lba_nbytes;
-			worker->vec[j].iov_len = work->geo->lba_nbytes;
+			worker->vec[j].iov_len = (work->io.nlb + 1) * work->geo->lba_nbytes;
 		}
```
karlowich, line 121: Bug: work->cur.slba is not updated after this loop. The sync path (_submit_sync) correctly updates it at line 187 with work->cur.slba += work->io.nlb + 1, but this async path has no such update. This causes subsequent I/Os to use stale SLBA values, leading to PRP misalignment. Fix: add work->cur.slba += work->io.naddr; after this loop.

tests/ioworker.c
```diff
@@ -139,7 +141,7 @@ _submit(struct iowork *work, struct ioworker *worker, struct xnvme_cmd_ctx *ctx)
...
 	} else {
 		work->cur.slba = ctx->cmd.nvm.slba;
-		void *dbuf = work->cur.data + work->cur.slba * work->io.nbytes;
+		void *dbuf = work->cur.data + work->cur.slba * work->geo->lba_nbytes;
```
karlowich, line 144: Potential bug: Original code used work->io.nbytes here. Current code uses work->geo->lba_nbytes which may cause incorrect buffer offsets when nlb > 0.

naddinadja:
Can you put a bit more context into the last commit message?

karlowich:
@naddinadja done :)

tests/ioworker.c
```diff
@@ -298,9 +298,8 @@ iowork_from_cli(struct xnvme_cli *cli, struct iowork *work)
 		return -EINVAL;
 	}
 
-	work->io.nbytes = work->vec_cnt * work->geo->lba_nbytes;
-	work->io.naddr = work->vec_cnt;
-	work->io.nlb = work->io.naddr - 1;
+	work->io.nbytes = work->vec_cnt * (work->io.nlb + 1) * work->geo->lba_nbytes;
+	work->io.naddr = work->vec_cnt * (work->io.nlb + 1);
```
safl, line 300: Guard both failures right after this line:

```c
if (!work->io.naddr || work->range.naddr % work->io.naddr) {
    XNVME_DEBUG("FAILED: io.naddr must divide range.naddr");
    return -EINVAL;
}
```

Rejects `nio=0` (io.naddr > range.naddr) and non-divisible `--nlb` (tail-unwritten diff EIO). Also forecloses the `io.nbytes` overrun and 16-bit `nlb` truncation, since both need `io.naddr > range.naddr` first.

safl (changes requested):
One thing. `--nlb 2` on a 4K-LBA device leaves a tail unwritten and `xnvme_buf_diff` EIOs on a healthy drive; larger nlb/vec_cnt combos give `nio=0` and "run" zero I/O before failing the same compare. Same 3-line guard fixes both, inline. Everything else I looked at is latent, hygiene, or foreclosed by the guard.

tests/ioworker.c
karlowich: Great catch!
It's fixed now.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -50,18 +78,54 @@ def test_verify_sync_iovec(cijoe, device, be_opts, cli_args):
...
 
+    # small nlb: sub-page segments require an SGL, rejected on nosgl
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 1 --vec-cnt 4"
+    )
+
+    if be_opts["be"] == "spdk" and "nosgl" in device["labels"]:
+        assert err
+    else:
+        assert not err
```
naddinadja, line 89: A question for this: is there something inherent in our code that MUST fail if "nosgl" is in labels, or is it just that "this is not available and will therefore fail"? Because if it's the latter, then I think it would be better with a `pytest.skip()`

(there are 4 instances of this in this file, i only comment once)

naddinadja (commented):
A single question :)

cijoe/tests/logic/test_xnvme_tests_ioworker.py
karlowich: Yes, if the device doesn't support SGLs then we can't handle iovecs with segments smaller than the host page size. If the segments are matching the host page size, then we can use PRP lists instead of SGLs on these nosgl devices.

tests/ioworker.c
```diff
@@ -298,9 +298,13 @@ iowork_from_cli(struct xnvme_cli *cli, struct iowork *work)
...
+
+	if (!work->io.naddr || work->range.naddr % work->io.naddr) {
```
safl: This runs before `work->range.naddr` is assigned a few lines below, so it compares against 0 (the buffer is zeroed at function entry) and `0 % io.naddr == 0`, so the check is dead. `--nlb 2` on a 4K-LBA device (`io.naddr=3`, `range.naddr=4096`) passes here and then fails at `xnvme_buf_diff`, which is the case this guard was meant to catch. Move the block below the `work->range.naddr = ...` assignment.

safl (commented):
The divisibility guard added in this revision runs against a zeroed `range.naddr` so it never fires, see inline. The reworked completion-path SLBA tracking and the buffer/label cleanups otherwise read cleanly on this diff.

tests/ioworker.c
karlowich: Now it should be correct. The only thing changed was that block

naddinadja (dismissed):
lgtm :)

tests/ioworker.c
```diff
@@ -232,16 +229,18 @@ on_completion(struct xnvme_cmd_ctx *ctx, void *cb_arg)
 		goto error;
 	}
 
-	if (work->cur.slba >= work->range.elba) {
+	// Workers interleave, each striding by 'nworkers * io.naddr'; stop this worker when its
+	// next I/O would fall outside the range.
+	if (ctx->cmd.nvm.slba + (work->nworkers + 1) * work->io.naddr > work->range.naddr) {
```
safl: Uses `range.naddr` as if it were the exclusive end LBA; only correct because `range.slba` is hardcoded to 0. Using `range.slba + range.naddr` on the right-hand side keeps this robust if a subcommand ever starts the range mid-device.

tests/ioworker.c
```diff
@@ -298,23 +298,27 @@ iowork_from_cli(struct xnvme_cli *cli, struct iowork *work)
...
-	work->io.naddr = work->vec_cnt;
-	work->io.nlb = work->io.naddr - 1;
+	work->io.nbytes = work->vec_cnt * (work->io.nlb + 1) * work->geo->lba_nbytes;
+	work->io.naddr = work->vec_cnt * (work->io.nlb + 1);
 
 	// work->range.nbytes = 1 << 27;
 	work->range.nbytes = 1 << 24;
 	work->range.naddr = work->range.nbytes / work->geo->lba_nbytes;
 	work->range.slba = 0;
 	work->range.elba = work->range.slba + work->range.naddr - 1;
```
safl: After the SLBA-tracking rework nothing reads `range.elba` or `range.nlb` except the `iowork_pp` printout of `elba`; both assignments (and the fields themselves) could go.

safl (commented):
This is largely a cleanup pass, so two small optional items alongside if you want to fold them in.

tests/ioworker.c
karlowich: fixed

tests/ioworker.c
karlowich: fixed

safl (approved):
lgtm!

naddinadja (approved):
lgtm :)

PR #592, opened 2025-04-10
Fixes for CI

NaddiNadja:
These drives have been removed from the machine.

PR #593, opened 2025-04-25
fix(ci/gha): update jobs to run-on ubuntu-latest

NaddiNadja:
The ubuntu-20.04 image label has been deprecated, see: https://github.com/actions/runner-images/issues/11101


NaddiNadja:
I'm closing this because there is a circular dependency with #592 , so I've rebased that PR on this commit and hope that tests will go through.

PR #596, opened 2025-07-04
Experimental embedded NVMe Driver

safl:
This PR introduces a prototype of a minimal embedded NVMe driver in **xNVMe**. It enables efficient user-space operation without requiring **SPDK** or **libvfn**.

The motivation is to support NVMe devices in scenarios involving accelerator-initiated I/O, where **vfio/iommu** is unavailable and **libvfn** cannot be used. In such cases, embedding the driver logic offers a leaner alternative to the **DPDK** ``rte`` used by **SPDK**, making it more suitable for integration into accelerator runtimes.

This implementation includes:

- A minimal ``rte`` built with the header-only **uPCIe** library  
- DMA-capable memory backed by hugepages  
- MMIO operations via BAR space mapping  
- Basic NVMe operations: controller enablement, queue initialization, and command submission

The fundamental backend interface is in place. However, several enhancements are needed before progressing further. First, fix regressions:

- [x] Gracefully fallback when hugepages are not available, currently the enumeration tests fail bringing e.g. "xnvme enum" and the API backing it down as the **uPCIe** backend is always built
- [x] Only built on Linux, currently the **uPCIe** backend is not disabled when building on macOS which of course breaks the build on macOS among other things, then the UAPI and memfd is not available

Then, expand with improving payload handling. It is currently oversimplified -- it only supports **PRP1**, limiting I/O to a single page. Proper support should include:

- [x] **PRP lists**
- [ ] ~**SGLs**~
- [x] **Meta-data payloads**
- [x] **iovecs**

These features should ideally be added as helpers in **uPCIe**, but due to its design, they can also be conveniently implemented within the **xNVMe** repository.

Documentation and testing are currently absent and will be addressed in future work:

- [x] Expand documentation to describe usage and capabilities  
- [x] Add tests similar to those used for **libvfn** and **SPDK**

The pseudo-commands for reset are dropped similar to **libvfn**. Instead, the user is expected to just close and re-open the device.


safl:
The first three patches are now up on its own PR #605

karlowich (commented):
Overall it looks good.
I don't agree with putting the xnvme code in `lib/upcie/` this breaks our previous convention. Of course we can make a new one, but it should at least be consistent.
We are missing an `_rte_term()` function in `xnvme_be_upcie_dev.c`, I will add it.

karlowich:
In the MaaS setup, we have found that `vfio-pci` causes `enum` to hang forever with `upcie`.
With `uio_pci_generic` it hangs a while and then fails on "Resource temporarily unavailable" when trying to setup the `sync` queue.

karlowich:
The `rte_term` function I have introduced doesn't actually successfully free the hugepages, it needs more work.

NaddiNadja:
I've been playing around with this on our CI machine, and it looks like the tests when running inside a QEMU instance where a physical NVMe device has been passed-through with vfio, and inside the QEMU instance, the device is bound to the vfio-pci driver. I've run `xnvme enum` in gdb, and found specifically, it hangs in file `mmio.h`  in function `mmio_read32()` on line 39, when trying to dereference the pointer to the bar. The call go:
```C
// xnvme_be_upcie_dev_open() -> nvme_controller_open() -> nvme_mmio_cap_read() -> mmio_read64() -> mmio_read32()

static inline uint32_t
mmio_read32(void *region, uint32_t offset)
{
	return *(volatile uint32_t *)((uintptr_t)region + offset);
}
```
the `region` pointer is just the pointer retrieved from `mmap`'ing the bar in file `pci.h` in function `pci_bar_map()`:
```C
bar->region = mmap(NULL, bar->size, PROT_READ | PROT_WRITE, MAP_SHARED, bar->fd, 0x0);
```
So it looks like the QEMU instance doesn't have access to the memory at the given address. I've tried a bunch of different settings for running QEMU, but I couldn't really figure out any that would give me a different result. Running `xnvme enum` on the host with these changes, it does not hang forever (but fails because `nvme_mmio_csts_wait_until_not_ready()` from file `nvme_mmio.h` fails after the 2 minute timeout).

BUT - this is with the vfio-pci driver. Changing to the uio_pci_generic driver inside the guest, it still fails, but does not hang infinitely. It fails because the call to function `nvme_qpair_reap_cpl()` in file `nvme_qpair.h` does not succeed. The same goes for running it on the host, but I have not had time to look more into that yet. 

However, I will save this for the new year and go home now for the holidays! Merry christmas and have a happy new year 🎄

safl:
Thanks for looking into this @NaddiNadja , happy holidays!

lib/upcie/xnvme_be_upcie_admin.c
```diff
@@ -0,0 +1,45 @@
...
+	struct nvme_controller *ctrlr = state->ctrlr;
+	struct nvme_command *cmd = (struct nvme_command *)&ctx->cmd;
+	struct nvme_completion *cpl = (struct nvme_completion *)&ctx->cpl;
+	int err;
+
+	err = nvme_qpair_submit_sync_contig_prps(&ctrlr->aq, &g_upcie_rte.heap, dbuf, dbuf_nbytes,
+						 cmd, ctrlr->timeout_ms,
+						 (struct nvme_completion *)&ctx->cpl);
+
+	err = nvme_qpair_submit_sync(&ctrlr->aq, cmd, ctrlr->timeout_ms, cpl);
```
karlowich: The admin command is submitted twice here.

karlowich (changes requested):

lib/upcie/xnvme_be_upcie_admin.c
karlowich: This has been fixed

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	return 0;
+}
+
+void
+xnvme_be_upcie_dev_close(struct xnvme_dev *dev)
+{
+	(void)dev;
+}
```
NaddiNadja, line 206: We don't want to do anything here? It looks like in other backends that we at least do
```C
memset(&dev->be, 0, sizeof(dev->be));
```

lib/upcie/xnvme_be_upcie_async.c
```diff
@@ -155,13 +155,66 @@ xnvme_be_upcie_async_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_n
...
+	req = nvme_request_alloc(upcie_queue->qpair.rpool);
+	if (!req) {
+		XNVME_DEBUG("FAILED: nvme_request_alloc(); errno(%d)", errno);
+		return -errno;
+	}
+
+	req->user = ctx;
+	cmd->cid = req->cid;
+
+	nvme_request_prep_command_prps_iov(req, &g_upcie_rte.heap, dvec, dvec_cnt, cmd);
```
karlowich: I think this call should be guarded behind an if-block checking if `dbuf` is `NULL`

lib/upcie/xnvme_be_upcie_sync.c
```diff
@@ -52,13 +52,51 @@ xnvme_be_upcie_sync_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nb
...
+		ctx->cmd.nvm.slba = ctx->cmd.nvm.slba >> ctx->dev->geo.ssw;
+		ctx->cmd.common.opcode = XNVME_SPEC_NVM_OPC_WRITE;
+		break;
+	}
+
+	if (mbuf) {
+		ctx->cmd.common.mptr = hostmem_dma_v2p(&g_upcie_rte.heap, mbuf);
+	}
+
+	err = nvme_qpair_submit_sync_iov_prps(&state->ctrlr->sync, &g_upcie_rte.heap, dvec,
```
karlowich: I think this call should be guarded behind an if-block checking if `dbuf` is `NULL`

safl:
@NaddiNadja @karlowich putting changes to uPCIe on https://github.com/safl/upcie/pull/8 once they hit uPCIe ill tag it and replace the patches here with a general update of uPCIe to that tag.

docs/background/backends/upcie/index.md
```diff
@@ -0,0 +1,50 @@
...
+reactor, threading model, or application framework — just direct PCIe BAR
+access and hugepage-based DMA.
+
+(sec-backends-upcie-identifiers)=
+
+## Device Identifiers
+
+When using user-space NVMe-drivers, the operating-system kernel NVMe driver is
+detached and the device bound to `uio_pci_generic`. Thus, the device-files in
+`/dev/`, such as `/dev/nvme0n1` are not available. Devices are instead
```
NaddiNadja: No hyphen between "NVMe" and "drivers", and I would also say "device files" without the hyphen. 

I wanted to also do something about "user-space", but I see that we use both "userspace", "user-space" and "user space" in our docs right now, so maybe streamlining this is work for a future commit.

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -68,6 +68,57 @@ _rte_init(void)
...
+ *
+ * Reads the PCI Command register via sysfs config space and sets the Bus Master Enable bit,
+ * unless it is already set, then it exits early.
+ *
+ * @param bdf PCI Bus-Device-Function string, e.g. "0000:01:00.0"
+ *
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
+ */
+static int
+_pci_enable_bus_master(const char *bdf)
```
karlowich, line 82: Is this a function that makes sense to upstream in uPCIe?

karlowich (changes requested):
Very nice!
Can you also upstream the changes added to support IOV to uPCIe?

Few comments inline

lib/upcie/xnvme_be_upcie_dev.c
safl: I am a bit on the fence on this one... e.g. the **uPCIe** ``devbind`` tool does this ahead of time, thus, no need for a **uPCIe** user to do it themselves. Then, **xNVMe** have not adopted ``devbind`` and have not added BME to the the ``xnvme-driver`` tool. Thus, either this work-around thus stays until these other matters are addressed, currently leaning towards migrating the ``xnvme-driver`` tool to use the **uPCIe** tools ``devbind`` and ``hugepagge`` in one way or the other, however, I find this is out of scope of this PR and thus this relatively simple thing inside of the **xNVMe** / **uPCIe** backend implementation.

lib/upcie/xnvme_be_upcie_sync.c
safl: There is no ``dbuf`` here only ``dvec`` and ``mbuf``.

lib/upcie/xnvme_be_upcie_async.c
safl: There is no ``dbuf`` here only ``dvec`` and ``mbuf``.

NaddiNadja (changes requested):
A single thing in the docs, and then a question for the dev_close :)

lib/upcie/xnvme_be_upcie_async.c
karlowich: Right, I meant `dvec`. It is not safe to call this function if `dvec == NULL`

lib/upcie/xnvme_be_upcie_sync.c
karlowich: Right, I meant `dvec`. It is not safe to call this function if `dvec == NULL`

lib/upcie/xnvme_be_upcie_dev.c
karlowich: Okay, I see. Then let's keep it as is

lib/upcie/xnvme_be_upcie_dev.c
safl: I think this should be handled by the ctrlr-init/term now and not require anything else in ``dev_open()``/``dev_close()``, but let me verify that...

safl:
@karlowich @NaddiNadja more stuffs: https://github.com/safl/upcie/pull/9 :)

lib/upcie/xnvme_be_upcie_dev.c
safl: Yeah, init/term does all the heavy lifting. However, i did change to make use of ``XNVME_UNUSED``.

docs/background/backends/upcie/index.md
safl: I gave it a shot, but only on the docs touched in this PR.

lib/upcie/xnvme_be_upcie_sync.c
safl: As i recall then xNVMe would consider it a caller-error since the cmd_iov paths are only calls when an iov-payload is explicitly sought after by the caller, thus it is not prone to the same issue as dbuf is on the non-iovec path as it can receive a dbuf which is null for things like write-zeroes. So yeah... no guard against user-error here, the user just should not call iov with ``dvec=NULL``, then they ought to call the regular path  with ``dbuf=NULL``.

lib/upcie/xnvme_be_upcie_async.c
safl: See comment on the other instance of this.

lib/upcie/xnvme_be_upcie_sync.c
karlowich: SPDK *only* sends IOV commands to the bdev, so our `bdev_xnvme` would also send write zeroes (or similar) through the IOV path. (link: [BDEV xNVMe Submission](https://github.com/spdk/spdk/blob/3529e073c80337406643bac16d7ad552839fa5f8/module/bdev/xnvme/bdev_xnvme.c#L235))

lib/upcie/xnvme_be_upcie_async.c
```diff
@@ -8,18 +8,160 @@
...
+	req = nvme_request_alloc(upcie_queue->qpair.rpool);
+	if (!req) {
+		XNVME_DEBUG("FAILED: nvme_request_alloc(); errno(%d)", errno);
+		return -errno;
+	}
+
+	req->user = ctx;
+	cmd->cid = req->cid;
+
+	nvme_request_prep_command_prps_contig(req, &g_upcie_rte.heap, dbuf, dbuf_nbytes, cmd);
```
karlowich: We also need to protect against `dbuf == NULL` here

karlowich (changes requested):

lib/upcie/xnvme_be_upcie_sync.c
safl: That is true. Hmm.

safl: Having said that, then I don't think any of the user-space driver backends support passing NULL for ``dvec``, perhaps the SPDK one does if SPDK itself handles it. Let me check.

safl: Yeah... adding guards.

lib/upcie/xnvme_be_upcie_async.c
safl: Guards added and squashed.

lib/upcie/xnvme_be_upcie_sync.c
safl: Guards added and squashed.

lib/upcie/xnvme_be_upcie_async.c
safl: Guards added and squashed.

karlowich (approved):
Great!

NaddiNadja (approved):
looks goodie to me! :)

PR #597, opened 2025-07-17
fix(queue): misleading debug message

karlowich:


safl (approved):
LGTM.

Sigh... seems like FreeBSD runners are failing for unrelated reasons.

PR #598, opened 2025-08-04
CI fixes

NaddiNadja:
### FreeBSD provisioning breaking: switch to libuuid
From e2fsprogs-libuuid to libuuid. It seems the e2fsprogs version has been removed as a port. See commit in the freebsd-ports repository: https://github.com/freebsd/freebsd-ports/commit/e81eda36a8fe2b02133674d52cd27f38a3b65d8b
Error "call to undeclared function 'uuid_generate_sha1'" in SPDK build, possibly also due to the above change. See this issue: https://github.com/spdk/spdk/issues/3696 and the patch to fix it

### perflatwin provisioning: Use latest fio commit
This is the same as my previous commit for the GHA jobs, it just needed to be applied for provisioning perflatwin too: https://github.com/xnvme/xnvme/commit/610a6b74d0973c911d410705b4fea4a61d4d077f

### DIFF
--- a/cijoe/configs/perf-lat-windows.toml
+++ b/cijoe/configs/perf-lat-windows.toml
@@ -29,7 +29,8 @@ bin = "C:/Users/root/git/fio/fio.exe"
 [fio.repository]
 remote = "https://github.com/axboe/fio.git"
 path = "C:/Users/root/git/fio"
-tag = "fio-3.38"
+# tag = "fio-3.38"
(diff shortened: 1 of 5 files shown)


karlowich (approved):
LGTM

safl (approved):
Thanks for getting this back on track!
I have made a couple of notes for the changes that we should address before the next release:

https://github.com/xnvme/xnvme/issues/599

PR #601, opened 2025-08-15
Refactoring GitHUB Action / Workflow

safl:
This refactor eliminates duplicated logic in ``verify.yml``, improving maintainability. It sets the stage for the main change in this PR: switching the latency benchmark on Linux and FreeBSD from bare-metal in the lab to a MaaS provider.

As noted in the commit message, this change affects the absolute latency numbers. However, the purpose and usage of the benchmark remain the same: evaluating relative latency overhead. For reference then old and new reports are provided below.

Reports on MaaS AX42
--------------------

- Linux Latency Report [latency-linux.pdf](https://github.com/user-attachments/files/22005366/latency-linux.pdf)

- FreeBSD 14 Latency Report [latency-freebsd.pdf](https://github.com/user-attachments/files/22005363/latency-freebsd.pdf)

Reports on Lab N6005
--------------------

- Linux Latency Report [report-latency-linux.pdf](https://github.com/user-attachments/files/22005101/report-latency-linux.pdf)

- FreeBSD 13 Latency Report [report-latency-freebsd13.pdf](https://github.com/user-attachments/files/22005100/report-latency-freebsd13.pdf)


cijoe/scripts/fio_latency.py
```diff
@@ -409,7 +409,9 @@ def main(args, cijoe: Cijoe):
 
                 log.info(f"Executing the command: {cmd}")
 
-                err, state = cijoe.run(cmd, env=fio.env())
+                # err, state = cijoe.run(cmd, env=fio.env())
```
karlowich: Do we need to keep this line?

karlowich (dismissed):
Looks good.
Very minor comment.

cijoe/scripts/fio_latency.py
safl: Fixed thanks!

karlowich (approved):

cijoe/configs/ax42-freebsd.toml
```diff
@@ -0,0 +1,144 @@
...
+#
+# Used by: cijoe.run() / cijoe.get() / cijoe.put()
+[cijoe.transport.qemu_guest]
+hostname = "localhost"
+port = 4200
+username = "root"
+password = ""
+
+[cijoe.run]
+shell = "csh"
```
NaddiNadja, line 12: Change to "sh" if root shell is no longer csh on FreeBSD 14? :)

NaddiNadja (changes requested):
Only one thing - but this *might* also fix the env vars not being injected properly by cijoe in the `fio_latency` script

cijoe/configs/ax42-freebsd.toml
safl: Other stuff breaks in that case with the fio build, some assumption on something...

NaddiNadja: ah well .... 😅

NaddiNadja (approved):
lgtm :)

PR #605, opened 2025-09-01
Misc. changes in prep. for embedded NVMe driver

safl:
These are taken directly from #596 to get reviewed and integrated in preparation for #596.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -64,6 +64,7 @@ tests = {
   'znd_explicit_open.c': [],
   'znd_state.c': [],
   'znd_zrwa.c': [],
+  'async_limbo.c': [],
 }
 
 if not is_windows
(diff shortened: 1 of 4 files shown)


tests/meson.build
```diff
@@ -64,6 +64,7 @@ tests = {
   'znd_explicit_open.c': [],
   'znd_state.c': [],
   'znd_zrwa.c': [],
+  'async_limbo.c': [],
```
karlowich, line 67: This test looks ideal for ramdisk testing, so let's add it to the tests run by `meson test`.

karlowich (changes requested):
In general good, one comment

tests/meson.build
safl: It might not be a great fit since it runs until it gets a signal, so, it would run forever in e.g. CI.

karlowich: Ah, I missed that

karlowich: Let's have the text for the cli tool state that it runs forever

NaddiNadja (approved):
lgtm :)

karlowich (approved):
Great

PR #606, opened 2025-09-02
Run windows latency benchmark on MaaS

NaddiNadja:


NaddiNadja:
As source for comparison between the latency on the new CI setup (MaaS) and the old CI setup (physical machines):
- Here is the latency report from the benchmark in this PR: [latency-maas.pdf](https://github.com/user-attachments/files/22132516/latency-maas.pdf)
- Here is the report from the latest release, v0.7.5: [latency-v0.7.5.pdf](https://github.com/user-attachments/files/22132626/latency-v0.7.5.pdf)
- Here is the report from the latest commit to branch next, [`118f706`](https://github.com/xnvme/xnvme/commit/118f706cc76af3a2d4d696e7e95803d83cdfae1f): [latency-next.pdf](https://github.com/user-attachments/files/22132635/latency-next.pdf)

safl (dismissed):
This is excellent! 🎉
It looks like in the new environment, both IOCP and IOCP-thread scale worse than before. This is most likely related to the number of cores assigned to the guest and how they’re pinned. IOCP-thread is also showing a slightly different pattern compared to earlier runs. That said, we can revisit these performance aspects later. For now, the key achievement is that everything runs successfully in the new setting. Thanks!

safl:
@karlowich any feedback on this?

safl (approved):
Excellent!

PR #609, opened 2025-09-04
feat(docs): guide for creating a windows qcow2 image

naddinadja:


docs/contributing/ci/runners/windows-vm/index.rst
```diff
@@ -0,0 +1,330 @@
...
+  and run again.
+    
+.. code-block:: console
+
+  C:\Users\root\Downloads\windows-2022.ps1
+    
+
+OpenSSH:
+
+- Download `OpenSSH-Win64.zip`_.
```
safl: Can this be done by enabling the component instead? As described here:

https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=gui&pivots=windows-11#install-openssh-server--client

docs/contributing/ci/runners/windows-vm/index.rst
```diff
@@ -0,0 +1,330 @@
...
+
+  mkdir -p /home/ghr/guests/tmp
+  rm -rf /home/ghr/guests/tmp/*
+  cp /var/lib/libvirt/images/<qcow2 image> /home/ghr/guests/tmp/boot.img
+
+- Start a swtpm socket
+    
+.. code-block:: bash
+
+  swtpm socket --tpm2 --ctrl type=unixio,path=/home/ghr/guests/tmp/swtpm.sock --tpmstate dir=/home/ghr/guests/tmp --flags not-need-init --daemon
```
safl, line 290: When the host is rebooted  must this be started manually again?

docs/contributing/ci/runners/windows-vm/index.rst
NaddiNadja: No the problem is that OpenSSH is not in the list of components (for some reason ...), so I had to download it and install it first before being able to follow that guide to enable it

To clarify: In the guide, it says to run `Get-WindowsCapability -Online | Where-Object Name -like 'OpenSSH*'`, and then it should return some output of components that are not enabled/present. But running it on this image, nothing showed up. After installing this zip, it showed up as "NotPresent".

I could add a sentence about checking whether it's there, before beginning to download it all :)

docs/contributing/ci/runners/windows-vm/index.rst
NaddiNadja: Yes, this is a step that must be done before running the VM - which is also why it's under the "Test with QEMU" header. I've had some problems with not being able to connect to the socket after closing down qemu, so in our CI, it's part of the `windows_qemu_prep.py` script

docs/contributing/ci/runners/windows-vm/index.rst
safl: Oh, that is surprising, ok, great then!

docs/contributing/ci/runners/windows-vm/index.rst
safl: Ah, I see, so although it is daemon, then it is sort of "one-off". Tying it into the guest-start-script makes sense then. Thanks

PR #610, opened 2025-09-05
feat(gha/benchmark): run benchmark-bdev_xnvme on MaaS

NaddiNadja:


NaddiNadja:
Reports for comparison:

- [scalability-maas.pdf](https://github.com/user-attachments/files/22170119/scalability-maas.pdf)
- [scalability-next.pdf](https://github.com/user-attachments/files/22170120/scalability-next.pdf)
- [scalability-v0.7.5.pdf](https://github.com/user-attachments/files/22170121/scalability-v0.7.5.pdf)

Note that since the latest release, we have made an update, which fixes the colouring of each plot to be the for each run, such that comparison between reports is easier. But the report from v0.7.5 is older than this, and therefore the colours are different.

safl (approved):
Excellent!

PR #617, opened 2025-09-23
feat(macos/driverkit): update for block storage support

Baekalfen:
MacVFN has been updated with block storage support, which changes the IOService to match.

### DIFF
--- a/lib/xnvme_be_macos_driverkit_dev.c
+++ b/lib/xnvme_be_macos_driverkit_dev.c
@@ -20,7 +20,7 @@ xnvme_be_macos_driverkit_enumerate(const char *sys_uri, struct xnvme_opts *opts,
 				   xnvme_enumerate_cb cb_func, void *cb_args)
 {
 	XNVME_DEBUG("INFO: xnvme_be_macos_driverkit_enumerate()");
-	CFMutableDictionaryRef matching_dict = IOServiceMatching("IOUserService");
+	CFMutableDictionaryRef matching_dict;
 	struct xnvme_opts tmp_opts = *opts;
 	io_iterator_t iterator;
 	kern_return_t ret;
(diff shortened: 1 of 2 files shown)


karlowich (approved):
Looks good.
Currently we don't have the physical MacOS CI, so we can't run our tests

safl:
We expect to get CI back online with the mac-HW in a month or so to test this.
Meanwhile, is a new tag of the macvfn needed for this? Anything to update on the docs?

Baekalfen:
>  is a new tag of the macvfn needed for this?

What kind of tag?

> Anything to update on the docs?

Were the docs moved? I cannot find them. But anyways, all user-facing changes are on the MacVFN side.

safl:
> > is a new tag of the macvfn needed for this?
> 
> What kind of tag?

xNVMe CI currently uses MacVFN at [xnvme_v0.9.6](https://github.com/SamsungDS/MacVFN/releases/tag/xnvme_v0.9.6), is this the tag to use for the changes on this PR?

> 
> > Anything to update on the docs?
> 
> Were the docs moved? I cannot find them. But anyways, all user-facing changes are on the MacVFN side.

Still here:

- https://xnvme.io/en/next/backends/xnvme_be_macos.html
- https://github.com/xnvme/xnvme/blob/next/docs/backends/xnvme_be_macos.rst

Anything to add / update there?

Baekalfen:
> xNVMe CI currently uses MacVFN at [xnvme_v0.9.6](https://github.com/SamsungDS/MacVFN/releases/tag/xnvme_v0.9.6), is this the tag to use for the changes on this PR?

Ah, that tag. I was wondering about that tag. Was there supposed to be an xNVMe version v0.9.6? It's a number I got from you at the last PR.

But yes, I will tag a version of xNVMe when I know which one has the patches.

> * https://xnvme.io/en/next/backends/xnvme_be_macos.html
> * https://github.com/xnvme/xnvme/blob/next/docs/backends/xnvme_be_macos.rst
> 
> Anything to add / update there?

No, that looks fine. The only change is that it's also registered for block storage.

safl:
> > xNVMe CI currently uses MacVFN at [xnvme_v0.9.6](https://github.com/SamsungDS/MacVFN/releases/tag/xnvme_v0.9.6), is this the tag to use for the changes on this PR?
> 
> Ah, that tag. I was wondering about that tag. Was there supposed to be an xNVMe version v0.9.6? It's a number I got from you at the last PR.
> 
> But yes, I will tag a version of xNVMe when I know which one has the patches.

We’re facing a chicken-and-egg problem: we don’t pull in untested changes, and there’s no stable reference point for the new functionality in MacVFN. How about adding semantic versioning tags to MacVFN instead of xNVMe-specific tags? That way, we can test with something like v0.1 of MacVFN. If it works, we update the xNVMe documentation to say “tested with v0.1 of MacVFN.” Users can then install that version, knowing that xNVMe next was tested against it. When the next xNVMe release is published, this reference can be promoted to the release documentation, making it clear that, for example, xNVMe v0.9.6 was tested with MacVFN v0.1.

> 
> > * https://xnvme.io/en/next/backends/xnvme_be_macos.html
> > * https://github.com/xnvme/xnvme/blob/next/docs/backends/xnvme_be_macos.rst
> > 
> > Anything to add / update there?
> 
> No, that looks fine. The only change is that it's also registered for block storage.

Sounds good.

Baekalfen:
> we don’t pull in untested changes, and there’s no stable reference point for the new functionality in MacVFN.

Nothing in the RPC API has changed. It's just that MacVFN now registers as a block storage device. And that means the enumeration (and not open/close/etc) was looking for the wrong type of service. This enumeration only exists in xNVMe. And I made it backwards compatible. You can test this PR with an old MacVFN if you want to.

safl:
> Nothing in the RPC API has changed. It's just that MacVFN now registers as a block storage device. And that means the enumeration (and not open/close/etc) was looking for the wrong type of service. This enumeration only exists in xNVMe. And I made it backwards compatible. You can test this PR with an old MacVFN if you want to.

Ack. No change needed to the MacVFN tag used by the xNVMe CI then.

That said, I suggest adopting semantic versioning for MacVFN. Among other benefits, it would allow testing against the specific version this PR targets, presumably the preferred or latest, instead of relying on backward compatibility.

safl (approved):
LGTM.

PR #623, opened 2026-01-27
Platform-managed backends with cref and config selection

safl:
Replace the mixin-based backend system with platform-managed backend configs and centralized controller reference counting (cref).

  - **Platform architecture**: One `xnvme_platform_<os>.c` per OS, compile-time selected, owning flat lists of `xnvme_be_config` with caps-based auto-selection from URI
  - **Config composability**: Each config defines a complete I/O recipe (async+sync+admin+dev+mem) with optional `mem_overrides` for runtime memory driver swapping (e.g. `--mem hugepage` on Linux configs)
  - **cref orchestration**: `ctrl_init`/`ctrl_term` hooks on `xnvme_be_dev` let the platform manage controller lifecycle — userspace drivers (vfio, driverkit) share controllers across namespace opens automatically
  - **`xnvme_scan()`**: New device discovery API with `kernel_driver` field on `xnvme_ident`, enumerate reimplemented on top of scan + dev_open
  - **Unique config names and driver IDs**: Every backend config has a unique name; `library-info` prints all configs with driver IDs
  - **Test rework**: Skip logic uses driver names and OS checks; enum test validates unique config names; library-info test verifies all backend combos

  ### Note on SPDK

  SPDK has **not** been transitioned to platform-managed cref yet. It still calls `xnvme_be_cref_*()` directly. The SPDK cref transition will follow in a separate PR once the current changes are validated.

  ### Commit highlights

  | Area | What changed |
  |------|-------------|
  | Platform | `xnvme_platform` introduced, one per OS, replaces backend registry |
  | Configs | Mixin system → flat config lists with caps, descr, unique names, mem_overrides |
  | cref | Extracted shared module; platform orchestrates via `ctrl_init`/`ctrl_term` |
  | vfio | Hoisted cref to platform, fixed multi-namespace |
  | driverkit | Hoisted cref to platform |
  | scan | `xnvme_scan()` on Linux/FreeBSD/Windows/macOS; `kernel_driver` on ident |
  | tests | Rewritten skip logic, enum validation, library-info coverage |
  | docs | Background docs converted to MyST; architecture design doc added |
  | misc | Vendor uPCIe v0.3.2, rename vfio→libvfn, spdk interface ids |
  
  ### Regressions and other TODOs

- [x] NVMe is gone? Add a test verifying that the amount of enumerated namespaces match those in the configuration  
- [x] The SPDK backend is not converted to the platform-managed cref-logic, also those cref-helpers are ports of the old cref-logic, there is rich opportunity for simplifications once SPDK is fixed
  - Check whether this is true by running the test-multi
- [x] Several ZNS/Zoned tests are skipped which they probably should not be
- [x] Linux namespace reporting gave the Linux enumeration index and not the actual nsid. This is fixed now by reading ``nsid`` from sysfs during scan.


lib/xnvme_be_cref.c
```diff
@@ -17,23 +17,26 @@ struct xnvme_be_cref_entry {
 static struct xnvme_be_cref_entry g_cref_table[XNVME_BE_CREF_MAX_ENTRIES];
 
 static int
-_entry_matches(const struct xnvme_be_cref_entry *entry, const char *uri)
+_entry_matches(const struct xnvme_be_cref_entry *entry, const char *uri, const char *be_name)
 {
 	if (strncmp(entry->uri, uri, XNVME_IDENT_URI_LEN)) {
 		return 0;
 	}
+	if (be_name && entry->be_name != be_name) {
```
NaddiNadja: can you compare strings like that?

NaddiNadja (commented):
I've done a look-through this PR now - a lot of great things! I'm not the most knowledgeable on the intricacies of xNVMe, but I have a single thing inline for now. Also: the last commit describes that it has the "backend_refactor.md" file in the docs, but this was included already in commit [`f9ed72a`](https://github.com/xnvme/xnvme/commit/f9ed72ae6a3d743a1ace1ad50751d5604f1b640c). I was at least confused as to why it was part of that commit, but I assume it was a mistake?

lib/xnvme_be_cref.c
safl: No definitely not. However, the trick here is that all callers pass ``cfg->attr.name``, thus, what is compared are not the string-representation aka string-comparison of the backend-configurations, but rather the pointer-value. This works because the pointers have stable addresses as they point to string-literals. In other words, they are identifying the backend-configuration by the pointer-address rather than by the content of what it is pointing to.

Having said that... then it is a bit of a loaded foot-gun... at some point one will probably call the function with a stack-allocated or dynamically build backend name and then it won't match... so... yeah... let me just go ahead and replace this with ``strcmp``.

safl:
> I've done a look-through this PR now - a lot of great things! I'm not the most knowledgeable on the intricacies of xNVMe, but I have a single thing inline for now. Also: the last commit describes that it has the "backend_refactor.md" file in the docs, but this was included already in commit [`f9ed72a`](https://github.com/xnvme/xnvme/commit/f9ed72ae6a3d743a1ace1ad50751d5604f1b640c). I was at least confused as to why it was part of that commit, but I assume it was a mistake?

That was a mistake, nice catch! I hoisted it as it has been integrated into the background/backend docs.

lib/xnvme_be_cref.c
safl: Fixed now. Also, fixed some missing XNVME_DEBUG on the error-path.

safl: Also reimplemented the hugepage allocator using uPCIe instead of the "handrolled" one.

NaddiNadja: I was suspecting that it was something like that, that was going on, hence my questionmark 😅 but great!

lib/xnvme_be_vfio_admin.c
```diff
@@ -31,7 +31,7 @@ xnvme_be_vfio_sync_cmd_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_
 #endif
 
 struct xnvme_be_admin g_xnvme_be_vfio_admin = {
-	.id = "nvme",
+	.id = "libvfn",
 #ifdef XNVME_BE_VFIO_ENABLED
```
karlowich, line 35: Should this macro also be renamed?

include/xnvme_be_cref.h
```diff
@@ -0,0 +1,92 @@
...
+ * @param uri Device URI to match or insert
+ * @param be_name Backend name pointer, stored for filtering in xnvme_be_cref_cleanup()
+ * @param ctrlr Controller handle to insert if no match is found, may be NULL for lookup-only
+ * @param destructor Callback invoked to destroy the controller when refcount reaches zero
+ *
+ * @return On success, the controller handle. On error, NULL.
+ */
+void *
+xnvme_be_cref_ref(const char *uri, const char *be_name, void *ctrlr,
+		  xnvme_be_cref_destructor_fn destructor);
```
karlowich, line 63: Do we need this function? I'm trying to wrap my head around this API, and it seems that it is intended to be used in the `open` path like this:
* Call `xnvme_be_cref_lookup()`, if it exists we are done
* If `NULL` open the ctrlr and call `xnvme_be_cref_insert()`
This how it is used in the libvfn backend.

In the SPDK backend we never call `xnvme_be_cref_lookup()` and instead use `xnvme_be_cref_ref()` which means that we don't get any protection against opening the controller multiple times.

lib/xnvme_be_vfio.c
karlowich, line 1: Should we also rename the files to `libvfn` instead of `vfio`?

include/xnvme_platform.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
+ */
+int
+xnvme_platform_dev_open(struct xnvme_dev *dev, struct xnvme_opts *opts);
+
+/**
+ * Shared enumerate implementation - iterates platform->backends, calls each driver's enumerate
+ *
+ * @param sys_uri System URI filter (NULL for all devices)
```
karlowich: Is this the best name for this parameter? To me it is unclear how "System URI" is different from "URI". Perhaps we can drop the `sys` prefix. 

Alternatively, if this is really meant as a filter we could name it as such. Is the intention that one could call it with `sys_uri = "ng"` and only get the char devices?

meson.build
```diff
@@ -167,6 +167,9 @@ conf = configure_file(
   output : 'xnvme_config.h',
 )
 conf_inc = [ include_directories('.') ]
+if is_linux
+  conf_inc += [ include_directories('toolbox' / 'third-party' / 'linux') ]
```
karlowich, line 185: Should this be `toolbox/third-party/linux/upcie/` instead? Just in case we add other things to `toolbox/third-party/linux`?

lib/xnvme_platform_freebsd.c
```diff
@@ -7,13 +7,145 @@
...
 {
-	return -ENOSYS;
+	struct xnvme_fbsd_scan_args args = {0};
+	struct pci_conf_io pc = {0};
+	struct pci_conf conf[32];
+	int pci_fd;
+	int err = 0;
+
+	if (sys_uri) {
+		XNVME_DEBUG("FAILED: sys_uri: %s is not supported", sys_uri);
```
karlowich, line 186: Why this check? To me it looks like the `sys_uri` is never used...
--- Also
What will happen to the debug print format when the string is NULL? Because it only gets printed when sys_uri is NULL.
Perhaps the message could be adjusted to something like:
"FAILED: sys_uri == NULL is not supported'

lib/xnvme_platform_windows.c
```diff
@@ -8,12 +8,93 @@
...
+	HDEVINFO disk_class_devices;
+	GUID disk_class_device_interface_guid = GUID_DEVINTERFACE_DISK;
+	SP_DEVICE_INTERFACE_DATA device_interface_data;
+	DWORD device_index = 0;
+	int num_devices = 0;
+
+	if (sys_uri) {
+		XNVME_DEBUG("FAILED: sys_uri: %s is not supported", sys_uri);
+		return -ENOSYS;
+	}
```
karlowich, line 55: Same as for the FreeBSD, is `sys_uri` used for anything?

lib/xnvme_platform_linux.c
```diff
@@ -2,10 +2,167 @@
...
+static int
+xnvme_platform_linux_scan(const char *sys_uri, struct xnvme_opts *XNVME_UNUSED(opts),
+			  xnvme_scan_cb cb_func, void *cb_args)
+{
+	struct xnvme_linux_scan_args args = {0};
+
+	if (sys_uri) {
+		XNVME_DEBUG("FAILED: sys_uri: %s is not supported", sys_uri);
+		return -ENOSYS;
+	}
```
karlowich, line 329: Same as for the FreeBSD is `sys_uri` used for anything?

include/xnvme_platform.h
karlowich: I see we use `sys_uri` all over the place, so perhaps it is irrelevant

NaddiNadja (dismissed):
This is great, I don't any more comments for now! :)

cijoe/tests/tools/test_xnvme.py
```diff
@@ -27,49 +27,69 @@ def test_enum_fabrics(cijoe, device, be_opts, cli_args):
...
+    Verify that xnvme_scan() discovers devices.
 
+    On Linux/FreeBSD: discovers controllers (dtype: 0x1) and tests both
+    kernel-attached and kernel-detached states.
+    On Windows/macOS: discovers namespaces (dtype: 0x2) only in attached state.
+    """
     from ..conftest import cijoe_config_get_all_devices
 
-    # Get expected controller count from config
+    osname = pytest.cijoe_instance.getconf("os.name", "linux")
```
karlowich: Perhaps it is better to fail if the config is malformed instead of assuming Linux?

lib/xnvme_platform_macos.c
```diff
@@ -8,12 +8,129 @@
...
-xnvme_platform_nosys_scan(const char *XNVME_UNUSED(sys_uri), struct xnvme_opts *XNVME_UNUSED(opts),
-			  xnvme_scan_cb XNVME_UNUSED(cb_func), void *XNVME_UNUSED(cb_args))
+xnvme_platform_macos_scan(const char *sys_uri, struct xnvme_opts *XNVME_UNUSED(opts),
+			  xnvme_scan_cb cb_func, void *cb_args)
 {
-	return -ENOSYS;
+	if (sys_uri) {
+		XNVME_DEBUG("FAILED: sys_uri: %s is not supported", sys_uri);
+		return -ENOSYS;
+	}
```
karlowich, line 63: Again, is `sys_uri` used for anything?

lib/xnvme_be_spdk_dev.c
```diff
@@ -374,19 +374,22 @@ attach_cb(void *cb_ctx, const struct spdk_nvme_transport_id *trid, struct spdk_n
 
 	XNVME_DEBUG("INFO: nsid: %d", opts->nsid);
```
karlowich, line 376: Perhaps this should be moved into the if block, so we only print the nsid if there is one?

tests/ctrlr.c
```diff
@@ -0,0 +1,203 @@
...
+	const struct xnvme_ident *ident;
+
+	dev = ctrlr_open(cli);
+	if (!dev) {
+		return -errno;
+	}
+
+	ident = xnvme_dev_get_ident(dev);
+	if (ident->dtype != XNVME_DEV_TYPE_NVME_CONTROLLER) {
+		xnvme_cli_pinf("FAILED: expected dtype CONTROLLER, got %d", ident->dtype);
```
karlowich: Perhaps we can replace "CONTROLLER" with the numeric representation making it easier to compare to the `ident->dtype`

tests/meson.build
```diff
@@ -28,6 +28,7 @@ tests = {
     ['no args', ['copy', '1GB']],
     ['slba=0 sdlba=8', ['copy', '1GB', '--slba', '0', '--nlb', '7', '--sdlba', '8']],
     ['slba=8 sdlba=0', ['copy', '1GB', '--slba', '8', '--nlb', '7', '--sdlba', '0']]],
+  'ctrlr.c': [],
```
karlowich, line 31: If this works with the ramdisk, you can add the parameters here

lib/xnvme_be_spdk_dev.c
```diff
@@ -567,12 +567,10 @@ xnvme_be_spdk_enumerate(const char *sys_uri, struct xnvme_opts *opts, xnvme_enum
 			XNVME_DEBUG("FAILED: ident_from_uri, uri: %s", sys_uri);
 			return -EINVAL;
 		}
-		if (sscanf(ident.uri, "%[^:]:%d", addr, &port) != 2) {
-			XNVME_DEBUG("FAILED: invalid sys_uri: %s", sys_uri);
-			return -EINVAL;
-		}
+		sscanf(ident.uri, "%[^:]:%d", addr, &port);
 	}
```
karlowich: Why are we no longer checking if the sys_uri is malformed?

lib/xnvme_platform.c
```diff
@@ -160,67 +162,150 @@ xnvme_platform_dev_open(struct xnvme_dev *dev, struct xnvme_opts *opts)
...
 
-	for (int i = 0; g_xnvme_platform->backends[i]; ++i) {
-		struct xnvme_be be = *g_xnvme_platform->backends[i];
+	idfy_buf = xnvme_buf_alloc(ctrl_dev, sizeof(*idfy_buf));
+	if (!idfy_buf) {
+		XNVME_DEBUG("FAILED: xnvme_buf_alloc()");
+		xnvme_dev_close(ctrl_dev);
+		return -ENOMEM;
+	}
+	memset(idfy_buf, 0, sizeof(*idfy_buf));
```
karlowich: Use `xnvme_buf_clear` this way we can ensure compatibility with other memory types introduced in the future.

lib/xnvme_be_fbsd_dev.c
```diff
@@ -198,7 +160,7 @@ xnvme_be_fbsd_dev_open(struct xnvme_dev *dev)
 
 struct xnvme_be_dev g_xnvme_be_fbsd_dev = {
 #ifdef XNVME_BE_FBSD_ENABLED
-	.enumerate = xnvme_be_fbsd_enumerate,
+	.enumerate = xnvme_be_nosys_enumerate,
```
karlowich, line 163: I think we can just remove this field from the struct instead of setting it to `nosys` everywhere

lib/xnvme_be.c
```diff
@@ -68,13 +68,24 @@ xnvme_be_registry_fpr(FILE *stream, enum xnvme_pr XNVME_UNUSED(opts))
 
 	wrtn += fprintf(stream, "xnvme_platform:\n");
 	wrtn += fprintf(stream, "  name: '%s'\n", g_xnvme_platform->name);
-	wrtn += fprintf(stream, "  backends:\n");
+	wrtn += fprintf(stream, "  drivers:\n");
```
karlowich: They are still called "backends", right?

lib/xnvme_be_fbsd.c
```diff
@@ -5,107 +5,63 @@
...
-		.async = &g_xnvme_be_cbi_async_nil,
-		.check_support = xnvme_be_supported,
-	},
+const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_psync = {
+	.async = &g_xnvme_be_fbsd_async,
+	.sync = &g_xnvme_be_cbi_sync_psync,
+	.admin = &g_xnvme_be_fbsd_admin_nvme,
+	.dev = &g_xnvme_be_fbsd_dev,
+	.mem = &g_xnvme_be_cbi_mem_posix,
+	.attr = {.name = "fbsd", .descr = "kqueue with POSIX sync", .enabled = 1},
```
karlowich: Shouldn't the name be unique? I.e. `fbsd_kqueue_psync`?

Is there a reason why we need to specify `enabled = 1`?
Looks to me like this is controlled by macros now.

lib/xnvme_platform.c
```diff
@@ -152,12 +95,12 @@ xnvme_platform_dev_open(struct xnvme_dev *dev, struct xnvme_opts *opts)
 			return err;
 
 		default:
-			XNVME_DEBUG("INFO: skipping backend due to err: %d", err);
+			XNVME_DEBUG("INFO: skipping config due to err: %d", err);
```
karlowich: The naming is a little muddy. It varies between config, driver, and backend. Maybe we can choose just one?

include/xnvme_be.h
```diff
@@ -104,6 +104,20 @@ struct xnvme_be_dev {
 	void (*dev_close)(struct xnvme_dev *);
 
 	const char *id;
+
+	/**
+	 * Initialize the controller for a userspace driver.
+	 * Returns an opaque controller handle, or NULL on failure.
+	 * NULL pointer means the backend does not use platform-managed cref.
+	 */
+	void *(*ctrl_init)(struct xnvme_dev *);
```
karlowich: We abbreviate controller like `ctrlr` almost everywhere else. I prefer this, do you agree?

lib/xnvme_platform.c
```diff
@@ -92,17 +93,59 @@ xnvme_platform_dev_open(struct xnvme_dev *dev, struct xnvme_opts *opts)
...
+				ctrlr = xnvme_be_cref_lookup(dev->ident.uri, cfg->attr.name);
+				if (!ctrlr) {
+					void *conflict =
+						xnvme_be_cref_lookup(dev->ident.uri, NULL);
+					if (conflict) {
+						xnvme_be_cref_deref(conflict, XNVME_BE_CREF_NONE);
+						XNVME_DEBUG("FAILED: uri '%s' claimed by "
+							    "another driver",
+							    dev->ident.uri);
+						err = -EBUSY;
```
karlowich: Will this error ever be useful? Or is it another case like the "enumeration error", i.e., an error is returned but not necessarily the one you care about.

lib/xnvme_platform.c
```diff
@@ -92,17 +93,59 @@ xnvme_platform_dev_open(struct xnvme_dev *dev, struct xnvme_opts *opts)
...
+						goto next_config;
+					}
+					ctrlr = be.dev.ctrl_init(dev);
+					if (!ctrlr) {
+						XNVME_DEBUG("INFO: ctrl_init failed");
+						err = -errno ? -errno : -EIO;
```
karlowich: Same here, will this error have an effect?

cijoe/tests/apps/test_fio.py
```diff
@@ -11,7 +11,7 @@ def test_fio_engine(cijoe, device, be_opts, cli_args):
     """
     The construction of the fio-invocation is done in 'fio_fancy'
     """
-    is_windows = be_opts["mem"] == "windows" or be_opts["be"] == "windows"
+    is_windows = be_opts["mem"] == "windows"
```
karlowich: I think we should use `get_osname()` here instead

cijoe/tests/apps/test_fio.py
```diff
@@ -45,14 +45,14 @@ def test_fio_engine_iov(cijoe, device, be_opts, cli_args):
...
-
-    is_windows = be_opts["mem"] == "windows" or be_opts["be"] == "windows"
+    if be_opts["sync"] == "psync":
+        pytest.skip(reason="[sync=psync] does not implement iovec")
+    if get_osname() == "freebsd" and be_opts["sync"] == "nvme":
+        pytest.skip(reason="[sync=nvme] on FreeBSD does not implement iovec")
+    if be_opts["admin"] == "driverkit":
+        pytest.skip(reason="[admin=driverkit] does not implement iovec")
+
+    is_windows = be_opts["mem"] == "windows"
```
karlowich: Again, I think we should use `get_osname()` here

karlowich: There are two other identical places in the file too.

cijoe/tests/examples/test_zoned_io_async.py
```diff
@@ -1,33 +1,33 @@
...
 
-from ..conftest import xnvme_parametrize
+from ..conftest import get_osname, xnvme_parametrize
 
 
 @xnvme_parametrize(labels=["zns"], opts=["be", "admin", "async"])
 def test_write(cijoe, device, be_opts, cli_args):
-    if be_opts["be"] == "fbsd":
+    if get_osname() == "freebsd":
         pytest.skip(reason="Freebsd kernel doesn't support zns")
```
karlowich: I think this check is too broad. One could still use SPDK for this on FreeBSD, right?

karlowich: Same goes other places in this file and in `test_zoned_io_sync.py`

cijoe/tests/logic/test_xnvme_tests_enum.py
```diff
@@ -5,10 +5,12 @@
...
 def test_open(cijoe, device, be_opts, cli_args):
-    if be_opts["be"] in ["ramdisk"]:
+    if be_opts["admin"] == "ramdisk":
         pytest.skip(reason=f"[be={be_opts['be']}] does not support enumeration")
-    if be_opts["be"] == "driverkit":
-        pytest.skip(reason="[be=driverkit] does not support repeatedly opening devices")
+    if be_opts["admin"] == "driverkit":
+        pytest.skip(
+            reason="[admin=driverkit] does not support repeatedly opening devices"
+        )
```
karlowich: Is this fixed with the enumeration fixes?

cijoe/tests/logic/test_xnvme_tests_enum.py
```diff
@@ -35,9 +37,9 @@ def test_keep_open(cijoe, device, be_opts, cli_args):
 
 @xnvme_parametrize(labels=["dev"], opts=["be"])
 def test_multi(cijoe, device, be_opts, cli_args):
-    if be_opts["be"] in ["ramdisk"]:
+    if be_opts["admin"] == "ramdisk":
         pytest.skip(reason=f"[be={be_opts['be']}] does not support enumeration")
-    if be_opts["be"] in ["driverkit", "spdk"]:
+    if be_opts["admin"] in ["driverkit", "spdk"]:
         pytest.skip(
             reason=f"[be={be_opts['be']}] does not support repeatedly enumerating"
```
karlowich: Is this fixed with the enumeration fixes?

cijoe/tests/examples/test_zoned_io_async.py
karlowich: And `test_zoned.py`

karlowich: And the other zoned files

lib/xnvme_platform_freebsd.c
```diff
@@ -197,7 +197,7 @@ xnvme_platform_fbsd_scan(const char *sys_uri, struct xnvme_opts *XNVME_UNUSED(op
 }
 
 struct xnvme_platform g_xnvme_platform_freebsd = {
-	.name = "freebsd",
+	.name = "fbsd",
```
karlowich: I would much prefer the opposite. Moving away from `fbsd` to `freebsd` everywhere else.

lib/libxnvme.map
```diff
@@ -23,6 +23,9 @@
 		xnvme_be_registry_fpr;
 		xnvme_be_registry_pr;
 
+		# xnvme_platform.h
+		g_xnvme_platform;
```
karlowich: I would prefer to have a getter function instead. Exporting a global variable is a little rough.

tests/enum.c
```diff
@@ -252,18 +252,21 @@ test_enum_backend(struct xnvme_cli *cli)
 {
 	struct xnvme_opts opts = xnvme_opts_default();
 	struct backend_cb_args cb_args = {0};
-	char *backends[N_BACKENDS] = {"spdk",    "linux",   "fbsd",  "macos",
-				      "windows", "ramdisk", "libvfn"};
 	int err;
 
-	for (int i = 0; i < N_BACKENDS; ++i) {
+	for (int i = 0; g_xnvme_platform->backends[i]; ++i) {
```
karlowich: Perhaps a getter function could give you all available backends instead?

cijoe/workflows/test-debian-bookworm.yaml
```diff
@@ -6,8 +6,8 @@ steps:
 - name: sysinfo
   uses: linux.sysinfo
 
-- name: hugetlbfs
-  uses: linux_setup_hugetlbfs
```
karlowich, line 10: Should we complete kill the old script?

lib/xnvme_be_linux_hugepage.c
```diff
@@ -7,167 +7,137 @@
...
-	char path[PATH_MAX];
-	int fd;
-	void *addr;
-	size_t len;
+#include <upcie/debug.h>
+#include <upcie/hostmem.h>
+#include <upcie/hostmem_config.h>
+#include <upcie/hostmem_hugepage.h>
+#include <upcie/hostmem_heap.h>
+#include <upcie/hostmem_dma.h>
```
karlowich, line 28: Perhaps we could just include `upcie/upcie.h` ?
Then we don't have to add all of its includes as well

lib/xnvme_be_linux_hugepage.c
```diff
@@ -7,167 +7,137 @@
...
-	char line[PATH_MAX];
-	char search_str[PATH_MAX];
-	FILE *fp;
+	env_val = getenv("XNVME_HUGETLB_PATH");
+	if (env_val) {
+		strncpy(g_hugepage_config.hugetlb_path, env_val,
+			sizeof(g_hugepage_config.hugetlb_path) - 1);
+		g_hugepage_config.hugetlb_path[sizeof(g_hugepage_config.hugetlb_path) - 1] = '\0';
+		g_hugepage_config.backend = HOSTMEM_BACKEND_HUGETLBFS;
+	}
```
karlowich, line 66: Some of this code is almost identical to what is already in `hostmem_config_init` is there any reason we can't just piggyback on what is already there?
E.g., do we really need our own environment variables?

lib/xnvme_be_linux_hugepage.c
```diff
@@ -7,167 +7,137 @@
...
-		perror("hugetlb ftruncate()");
-		return NULL;
-	}
-
-	buf = mmap(NULL, nbytes, PROT_READ | PROT_WRITE, MAP_SHARED, hugepage_fd, 0);
-	if (buf == MAP_FAILED) {
-		XNVME_DEBUG("mmap failed on hugepage file");
-		return NULL;
+	if (phys) {
+		err = hostmem_heap_block_virt_to_phys(&g_hugepage_heap, buf, phys);
```
karlowich, line 115: **NOT A BLOCKER**

You are mixing APIs here. I understand this is necessary to get the error code. This is fuel for my argument here:
https://github.com/safl/upcie/issues/4

karlowich (changes requested):
Overall really nice. A million comments inline.

For the future, let's try to break huge PRs into smaller ones. There's so many unrelated things going here that it could've easily been 4 separate PRs.

lib/xnvme_be.c
karlowich: I see you fixed this in a later commit

lib/xnvme_be_fbsd.c
karlowich: I see you fixed the first part in a later commit. But the point about `enabled` still stands

lib/xnvme_platform_freebsd.c
NaddiNadja: Isn't it the other way around? It checks that `sys_uri != NULL` (for the print formatting part)

safl:
> Overall really nice. A million comments inline.
> 
> For the future, let's try to break huge PRs into smaller ones. There's so many unrelated things going here that it could've easily been 4 separate PRs.

Right, I agree that's why i started https://github.com/xnvme/xnvme/pull/635 with uPCIe included such that things depending on it could have gone in separate PRs and possibly in parallel, i recall making a note of that there as well.

lib/xnvme_platform_freebsd.c
karlowich: You are right

lib/xnvme_be_vfio.c
safl: Yes, making a note of that.

lib/xnvme_platform_freebsd.c
karlowich: In that case, should it really fail if you give an unused argument?
Perhaps we can just mark it unused with `XNVME_UNUSED`

include/xnvme_platform.h
safl: It might not be the best, however, it is what the codebase is currently using.

lib/xnvme_be_vfio_admin.c
safl: Yeah. Making a note of that.

include/xnvme_be_cref.h
safl: The SPDK backend uses ``xnvme_be_cref_ref()`` so yes it is needed.

meson.build
safl: I don't follow. If other things need to be added on Linux, then they can simply be added and one does not need to change meson. If the path is set to upcie only, then one has to change meson for every thing added on linux?

lib/xnvme_platform_freebsd.c
safl: Yes, we need to inform that what the user tried to do is not supported (in this case on FreeBSD).

lib/xnvme_platform_windows.c
safl: See response on Freebsd.

lib/xnvme_platform_linux.c
safl: See response on Freebsd.

cijoe/tests/tools/test_xnvme.py
safl: Yeah, you have point there. Making a note of that.

lib/xnvme_platform_macos.c
safl: See response on Freebsd.

lib/xnvme_be_spdk_dev.c
safl: Maybe, however, not the scope of this PR.

include/xnvme_be_cref.h
karlowich: Right. What I was getting at is, the SPDK backend has problems as a result of `xnvme_be_cref_ref()`. This function is a trap. SPDK still opens the device multiple times, and then uses `xnvme_be_cref_ref()` to increment the ref-count even though it effectively makes a new ctrlr each time.

include/xnvme_platform.h
karlowich: Yes, no need to change that. You're right.

tests/ctrlr.c
safl: In the print statement? as in:
```
xnvme_cli_pinf("FAILED: expected dtype CONTROLLER(%d), got %d", XNVME_DEV_TYPE_NVME_CONTROLLER, ident->dtype);
```
?

meson.build
karlowich: That assumes that every project added is a header only library. If we were to add something like SPDK, we would want to consume the headers from where they're installed and only the public headers. This way we include every single header file in all sub-directories, I think that scope is too wide.

tests/ctrlr.c
karlowich: Exactly

cijoe/workflows/test-debian-bookworm.yaml
safl: I considered that, however, since uPCIe supports memfd as well as hugetlbfs, then it is kept as it will come in handy when tests are added for both paths. Now, the priority is simply to get the easier-to-use path in and adopt the upcie API.

tests/enum.c
safl: What do you mean?

lib/xnvme_be_linux_hugepage.c
safl: But it only needs these, so no reason to add stuff that is no needed.

karlowich:
> uPCIe included such that things depending on it could have gone in separate PRs and possibly in parallel, i recall making a note of that there as well.

Yes this was the right call

lib/xnvme_be_linux_hugepage.c
safl: I think so. That is why added them.

lib/xnvme_be_linux_hugepage.c
karlowich: Okay

cijoe/workflows/test-debian-bookworm.yaml
karlowich: Ah, I see

lib/xnvme_be_linux_hugepage.c
karlowich: Okay, I guess it is simpler that xNVMe users doesn't need to know anything about uPCIe. But for the record there is no difference between `XNVME_HUGEPAGE_BACKEND` and the uPCIe `HOSTMEM_BACKEND`. Same goes for `XNVME_HUGETLB_PATH` and the uPCIe `HOSTMEM_HUGETLB_PATH`.

tests/enum.c
karlowich: Well instead of exporting a global variable as a symbol in our public API. Perhaps we could create a function in the xNVMe API getting the backends instead or if we need other things it can get the platform.

include/xnvme_be_cref.h
safl: I am not sure I follow. Are you saying this PR introduces new issues that are not covered by tests? Or are you referring to the known enumeration issues in the SPDK backend, or something else?

In general, this PR is not about fixing the SPDK backend. It is about refactoring and improving the codebase in preparation for the uPCIe backend. Part of that work is hoisting the cref logic so we do not have to replicate it yet again. Otherwise, we would end up with nearly identical implementations in libvfn, uPCIe, driverkit, and SPDK.

With these changes, the platform manages cref for libvfn, uPCIe, and driverkit. The SPDK backend is not fully converted, but it does make use of the new helpers. As stated in the PR description, converting and further improving the SPDK backend is out of scope for this PR.

meson.build
safl: I still do not follow. The change simply extends the include path for Linux to add vendored headers, matching what is already done for macOS (`toolbox/third-party/macos`) and Windows (`toolbox/third-party/windows`).

tests/enum.c
safl: I don’t follow. No global variable is added to the public API in this PR. The code you’re pointing at is internal, not part of the exported interface. I believe there’s a misunderstanding in how it’s being read.

tests/meson.build
safl: I does not, ramdisk has dtype (``XNVME_DEV_TYPE_RAMDISK``), the test is for dtype(``XNVME_DEV_TYPE_NVME_CONTROLLER``).

lib/xnvme_be_vfio_admin.c
safl: I’d prefer to defer these renames and naming alignment. They are largely cosmetic, and in several cases things were not aligned to begin with. For example, the libvfn admin interface was previously named `nvme` and is now named `libvfn`. I’ve collected the related cleanup items here: [https://github.com/xnvme/xnvme/issues/637](https://github.com/xnvme/xnvme/issues/637)

lib/xnvme_be_vfio.c
safl: I’d prefer to defer these renames and naming alignment. They are largely cosmetic, and in several cases things were not aligned to begin with. For example, the libvfn admin interface was previously named `nvme` and is now named `libvfn`. I’ve collected the related cleanup items here: [https://github.com/xnvme/xnvme/issues/637](https://github.com/xnvme/xnvme/issues/637)

lib/xnvme_be_fbsd_dev.c
safl: It is not nosys everywhere.

lib/xnvme_platform.c
safl: Now there are "xnvme_be" and "xnvme_be_config" in this case backend configs are iterated so using 'config' applies.

lib/xnvme_platform.c
safl: I believe so. E.g. open a bdf with spdk, then open the same with libvfn. With this it will tell you that it is busy since another driver is using it.

tests/enum.c
safl: Ohhh wait a minute...

lib/xnvme_be_vfio.c
karlowich: Sounds good

meson.build
karlowich: Yes, sure. It's not important. If we eventually add something to `toolbox/third-party/linux` that shouldn't be included 1:1, we can change it.

tests/meson.build
karlowich: Got it.

lib/xnvme_platform.c
karlowich: Okay, great

tests/enum.c
karlowich: You are adding `g_xnvme_platform` to `libxnvme.map` in the same commit, thus making it part of the exported symbols. I'm proposing that we instead create a function returning what we need.

lib/xnvme_be_vfio_admin.c
karlowich: Sounds good

tests/enum.c
safl: Good catch, you are absolutely right... hmm... adding a helper here. I had also overlooked some old remnants in ``libxnvme_be.h`` for the now removed "be_registry".

include/xnvme_be_cref.h
karlowich: I'm saying that the bug in SPDK is caused by `xnvme_be_cref_ref()` (or the equivalent function from before).
So, I propose not carrying on the faulty function. Since, if it is not present, then it is unlikely someone will fall into the trap of using it. However, it is not part of the public API and can be changed later if that is preferred.

lib/xnvme_be_fbsd_dev.c
karlowich: Your refactor makes it so the platform is responsible for the enumeration. Thus, I believe there's is no reason that the `enumerate` field should be part of the `xnvme_be_dev` struct anymore. I don't think there's any way to call this function through the public API after your changes.

lib/libxnvme.map
safl: agree

lib/xnvme_be_fbsd_dev.c
safl: Hmm so... thanks for this. There is something I have missed here... i though there still was one left that served the fabrics path such that enumerate would be the discovery in the context of TCP or RDMA. But I agree, it seems like it is now deadcode... this needs fixing...

include/xnvme_be.h
safl: yup, good catch

lib/xnvme_platform.c
safl: Ah good point!

include/xnvme_be_cref.h
safl: So.. I tried not messing with the SPDK backend, however, it does look like I managed to introduce a regression such that device-discovery for NVMe/TCP|RDMA is now br0k3n. Thus... it does motivate taking a harder look and also fixing the SPDK backend... sigh...

cijoe/tests/tools/test_xnvme.py
safl: Addressed in ``refactor(tests): use get_osname() and add per-platform ramdisk configs``

tests/ctrlr.c
safl: Now addressed in ``fix(tests): show expected dtype value in ctrlr_open failure message``

lib/xnvme_be_fbsd_dev.c
safl: Yeah... this is a general regression... needs some more thought...

lib/xnvme_be_fbsd.c
safl: Right. The ``enabled`` does not make a lot of sense any longer now that they are simply not included when not ``enabled``. So it could be removed.

include/xnvme_be.h
safl: addressed in ```refactor(be): rename ctrl_init/ctrl_term to ctrlr_init/ctrlr_term```

tests/enum.c
safl: Now addressed in ```refactor(be): rename registry API to xnvme_be_attr and add getter```

lib/libxnvme.map
safl: Now addressed in ```refactor(be): rename registry API to xnvme_be_attr and add getter```

lib/xnvme_platform.c
safl: Addressed in ``refactor(platform,dev): replace memset with xnvme_buf_clear for DMA buffers``

lib/xnvme_be_fbsd.c
safl: Addressed in: ```refactor(be): remove unused 'enabled' field from xnvme_be_attr```

cijoe/tests/apps/test_fio.py
safl: Addressed in ``refactor(tests): use get_osname() for fio output path``

cijoe/tests/apps/test_fio.py
safl: Addressed in ``refactor(tests): use get_osname() for fio output path``

lib/xnvme_platform_freebsd.c
safl: Addressed in ``refactor(be,platform): rename from 'fbsd' to 'freebsd'`` (not pushed yet)

cijoe/tests/examples/test_zoned_io_async.py
safl: Addressed on ``fix(tests): scope FreeBSD ZNS skips to kernel backends only``

docs/background/backends/common/index.md
```diff
@@ -0,0 +1,50 @@
+(sec-backends-common)=
+
+# Common Backend Impl (CBI)
+
+The common backend implementations are implementations for re-use by other
+backends. For example, instead of the Linux, FreeBSD, and macOS backends all
+implement the same POSIX asynchronous I/O interface, then it is instead provided
+here, and the can all include it in their support for async. interfaces.
```
NaddiNadja: Would you say "Linux, FreeBSD, and macOS _platforms_" instead of backends?
I would remove the "then" and fix spelling error in "then it is instead provided here, and they* can all"

docs/background/backends/common/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+Similarly, then **xNVMe** provide several software-only implementations of the
+**xNVMe** API, such as the **thrpool** which implements the asynchronous parts
+of the **xNVMe** API by using a thread-pool.
```
NaddiNadja: Again, I would remove "then" in "Similarly, then xNVMe ..."

docs/background/backends/common/nil.md
```diff
@@ -0,0 +1,7 @@
+(sec-backends-common-async-nil)=
+
+# Async. I/O via `nil`
+
+The `nil` backend is entirely for debugging and measuring the IO-layer, all
+the `nil` async. implementation does is queue up commands and when polled for
+completion they are returned with success.
```
NaddiNadja: I would split this into two sentences, maybe:
"The purpose of the `nil` backend is entirely for debugging and measuring the IO-layer. The `nil` asynchronous implementation only queues up commands and when polled for completion, commands are returned with success."
(I assume "they" refer to the commands in the original text)

docs/background/backends/index.md
```diff
@@ -0,0 +1,74 @@
...
+: The schema for identifying devices differ across platforms and
+  interfaces. Varying from file device handles, e.g. `/dev/nvme0n1`, to
+  PCI device handles, e.g. `0000:02:00.0`, and NVMe/TCP endpoints, e.g.
+  `172.10.10.10:4420`.
+
+Backend Instrumentation
+: On a given platform multiple interface may be available. At runtime,
+  the **xNVMe** library makes a decision on which interface to use. You
+  can overrule this decision, however, to do so, you need to know what the
+  available options are, what they are named, and what they offer.
```
NaddiNadja: Comma after "On a given"
This is just a preference, but I would probably change the "however, to do so", to "but to do so", such that there is not so many commas in a row :)

docs/background/platforms/freebsd/index.md
```diff
@@ -0,0 +1,75 @@
+(sec-platforms-freebsd)=
+
+# FreeBSD
+
+(sec-platforms-freebsd-configuration)=
+
+## System Configuration
+
+To use the FreeBSD backend you need to be running FreeBSD. xNVMe is likely to
```
NaddiNadja: "the FreeBSD platform", right?

docs/background/platforms/freebsd/index.md
```diff
@@ -0,0 +1,75 @@
...
+
+For details see the man-pages for `nvme`, `nda`, `nvd`, `pci`, `disk`,
+and `nvmecontrol`.
+
+(sec-platforms-freebsd-instrumentation)=
+
+## Backend Instrumentation
+
+The backend identifier for the FreeBSD backend is: `fbsd`. The FreeBSD backend
+encapsulates the interfaces below. Some of these are **mix-ins** from the
```
NaddiNadja: Platforms instead of backends, right?

docs/background/platforms/linux/io_uring.md
```diff
@@ -0,0 +1,26 @@
+(sec-platforms-linux-async-io_uring)=
+
+# Async. I/O via `io_uring`
+
+**xNVMe** support **io_uring** it does so by mapping the NVMe commands read
+and write from the NVM command-set to **io_uring** opcodes:
```
NaddiNadja: " **xNVMe** supports **io_uring** by mapping ... "
Also, do you want quotation marks or italisized "read" and "write" just to highlight that these are the command names?
I see in file "docs/background/platforms/windows/iocp.md", we use bold font for **read** and **write**.

docs/background/platforms/macos/index.md
```diff
@@ -0,0 +1,54 @@
+(sec-platforms-macos)=
+
+# macOS
+
+(sec-platforms-macos-configuration)=
+
+## System Configuration
+
+To use the macOS backend you need to be running macOS. xNVMe is likely to
```
NaddiNadja: "macOS platform", right?

docs/background/platforms/macos/index.md
```diff
@@ -0,0 +1,54 @@
...
+
+`/dev/disk{ctrlr_id}`
+: **NVMe** Controller device node
+
+(sec-platforms-macos-instrumentation)=
+
+## Instrumentation
+
+The backend identifier for the macOS backend is: `macos`. The macOS backend
+encapsulates the interfaces below. Some of these are **mix-ins** from the
```
NaddiNadja: "platform" instead of "backend", right?

docs/background/platforms/windows/index.md
```diff
@@ -0,0 +1,70 @@
+(sec-platforms-windows)=
+
+# Windows
+
+The Windows backend communicates with the Windows NVMe driver StorNVMe.sys via
+IOCTLs. The backend relies on the NVMe pass-through through IOCTLs to submit
```
NaddiNadja: "Windows platform" instead of "backend"

docs/background/platforms/windows/index.md
```diff
@@ -0,0 +1,70 @@
...
+admin, IO and arbitrary user-defined commands.
+
+## System Configuration
+
+To use the Windows backend you need to be running Windows. xNVMe is likely
+to work on multiple different versions of Windows, however Windows 10, or
```
NaddiNadja: "Windows platform" instead of "backend"

docs/background/platforms/windows/index.md
```diff
@@ -0,0 +1,70 @@
...
+Windows exposes NVMe devices as files using the following naming schemes:
+
+`\\.\PhysicalDrive{ctrlr_id}`
+: **NVMe** Controller device node
+
+
+## Instrumentation
+
+The backend identifier for the Windows backend is: `windows`. The Windows backend
+encapsulates the interfaces below. Some of these are **mix-ins** from the
```
NaddiNadja: platform instead of backend, right?

docs/background/platforms/windows/iocp_th.md
```diff
@@ -0,0 +1,19 @@
+(sec-platforms-windows-async-iocp_th)=
+
+# Async I/O via `iocp_th`
+
+Similar to `iocp` interface, only difference is separate poller is used to
+fetch the completed IOs.
```
NaddiNadja: I suggest:
"The `iocp_th` interface is similar to `iocp`, the only difference being that a separate poller is used to fetch the completed IOs."

docs/background/platforms/index.md
```diff
@@ -0,0 +1,18 @@
+(sec-platforms)=
+
+# Platforms
+
+xNVMe supports one platform per operating system. Each platform registers its
```
NaddiNadja: Bold font xNVMe? :)

docs/background/index.md
```diff
@@ -0,0 +1,22 @@
+(sec-background)=
+
+# Background
+
+**xNVMe** abstracts platform-specific NVMe interfaces behind a common API.
+Each supported operating system provides a **platform** implementation that
+registers a set of **backend configs** — combinations of implementations for
+admin, sync, async, memory, and device operations. At runtime, xNVMe selects a config
+whose **caps** match the target device, or you can override the selection via
+URI options.
```
NaddiNadja: I think "whose" is only really used for people. Maybe go with:
"At runtine, **xNVMe** selectes a config with **caps** matching the target device. If you wish to override the backend configuration, you can do so via URI options."

NaddiNadja (changes requested):
I hadn't read the docs before now. I have a couple things inline only for that one commit :)

docs/background/platforms/index.md
safl: Did a pass and ensure this all over the docs. (not pushed).

docs/background/index.md
safl: Gave it a pass based on your feedback, did (not pushed):

  Bolding

  - All prose mentions of xNVMe now formatted as xNVMe across 26 doc files
  - HTML contexts (material/index.rst) use <b>xNVMe</b>
  - Correctly excluded: URLs, alt text, inline code, RST cross-references

  Terminology (platform vs backend)

  - Linux, FreeBSD, macOS, Windows are now described as platforms, not backends
  - "Backend Instrumentation" sections renamed to "Backend Configs"
  - Backends index page rewritten around backend config selection via caps matching
  - "mix-ins" → "cross-platform implementations" (CBI) in FreeBSD, macOS, Windows docs
  - "drivers" → "backend configs" where not referring to actual kernel/hardware drivers
  - macOS DriverKit: "backend" → "backend config"

  Stale references fixed

  - Backend interface attributes: removed enabled, schemes, nschemes; added descr, caps
  - FreeBSD identifier: fbsd → freebsd
  - Linux example output: updated to current format (dev/mem lines, attr with descr/caps, be='emu_bdev')

  Phrasing improvements

  - Platform descriptions open with "X is one of several supported platforms"
  - "whose caps match" → "with caps matching the target device"
  - "however, to do so" → "but to do so" (fewer commas)
  - nil description split into two clear sentences
  - iocp_th rewritten as proper sentence
  - "Similarly, then xNVMe provide" → "Similarly, xNVMe provides"
  - "driver name" → "drive letter" in Windows files doc

  Typo fixes

  - "effiency" → "efficiency"
  - "the can" → "they can"
  - "wether" → "whether"

docs/background/backends/xnvme_be_macos.md
```diff
@@ -0,0 +1,23 @@
...
+This backend allows users of Thunderbolt NVMe devices to access them through
+**xNVMe**. It supports both custom admin commands and I/O.
+
+MacVFN currently has some limitations, as you'll need to disable System
+Integrity Protection (SIP). The device will also be completely hooked by the
+MacVFN driver and not appear as a mountable volume anymore.
+
+To use with **xNVMe**, the fastest way to identify devices, is to run `xnvme enum` on the target
+system. If succesful, it will present devices in the form of `MacVFN-abcd`
+, where the latter part is the serial number of the device.
```
NaddiNadja: I don't know if it's important enough to actually change, but I just wanted to point out that there's something about the formatting of this line. Line 14 is very long, and the line break after line 15 might cause a space between `MacVFN-abcd` and the following comma.

NaddiNadja (changes requested):
Amazing that the multi-enumeration now works! :)

I have only one comment this time, but I saw there was still one comment hiding down in the `docs/background/Hplatforms/linux/io_uring.md` file from my last review, just so you're aware :)

lib/xnvme_platform.c
```diff
@@ -0,0 +1,229 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <errno.h>
+#include <stddef.h>
+#include <libxnvme.h>
+#include <xnvme_be.h>
+#include <xnvme_dev.h>
+#include <xnvme_platform.h>
+
+struct xnvme_platform *g_xnvme_platform =
```
karlowich, line 14: I see that it is technically possible that `g_xnvme_platform == NULL`. Do we want to protect against this case? Looks like every function assumes that `g_xnvme_platform != NULL`.

docs/background/backends/xnvme_be_macos.md
safl: I would prefer to defer to a general docs cleanup, it could also include tasks such as converting remaining ``.rst`` to ``.md``, fixing line-width like the one missing here and other stylistic alignment.

lib/xnvme_platform.c
safl: Hmm... I guess it technically should not be possible and it ought to have a compile-errror in case it is built without a platform. Along these lines then the macros ``XNVME_BE_*_ENABLED`` should actually be ``XNVME_PLATFORM_*_ENABLED`` since that is way they toggle..

lib/xnvme_be_spdk_dev.c
```diff
@@ -567,12 +567,10 @@ xnvme_be_spdk_enumerate(const char *sys_uri, struct xnvme_opts *opts, xnvme_enum
 			XNVME_DEBUG("FAILED: ident_from_uri, uri: %s", sys_uri);
 			return -EINVAL;
 		}
-		if (sscanf(ident.uri, "%[^:]:%d", addr, &port) != 2) {
-			XNVME_DEBUG("FAILED: invalid sys_uri: %s", sys_uri);
-			return -EINVAL;
-		}
+		sscanf(ident.uri, "%[^:]:%d", addr, &port);
```
karlowich: BUMP:
Why are we no longer checking if `sys_uri` is malformed?

docs/background/backends/xnvme_be_macos.md
safl: Addressing it now since a respin is needed anyway...

cijoe/tests/examples/test_xnvme_single_async_efd.py
```diff
@@ -5,15 +5,9 @@
...
-    if be_opts["be"] in ["linux"] and be_opts["async"] not in [
-        "libaio",
-        "io_uring",
-        "io_uring_cmd",
-    ]:
-        pytest.skip(
-            reason=f"[be={be_opts['be']}, async={be_opts['async']}] Eventfd not supported"
-        )
+    if be_opts["admin"] == "libvfn":
+        pass  # libvfn supports eventfd
```
karlowich: This check does nothing?

lib/xnvme_platform.c
safl: Addressed in ```refactor(be): rename XNVME_BE_{OS}_ENABLED to XNVME_PLATFORM_{OS}_ENABLED```

docs/background/backends/xnvme_be_macos.md
safl: done :)

karlowich (changes requested):
Getting very close :)

include/xnvme_be_cref.h
karlowich: Looks like this is working now, right?

lib/xnvme_be_fbsd_dev.c
karlowich: What is the status on this?

cijoe/tests/examples/test_xnvme_single_async_efd.py
safl: It avoids libvfn getting skipped. But it could be simplified.

lib/xnvme_be_spdk_dev.c
safl: Good catch, thx.

cijoe/tests/examples/test_xnvme_single_async_efd.py
safl: addressed in ```fix(tests): simplify eventfd skip logic in test_xnvme_single_async_efd```

lib/xnvme_be_spdk_dev.c
safl: Addressed in ``` fix(be/spdk): add error handling for sys_uri sscanf in enumerate```

cijoe/tests/examples/test_xnvme_single_async_efd.py
karlowich: Ah, I completely missed that the others were listed as "not in"

karlowich (dismissed):
All good now

docs/background/platforms/linux/io_uring.md
NaddiNadja: @safl, you have not addressed this comment

safl: This is not re-done for the entire background-section.

NaddiNadja (dismissed):
It looks overall good. As mentioned in my last review, there was still an unresolved comment from my review from before that - but I'll approve this PR so that you can merge. If we'll do a fixup of the docs in the future, this sentence might not be necessary to fix right now.

safl:
> It looks overall good. As mentioned in my last review, there was still an unresolved comment from my review from before that - but I'll approve this PR so that you can merge. If we'll do a fixup of the docs in the future, this sentence might not be necessary to fix right now.

Address now and squashed in.

karlowich (approved):

NaddiNadja (approved):
great! :)

Issue #618, opened 2025-10-03
Update toolbox packages


safl:
It has been a long time since the last update of the packaging scripts, e.g. atleast these;

## Ubuntu

- Add 25.10
- Drop 24.10 (Oracular Oriole)
- Add 24.04 LTS (Noble Numbat)
- Drop 20.04 LTS (Focal Fossa)

## Fedora

- Add 42
- Drop 39

--- while at it, then ensure permissions are correct on the scripts some are missing +x ---

safl:
Closed by https://github.com/xnvme/xnvme/pull/625

PR #625, opened 2026-01-28
Distro bumps fixing #618 and more

safl:
This PR addresses #618 by updating distributions and fixing a missing executable bit. It also includes a Gentoo fix and several version bumps. See commits for details.


safl:
@NaddiNadja, could you help resolve the FreeBSD issues? They’re currently blocking the CI pipeline and need attention.

Somewhat related: I’ve bumped the Linux distributions and updated the miscellaneous OS builds. FreeBSD may ultimately require a full image rebuild, but I’m hoping we can find a workaround until new images for FreeBSD 14 and 15 are available.

toolbox/pkgs/freebsd-14.sh
```diff
@@ -13,7 +13,7 @@ pkg install -qy \
  cunit \
  devel/py-pipx \
  devel/py-pyelftools \
- libuuid \
+ e2fsprogs-libuuid \
```
NaddiNadja: This might be your problem, this has been removed. should just be `libuuid` as before. See this commit: https://github.com/xnvme/xnvme/commit/fa2e6f2addf600e214c3f2474cc26d06786b76d9

toolbox/pkgs/freebsd-15.sh
```diff
@@ -13,7 +13,7 @@ pkg install -qy \
  cunit \
  devel/py-pipx \
  devel/py-pyelftools \
- libuuid \
+ e2fsprogs-libuuid \
```
NaddiNadja: Again, this might be your problem, this has been removed. should just be libuuid as before.

NaddiNadja (changes requested):
I only looked at the freebsd part for now. The error said gmake didn't exist, and it's because it fails to run the toolbox/pkgs script, becasue  e2fsprogs-libuuid does not exist anymore. :)

toolbox/pkgs/freebsd-15.sh
safl: Ah, right! Trying out a fix for that!

safl: Great, there was a regression on the libuuid thing... however... that was not the only issue. Another emerged which CI will verify whether or not is sufficiently fixed.

NaddiNadja (approved):
This is great! :)

karlowich (approved):
Very nice!

karlowich:
You also have to change the merge requirements. Right now it depends on running the deprecated `verify (freebsd, 13)`

PR #629, opened 2026-02-05
Fixes to developer tooling

safl:
Migrate xNVMe reference development platform from Debian Bullseye to Debian Trixie, with fixes for kernel behavior changes and QEMU guest configuration improvements discovered during the migration.

## Platform Migration

- Replace Debian **Bullseye** (11) with **Trixie** (13) as reference platform
- Update all CI workflows, configs, and Makefile targets
- Pin cijoe to v0.9.58 and fix pipx install/inject
- Set `memlock unlimited` in boot image for DPDK/VFIO support

## QEMU Guest Configuration

- Refactor guest script with declarative config and auto-detect QEMU features (`features.get()`)
- Flatten NVMe config: one namespace per controller, individual PCIe root ports per controller (avoids IO port space exhaustion on Trixie kernel)
- Use `qemu-system-{arch}` from `PATH` for portability
- Add device enumeration diagnostic step to test workflows

## Bug Fixes

- **fix(dev):** Cap Linux NVMe ioctl passthrough to 32 blocks — Trixie kernel reduces `max_segments` from 127 to 33, causing `sc: 0x16` (Invalid
Field in Command) on writes exceeding the limit
- **fix(docgen):** Add `udevadm settle` after driver reset to fix device node race condition
- **fix(python):** List `libxnvme.h` first for clang2py header parsing (Python 3.12+ / LLVM 17+)
- **fix(tests):** Rebind driver after subsystem reset in `test_subsystem_reset` — the kernel requires a full unbind/rebind cycle to recover

## Documentation

- Update all `kmdo` files (`.cmd`) to use new device paths (ZNS on nvme1, FDP on nvme5, etc.)

 ## Hacks

Two tests are disabled on Debian Trixie pending proper investigation:

- [ ] https://github.com/xnvme/xnvme/issues/630 -- `test_list` in Python KV bindings hangs on QEMU KV emulation
- [ ] https://github.com/xnvme/xnvme/issues/631 -- `mem_map_unmap` in libvfn PCIe tests

These are not clever workarounds, they are hackjobs that punt the proper fix by disabling the tests entirely. However, this PR contains enough improvements that it seems more valuable to land these changes rather than hold them back until everything is perfect. Also, these issues are likely surfaced due to changes in the environment (Debian Trixie kernel, Python 3.12+, Clang/LLVM 17+) rather than introduced by the changes here.


cijoe/configs/ax42-freebsd.toml
```diff
@@ -33,7 +33,8 @@ default_systemimage = "freebsd-14-x86_64"
 prefix = "{{ local.env.HOME }}/opt/qemu"
 
 [qemu.systems.x86_64]
-bin = "/opt/qemu/bin/qemu-system-x86_64"
+#bin = "/opt/qemu/bin/qemu-system-x86_64"
```
karlowich: I think we can get rid of the comment. It doesn’t add much in my opinion. /opt/ is not a very common path for QEMU.

karlowich (changes requested):
Just the one comment.
I like this new wave of simplicity!

cijoe/configs/ax42-freebsd.toml
karlowich: Of course the same goes for the other files in the commit too

safl: fixed.

cijoe/scripts/xnvme_guest_start_nvme.py
```diff
@@ -114,32 +156,23 @@ def namespace(controller_id, nsid, aux={}):
...
-    # Nvme0n2 - ZNS namespace
-    zoned_attributes = {
+    # =========================================================================
+    # Feature flags - ZNS/FDP/PI are standard QEMU features (always enabled),
+    # KV requires QEMU patches (auto-detected)
+    # =========================================================================
+    ENABLE_ZNS = True  # Zoned Namespaces - standard since QEMU 6.0
+    ENABLE_KV = features.get("kv", False)  # Key-Value - requires QEMU patches
+    ENABLE_FDP = True  # Flexible Data Placement - standard since QEMU 8.0
+    ENABLE_PI = True  # Protection Information - standard QEMU feature
```
NaddiNadja: If they (zns/fpd/pi) are always enabled, what is the reason to retrieve them in the `detect_qemu_nvme_features()`? It seems odd that we specifically lookup these features and do not use them anywhere, at least with a `features.get("zns", True)`.

NaddiNadja (commented):
I have a single question inline :)

cijoe/scripts/xnvme_guest_start_nvme.py
safl: It is for those occassion where you sit and debug, and perhaps qemu is giving you a hard time with the FDP drives, but you are really looking at something else entirely, then it is now just a very easy manual edit, then run guest_start again and your dev. environment is working again.

safl: So, now there is a new little place to go and comment out, instead of having to parse all the device-setup logic of the script.

NaddiNadja: In this situation, you're saying you would just go and change line 165 to `ENABLE_FDP = False`? I still don't understand why we would in this case parse the features in `detect_qemu_nvme_features()` then. 
Having the original code be `ENABLE_FDP = features.get("fdp", True)` would not make it more complex to change the code when debugging to `= False`, right? Or if the parsing always return True, then maybe just not parse them?

safl: Right, now I get what you were asking. You are right. I got blind on that code because these things are hairpullingly frustrating.

safl: This should be fixed now.

NaddiNadja: great! :)

lib/xnvme_dev.c
```diff
@@ -412,8 +412,8 @@ xnvme_dev_derive_geo(struct xnvme_dev *dev)
 
 	/** quirk: Linux kernel max-segments potentially less than mdts **/
 	if ((!strncmp(dev->be.attr.name, "linux", 5)) && (!strncmp(dev->be.sync.id, "nvme", 4)) &&
-	    (dev->geo.lba_nbytes) && ((dev->geo.mdts_nbytes / dev->geo.lba_nbytes) > 127)) {
-		dev->geo.mdts_nbytes = dev->geo.lba_nbytes * 127;
+	    (dev->geo.lba_nbytes) && ((dev->geo.mdts_nbytes / dev->geo.lba_nbytes) > 32)) {
```
karlowich: Looks like getting the actual max-segments is as simple as reading `/sys/block/nvme0n1/queue/max_segments`. I don't think we should defer querying this, I see the value is still 128 on Ubuntu 24.04.

cijoe/scripts/xnvme_guest_start_nvme.py
```diff
@@ -149,114 +182,80 @@ def namespace(controller_id, nsid, aux={}):
...
+        # nvme1: Fabrics testing
+        ("nvme1", "adcdbeef", 7, 2, None, [(1, {})]),
+        # nvme2: Large MDTS
+        ("nvme2", "beefcace", 7, 3, None, [(1, {})]),
+    ]
+
+    # Add FDP controller if enabled
+    if ENABLE_FDP:
+        nvme_config.append(
+            ("nvme3", "beefcace", 7, 4, fdp_subsys, [(1, {"fdp.ruhs": "'0;5;6;7'"})])
```
karlowich: The `beefcace` id (subnqn?) is also used for nvme2

cijoe/configs/debian-trixie.toml
```diff
@@ -28,7 +28,8 @@ default_systemimage = "debian-13-x86_64"
 
 [qemu.systems.x86_64]
 #bin = "{{ local.env.HOME }}/opt/qemu/bin/qemu-system-x86_64"
-bin = "qemu-system-x86_64"
+bin = "/opt/qemu/bin/qemu-system-x86_64"
+#bin = "qemu-system-x86_64"
```
karlowich: I don't think you meant to commit this change

cijoe/scripts/xnvme_guest_start_nvme.py
karlowich, line 1: Looks like you have two commits completely rewriting this file. Perhaps they can be squashed?

karlowich (changes requested):
Great stuff. Left a few comments.

cijoe/scripts/xnvme_guest_start_nvme.py
karlowich: Probably irrelevant, since it gets changed in a later commit.

cijoe/configs/debian-trixie.toml
safl: Absolutely right.

cijoe/scripts/xnvme_guest_start_nvme.py
safl: They are two different major changes, one does feature-detection, the other flattens controller/namespace config.

safl: but... it might be beneficial to take a closer look at this.

lib/xnvme_dev.c
safl: ack.

karlowich: Great!

cijoe/scripts/xnvme_guest_start_nvme.py
safl: Squashed the guest-start-changes together.

cijoe/scripts/xnvme_guest_start_nvme.py
safl: Squashed.

karlowich (dismissed):
Great!

Makefile
```diff
@@ -232,7 +247,10 @@ endef
 .PHONY: cijoe-setup-xnvme-using-git
 cijoe-setup-xnvme-using-git:
 	@echo "## xNVMe: cijoe-setup-xnvme-using-git"
-	@cd cijoe && cijoe workflows/provision-using-git.yaml -l \
+	@cd cijoe && cijoe workflows/provision-using-git.yaml \
+		--monitor \
+		--config "configs/debian-bullseye.toml" \
+		-l \
```
NaddiNadja: You have removed the @ on some of the calls to cijoe, but not all of them (I only comment on this one, but htere are multiple places9. Do we want to streamline this?

NaddiNadja (changes requested):
Single thing :)

Makefile
safl: yeah... fixed.

karlowich (approved):

NaddiNadja (approved):

PR #633, opened 2026-02-20
Introduce P2P buffers to uPCIe

karlowich:
Don't be discouraged by the size of this PR. Currently, it includes all commits from #596 and #638. As these get merged the size of this will shrink.
Only the following commits introduce a new `upcie-cuda` backend:

- feat(thirdparty/upcie): vendor p2p requirements
- feat(build/upcie): add cuda dependency
- feat(be/upcie): add uPCIe CUDA P2P backend 

The `upcie-cuda` backend is CPU driven with CUDA P2P buffers as payload.


karlowich:
This is ready for review now.
I have refactored the cuda part out of the upcie header and into a cuda specific one.
Additionally, I added the iovec and metadata buffer support

safl (changes requested):
This is excellent. Some chores that needed:

- [ ] Documentation, how to use e.g. ``be=upcie-cuda`` and what are assumptions and requirements to system setup, this should include a description of the non-upstream kernel-patches needed to get it to work
- [ ] ~~Testing, although for now, let's just move ahead as we are multiple people capable of testing it manually, and we have a plan for the road ahead https://github.com/xnvme/xnvme/issues/648 https://github.com/xnvme/xnvme/issues/647~~

Then one minor thing, since this introduces the backend for which the buffer-pre-work and ``XNVME_CUDA_ENABLED`` was introduced. How about just renaming ``XNVME_CUDA_ENABLED`` to ``XNVME_BE_UPCIE_ENABLED``?

toolbox/third-party/linux/upcie/cudamem_heap.h
```diff
@@ -0,0 +1,337 @@
...
+
+				newblock->vaddr = block->vaddr + size;
+				newblock->free = 1;
+				newblock->size = remaining;
+				newblock->next = block->next;
+
+				block->next = newblock;
+				block->size = size;
+
+			}
```
NaddiNadja, line 333: remove empty line before curly bracket

toolbox/third-party/linux/upcie/cudamem_heap.h
```diff
@@ -0,0 +1,337 @@
...
+
+			block->free = 0;
+			return (void *) block->vaddr;
+		}
+		block = block->next;
+	}
```
NaddiNadja, line 340: insert empty line after curly bracket

toolbox/third-party/linux/upcie/cudamem_heap.h
```diff
@@ -0,0 +1,337 @@
...
+static inline void *
+cudamem_heap_block_alloc_aligned(struct cudamem_heap *heap, size_t size, size_t alignment)
+{
+	struct cudamem_heap_block *block = heap->freelist;
+
+	size = (size + alignment - 1) & ~(alignment - 1);
+
+	while (block) {
+		if (block->free && block->size >= size) {
+			size_t remaining = block->size - size;
```
NaddiNadja: Just to be sure that I'm not missing anything: I can see in hostmem_heap.c, that we use `(size+alignment)` in the if statement and for calculating `remaining` (although `size` has been overwritten in the same way as line 239 here). Is it on purpose that we "just" use `size` here?

toolbox/third-party/linux/upcie/dmabuf.h
```diff
@@ -0,0 +1,205 @@
...
+{
+	struct udmabuf_attach attach;
+	struct udmabuf_get_map *map = NULL;
+	int udmabuf_fd, err;
+	size_t map_size, pages_size;
+
+	udmabuf_fd = open("/dev/udmabuf", O_RDWR);
+	if (udmabuf_fd < 0) {
+		err = -errno;
+		UPCIE_DEBUG("FAILED: open(/dev/udmabuf), errno: %d", err);
```
NaddiNadja, line 108: this is not urgent, but we usually do `"FAILED: <...>; errno(%d)"`, so semicolon instead of comma and parenthesis around the number instead of colon and space

i only comment once, but it's in multiple places :)

toolbox/third-party/linux/upcie/dmabuf.h
```diff
@@ -0,0 +1,205 @@
...
+		goto exit;
+	}
+	memset(map, 0, sizeof(*map));
+	map->fd = dmabuf_fd;
+	map->count = attach.count;
+
+	err = ioctl(udmabuf_fd, UDMABUF_GET_MAP, map);
+	if (err) {
+		err = -errno;
+		UPCIE_DEBUG("FAILED: ioctl(UDMABUF_GET_MAP), errno: %d\n", err);
```
NaddiNadja, line 138: Do you need newline here?

toolbox/third-party/linux/upcie/dmabuf.h
```diff
@@ -0,0 +1,205 @@
...
+	memset(dmabuf, 0, sizeof(*dmabuf));
+	dmabuf->fd = dmabuf_fd;
+	dmabuf->npages = map->count;
+	pages_size = sizeof(struct dmabuf_page) * dmabuf->npages;
+	dmabuf->pages = malloc(pages_size);
+	if (!dmabuf) {
+		err = -errno;
+		UPCIE_DEBUG("FAILED: malloc(dmabuf->pages), errno: %d", err);
+		goto exit;
+	}
```
NaddiNadja, line 153: Should be `(!dmabuf->pages)`, right?

toolbox/third-party/linux/upcie/cudamem_heap.h
```diff
@@ -0,0 +1,337 @@
...
+	heap->nphys = heap->size / heap->pagesize;
+	heap->phys_lut = calloc(heap->nphys, sizeof(uint64_t));
+	if (!heap->phys_lut) {
+		err = -errno;
+		UPCIE_DEBUG("FAILED: calloc(phys_lut), errno: %d", err);
+		goto error;
+	}
+	err = dmabuf_get_lut(&heap->dmabuf, heap->nphys, heap->phys_lut, heap->pagesize);
+	if (err) {
+		UPCIE_DEBUG("FAILED: dmabuf_get_lut(), err: %d", err);
```
NaddiNadja: new line after curly bracket

NaddiNadja (changes requested):
Very nice! I allowed myself to have some styling comments, since there were also other things inline, so you would need a rerun of the CI anyway :)

karlowich:
@safl 
> Documentation, how to use e.g. be=upcie-cuda and what are assumptions and requirements to system setup, this should include a description of the non-upstream kernel-patches needed to get it to work

Right, of course.

> Then one minor thing, since this introduces the backend for which the buffer-pre-work and XNVME_CUDA_ENABLED was introduced. How about just renaming XNVME_CUDA_ENABLED to XNVME_BE_UPCIE_ENABLED?

I thought about this, I think it depends on whether we expect to have other backends using CUDA. Currently, the two flags control two very different things:
- XNVME_CUDA_ENABLED: handle CUDA buffers
- XNVME_BE_UPCIE_CUDA_ENABLED: enable the specific uPCIe CUDA backend

karlowich:
@NaddiNadja Your comments are for the uPCIe files. Since these are already upstream, how about creating a PR there fixing the things and then I will update it here after?

toolbox/third-party/linux/upcie/dmabuf.h
NaddiNadja: I've opened a PR on upcie

NaddiNadja:
I will come back and review when the fixes have been added to uPCIe, such that you can vendor the "correct" version :)

karlowich:
@NaddiNadja Merged your style fixes.

lib/upcie/xnvme_be_upcie_cuda_dev.c
```diff
@@ -0,0 +1,90 @@
...
+			cuCtxDestroy(g_upcie_cuda_rte.cu_ctx);
+			return NULL;
+		}
+	}
+
+	ctrlr = xnvme_be_upcie_ctrlr_init(dev);
+	if (!ctrlr) {
+		err = errno;
+		XNVME_DEBUG("FAILED: xnvme_be_upcie_ctrlr_init(); err(%d)", err);
+		errno = err;
```
NaddiNadja: err = errno and then errno = err ?

lib/upcie/xnvme_be_upcie_cuda_dev.c
NaddiNadja, line 1: This is kinda nit-picky, but in xnvme_be_upcie_dev, we have the init function before the term function - would be nice to have the same setup here.

lib/upcie/xnvme_be_upcie_cuda_sync.c
```diff
@@ -0,0 +1,156 @@
...
+		return -errno;
+	}
+
+	req->user = ctx;
+	cmd->cid = req->cid;
+
+	if (dbuf) {
+		nvme_request_prep_command_prps_contig_cuda(req, &g_upcie_cuda_rte.cuda_heap, dbuf,
+							   dbuf_nbytes, cmd);
+	}
```
NaddiNadja, line 47: In the non-cuda upcie implementation (`xnvme_be_upcie_sync_cmd_io()`), we have an else case for this if-statement. Is that not needed here? Also goes for the `iov` function.

lib/upcie/xnvme_be_upcie_cuda_dev.c
karlowich: This is the uPCIe backend breaking the convention.

lib/upcie/xnvme_be_upcie_cuda_sync.c
karlowich: The upcie_sync is using a different API which encapsulates every step, thus they need to change the function call. Here we manually do every step. So, if there's no dbuf we don't populate the PRP

NaddiNadja: great, i fell for this like 5 times, and somehow still got confused hahaha

NaddiNadja (dismissed):
fine fine fine - you'll get your lgtm now

karlowich:
The docs are here!

NaddiNadja (dismissed):
docs also look good to me! :)

karlowich:
Hold off on reviewing this, I don't think the latest commit was the right direction

karlowich:
All good now, dropped the commit that I didn't like

NaddiNadja (dismissed):
Looks good to me :)

safl:
> @safl
> 
> > Documentation, how to use e.g. be=upcie-cuda and what are assumptions and requirements to system setup, this should include a description of the non-upstream kernel-patches needed to get it to work
> 
> Right, of course.
> 
> > Then one minor thing, since this introduces the backend for which the buffer-pre-work and XNVME_CUDA_ENABLED was introduced. How about just renaming XNVME_CUDA_ENABLED to XNVME_BE_UPCIE_ENABLED?
> 
> I thought about this, I think it depends on whether we expect to have other backends using CUDA. Currently, the two flags control two very different things:
> 
>     * XNVME_CUDA_ENABLED: handle CUDA buffers
> 
>     * XNVME_BE_UPCIE_CUDA_ENABLED: enable the specific uPCIe CUDA backend

When is XNVME_CUDA_ENABLED useful without XNVME_BE_UPCIE_CUDA_ENABLED?

karlowich:
@safl It depends on whether we expect to have other backends using CUDA.
One tangible case is IO_Uring with GPU-buffers, this would be quite easy for us to implement - when it is upstream

safl:
> @safl It depends on whether we expect to have other backends using CUDA. One tangible case is IO_Uring with GPU-buffers, this would be quite easy for us to implement - when it is upstream

So, in other words, the split is only useful once additional CUDA-capable
backends exist. As things stand today, without those additional changes, it seems the two
flags do not provide separate value. Is that correct?

karlowich:
> it seems the two flags do not provide separate value

Yes and no. One **could** use the xNVMe API to abstract away types of buffers (host / device) in any project.
I don't think it is very realistic that this would be used. To me the deciding factor to separate them was that semantically it is illogical that enabling a backend can impact libxnvme_buf.h.

safl:
> > it seems the two flags do not provide separate value
> 
> Yes and no. One **could** use the xNVMe API to abstract away types of buffers (host / device) in any project. I don't think it is very realistic that this would be used. To me the deciding factor to separate them was that semantically it is illogical that enabling a backend can impact libxnvme_buf.h.

Keeping our eye on the proverbial ball, the goal here is to enable CPU-initiated I/O with P2P buffers on a CUDA-enabled device.

From that perspective, the changes to the library buffer helpers were necessary to support transparent use. I do not see a strong reason to enable that code more broadly for use cases that are not part of the current scope.

Additional CUDA-capable backends may emerge in the future, but since they are not part of the current work, it seems reasonable to address any required changes to flags or abstractions when those use cases materialize.

karlowich:
Okay, my last argument is that currently these are also separated in the meson build.
One option controls CUDA and one controls uPCIe. If both are enabled you upcie_cuda. However, you can disable upcie and still enable CUDA. This gives you the changes to libxnvme_buf.
What you are proposing would make CUDA a dead option if uPCIe is disabled.
The XNVME_* flags are not user-facing, but the meson options are. I would prefer to have one too many internal definitions compared to a potentially dead meson option.

safl:
> Okay, my last argument is that currently these are also separated in the meson build. One option controls CUDA and one controls uPCIe. If both are enabled you upcie_cuda. However, you can disable upcie and still enable CUDA. This gives you the changes to libxnvme_buf.

Yes, and I do not see the value in doing so. 

> What you are proposing would make CUDA a dead option if uPCIe is disabled. The XNVME_* flags are not user-facing, but the meson options are. I would prefer to have one too many internal definitions compared to a potentially dead meson option.

Let's keep focus. The functionality/feature is: CPU-initiated I/O with P2P buffers on CUDA enabled devices. This needs a backend implementation adn auxilirary code-paths. The backend implementation has dependencies. When dependencies are met the feature is enabled some flag is set and the necessary code-paths are defined. Otherwise they are not.

karlowich:
@safl I merged the flags. And added a bug fix.

karlowich:
... Unrelated MacOS CI failure ...

karlowich:
@safl @NaddiNadja This is good to go now.
To see more about the uPCIe changes, look here: https://github.com/safl/upcie/pull/23

safl (commented):
Why aren't these "upstream"?

[chore(thirdparty/upcie): sync to v0.4.1](https://github.com/xnvme/xnvme/pull/633/changes/26c61753d6ed1a2eba4122828b44117c98131623)
    [karlowich](https://github.com/xnvme/xnvme/commits?author=karlowich)
    karlowich
    committed
    20 hours ago

Commits on Mar 25, 2026

    [feat(thirdparty/upcie): add CUDA P2P support](https://github.com/xnvme/xnvme/pull/633/changes/c8d414b15eafb91d4e959be19c18fd049e5724ac)
    [karlowich](https://github.com/xnvme/xnvme/commits?author=karlowich)
    karlowich
    committed
    3 hours ago
    [fix(thirdparty/upcie): add missing trailing newlines to CUDA headers](https://github.com/xnvme/xnvme/pull/633/changes/45ce96a93d69cf1b7329a2fe1a7f56c365c92c69)
    [karlowich](https://github.com/xnvme/xnvme/commits?author=karlowich)
    karlowich
    committed
    2 hours ago
    [fix(thirdparty/upcie): close dma-buf fd after attach/detach](https://github.com/xnvme/xnvme/pull/633/changes/54a6bbeffed39d621d08dd24a284720d3d15b12f)
    [karlowich](https://github.com/xnvme/xnvme/commits?author=karlowich)
    karlowich
    committed
    2 hours ago
    [fix(thirdparty/upcie): free last block in cudamem heap freelist](https://github.com/xnvme/xnvme/pull/633/changes/3ef4e36299d1f9ef000c3a0968f1359fa04fc102)
    [karlowich](https://github.com/xnvme/xnvme/commits?author=karlowich)
    karlowich
    committed
    2 hours ago

karlowich:
> Why aren't these "upstream"?

They are. My mistake. Now everything is in the first commit.

safl (approved):
This is looking great, and really nice with the docs. We do have challenges but those are rooted in the design of xNVMe and how closely xNVMe buffers are tied to ``struct xnvme_dev`` and we need to address that separately and not hold this back due to that. It is quite impressive what you manage to do here withing that frame. In other words, the changes here allows for some very exciting exploration and utility, thus let's move forward with it. Awesome stuff.

NaddiNadja (approved):
looks great to me! ⭐

PR #635, opened 2026-02-23
Preparation for uPCIe and CI plumbing

safl:
While working on refactor on https://github.com/xnvme/xnvme/pull/623 for https://github.com/xnvme/xnvme/pull/596 then this collection of fixes would be good to get out of the way while tackling the logic in the backend refactor itself.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -725,26 +725,8 @@ jobs:
       run: |
         cijoe -r
 
-    # cd'ing into 'cijoe' to auto-collect the non-packaged github-specific configs and
-    # workflows.
-    - name: CIJOE, run provision-using-tgz.yaml
-      run: |
-        cd cijoe && cijoe "workflows/provision-using-tgz.yaml" \
-        --monitor \
-        --config "configs/${{ matrix.guest.os }}-${{ matrix.guest.ver }}.toml" \
-        --config "configs/fio.toml" \
-        --config "configs/xnvme.toml" \
-        --config "configs/system_imaging.toml" \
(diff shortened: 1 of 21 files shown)


.github/workflows/verify.yml
```diff
@@ -374,11 +374,19 @@ jobs:
         --config "configs/ramdisk.toml" \
         --output "${{ github.job }}-${{ matrix.container.os }}-${{ matrix.container.ver }}-test-ramdisk"
 
+    - name: CIJOE, upload report.html
```
karlowich, line 365: Can the html really standalone? It looks to me as if it has ties to the data in the directories generated by Cijoe

karlowich (commented):
It looks really good. I'm skeptical of including uPCIe in this. Why do we need to vendor it before any code that uses it?

.github/workflows/verify.yml
karlowich: Huh, just tried it out on the PR - I guess it can be standalone

safl:
> It looks really good. I'm skeptical of including uPCIe in this. Why do we need to vendor it before any code that uses it?

Because then we can work in parallel on different things that need it, and not have uPCIe on top for each of them.
Alternatively we can just leave it out, but I have something that does need, in its current form, so for that it would be nice to just get it in. Another piece of work that can go in parallel is replacing the tooling of ``xnvme-driver`` with ``devbind`` and ``hugepages``.

To expedite this, then i dropped the inclusion of uPCIe.

karlowich:
> Because then we can work in parallel on different things that need it

This is a very valid reason, I didn't think of that

karlowich (approved):
All good now.

NaddiNadja (approved):
great! lgtm :)

PR #636, opened 2026-02-23
Benchmarking tool: xnvmeperf

NaddiNadja:


karlowich:
You can add the tests to `cijoe/tests/tools`. You can look to the other tests in the folder for inspiration

karlowich:
Also, I think we need to use a meson feature to control whether to build xnvmeperf, since it has the cJSON dependency and it shouldn't break if isn't available

tools/xnvmeperf/src/xnvmeperf_run.c
```diff
@@ -0,0 +1,626 @@
...
+	case IOPATTERN_VERIFY:
+		return "verify";
+	case IOPATTERN_RESET:
+		return "reset";
+	case IOPATTERN_UNMAP:
+		return "unmap";
+	case IOPATTERN_FLUSH:
+		return "flush";
+	case IOPATTERN_WRITE_ZEROES:
+		return "write_zeroes";
```
karlowich: I don't think we need to define all of these before we need them. It looks like some of them will never be used (i.e. unmap or reset)

tools/xnvmeperf/src/xnvmeperf_run.c
```diff
@@ -0,0 +1,626 @@
...
+			goto exit;
+		}
+
+		job->buf = xnvme_buf_alloc(job->dev, args->iosize);
+		if (!job->buf) {
+			err = -errno;
+			xnvme_cli_perr("Failed: xnvme_buf_alloc()", err);
+			goto exit;
+		}
+		memset(job->buf, 0xA5, args->iosize);
```
karlowich: We should use `xnvme_buf_clear()` and `xnvme_buf_fill()` instead of memset, this way it can be CUDA compatible in the future.
Also, the choice of `0xA5` seems a little strange

tools/xnvmeperf.c
```diff
@@ -0,0 +1,626 @@
...
+{
+	size_t lba_size = nbytes / nlb;
+
+	for (uint16_t i = 0; i < nlb; i++) {
+		uint8_t *p = (uint8_t *)buf + i * lba_size;
+		uint64_t lba = slba + i;
+
+		memset(p, 0xAA, lba_size);
+		memcpy(p, &lba, sizeof(lba));
+	}
```
karlowich, line 643: This code is a little strange. Also we should use xnvme_buf_clear() and xnvme_buf_fill() instead of memset, this way it can be CUDA compatible in the future.

safl:
> Also, I think we need to use a meson feature to control whether to build xnvmeperf, since it has the cJSON dependency and it shouldn't break if isn't available

Adding to this, let's go with ``toml`` instead of ``json``. It is nicer for config-files, and fits other things we are working on as well.
Also, there might be a small toml-reader that can simply be vendored/ embedded. If not, then do a conditional built based on whether the package is available for linkage on the system, similar to the other dependencies.

tools/xnvmeperf/src/xnvmeperf_run.c
```diff
@@ -0,0 +1,626 @@
...
+		job.opcode = XNVME_SPEC_NVM_OPC_READ;
+		job.buf = read_buf;
+		job.offset = 0;
+		job.io_completed = 0;
+		job.io_failed = 0;
+
+		for (int i = 0; i < nios; i++) {
+			uint64_t slba;
+
+			memset(read_buf, 0, args->iosize);
```
karlowich: We should use xnvme_buf_clear() and xnvme_buf_fill() instead of memset, this way it can be CUDA compatible in the future.

tools/xnvmeperf.c
```diff
@@ -0,0 +1,626 @@
...
+			// the slba that was just submitted rather than capturing it before the
+			// call
+			slba = job.offset - job.nlb;
+
+			while (job.io_completed + job.io_failed < (uint64_t)(i + 1)) {
+				xnvme_queue_poke(job.queue, 0);
+			}
+
+			fill_pattern(expect_buf, args->iosize, slba, job.nlb);
+			if (memcmp(read_buf, expect_buf, args->iosize) != 0) {
```
karlowich: Looks like we don't need `expect_buf`, we can compare with the write_buf instead.
Also use `xnvme_buf_diff()` instead of `memcmp`

karlowich (changes requested):
There's a lot to take in, but overall it is very nice.
Left an initial round of comments.

tools/xnvmeperf/src/xnvmeperf_run.c
NaddiNadja: yea true, i just looked at bdevperf, added them all initially and never removed the ones i didn't use - will fix

tools/xnvmeperf.c
NaddiNadja: I'll fix xnvme_buf_fill and add a docstring to explain

tools/xnvmeperf.c
NaddiNadja: No, the write buffer is continuously overwritten `nios` times, so it's not enough to "just" compare to that, as it only holds the last iteration's data written. So in the second phase (read and compare), I need some buffer to compare to, hende `expect_buf` - it's true that I could re-use the write buffer, but I think it's better to have each variable serve only one purpose.

tools/xnvmeperf/src/xnvmeperf_run.c
NaddiNadja: The reasoning for using 0xA5 is unnecessary as I move to xnvme_buf_fill, but for clarity's sake, then it's common to use a nonuniform bit pattern to easier expose errors when verifying reads/writes

safl (changes requested):
A couple of things that would be great to add:

- [x] A little intro to the use here similar to https://xnvme.io/tools/fio/index.html
- [x] Simplify usage; boil it down to cli-arguments (drop json/toml for now), this might require expanding the "xnvme_cli" module, but that would be great.
- [x] A test exercising the tool with using all available backends (note we have a bunch of changes coming in this area on backend-instrumentation on (https://github.com/xnvme/xnvme/pull/623)

Down the line:

- [ ] Extend the latency testing with numbers using xnvmeperf

lib/xnvme_be_spdk_async.c
```diff
@@ -38,6 +38,11 @@ xnvme_be_spdk_queue_init(struct xnvme_queue *q, int XNVME_UNUSED(opts))
 	qopts.io_queue_size = XNVME_MAX(queue->base.capacity, qopts.io_queue_size);
 	qopts.io_queue_requests = qopts.io_queue_size * 2;
 
+	qopts.delay_cmd_submit = true;
```
safl, line 41: Excellent.

meson.build
```diff
@@ -46,8 +46,15 @@ add_project_arguments(
   language: 'c'
 )
 
-add_project_arguments('-Db_lto='+ (not is_windows).to_string(), language: 'c')
-add_project_arguments('-Db_pie='+ get_option('hardening').to_string(), language: 'c')
+if not is_windows
```
safl, line 49: I thought meson did its thing and emitted the flags fitting compiler... sigh... oh well.. nice catch!

One suggestion, i believe that LTO can be placed in default_options as a meson-ism. E.g. ``'b_lto=true'``. Then we can leave it to meson to figure out whether it has flags for msvc (windows).

Regarding the hardening flag it is a bit... hmm... we need that because for distro-packaging i seem to recall that it was a requirement. So, we should aim to satisfy that despite one might not want it when building from source and chasing efficiency. Looks like that in more recent meson (0.63) we could do:

```
if get_option('hardening')
  meson.override_option('b_pie', true)
endif
```

Then we get nice meson-isms. Having said that... then after checking the supported distros then Ubuntu 22.04 is at meson 0.61, so we would break support there.... i guess one could then argue that hardening is a linux packaging concern, thus in that case then gcc is used and once can safely just pass compiler-args. By that logic we can be pragmatic and do what you did for hardening:

```
if get_option('hardening')
  add_project_arguments('-fpie', language: 'c')
  add_project_link_arguments('-pie', language: 'c')
endif
```

And then for LTO, add it to default_options, then when we rotate Ubuntu 22.04 out, then we can bump meson and use the option_overide() by then.

include/libxnvme_cli.h
```diff
@@ -334,7 +338,12 @@ enum xnvme_cli_opt {
 	XNVME_CLI_OPT_APPTAG_MASK = 123, ///< XNVME_CLI_OPT_APPTAG_MASK
 
 	XNVME_CLI_OPT_SDLBA = 124,
-	XNVME_CLI_OPT_END   = 125, ///< XNVME_CLI_OPT_END
+
+	XNVME_CLI_OPT_TIME      = 126, ///< XNVME_CLI_OPT_TIME
```
safl: What happened to 125?

include/libxnvme_cli.h
```diff
@@ -170,6 +170,9 @@ struct xnvme_cli_args {
 	uint32_t time;
 	const char *iopattern;
 	const char *cpumask;
+
+	char **posn;    ///< Remaining positional args (points into argv)
```
safl: add a ``const`` just to signal that one should not go and try and change the strings pointed to... since these are argv might even want to signal that the pointers cannot be changed either.. something like: ```const char * const *posn``` .. or perhaps just ```const char **posn```

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+	}
+
+close_devs:
+	close_tids = calloc(args->ndevs, sizeof(*close_tids));
+
+	// We use threads to close the devices, because it took a long time
+	// to close each, so this is to speed up the process.
+	if (close_tids) {
+		for (int i = 0; i < args->ndevs; i++) {
+			pthread_create(&close_tids[i], NULL, dev_close_fn, devs[i]);
```
safl: You probably need to check whether it created correctly and handle when does not, otherwise you get undefined behavior when you then in the next loop body run thread_join() on an invalid ``pthred_t``

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+		uint64_t mismatches = 0;
+
+		dev = xnvme_dev_open(args->dev_uris[d], &args->opts);
+		if (!dev) {
+			fprintf(stderr, "Failed: xnvme_dev_open(%s): %s\n", args->dev_uris[d],
+				strerror(errno));
+			err = -errno;
+			continue;
+		}
+		xnvme_dev_derive_geo(dev);
```
safl: You probably want error-handling here... since... if there is an issue deriving geometry then you are bound to hit other much worse issues soon after... or perhaps just very unexpected behavior that could have been caught early.

Makefile
```diff
@@ -825,6 +825,25 @@ clobber: clean
...
+	@echo "## xNVMe: xnvmeperf-example [DONE]"
+
+define xnvmeperf-verify-help
+# Verify xnvmeperf using NVMe device at /dev/nvme0n1
+endef
+.PHONY: xnvmeperf-verify
+xnvmeperf-verify:
+	@echo "## xNVMe: xnvmeperf-verify"
+	xnvme-driver reset
+	xnvmeperf verify --iosize 512 --count 1000 --be linux --async io_uring /dev/nvme0n1
```
safl: This might be a bit of a hazard... one might hit "make xnvmeperf" then see oh that was need and then try "make xnvmeperf-verify" and wipe their OS... I would prefer that the examples here made use of the ramdisk backend, then less severe accidents and then put examples in the docs as ``.cmd`` files and in a section similar to the fio description.

lib/xnvme_cli.c
```diff
@@ -862,6 +862,24 @@ static struct xnvme_cli_opt_attr xnvme_cli_opts[] = {
...
+	{
+		.opt = XNVME_CLI_OPT_TIME,
+		.vtype = XNVME_CLI_OPT_VTYPE_NUM,
+		.name = "time",
+		.descr = "Run for 'NUM' seconds",
+	},
+	{
+		.opt = XNVME_CLI_OPT_IOPATTERN,
+		.vtype = XNVME_CLI_OPT_VTYPE_STR,
+		.name = "iopattern",
```
safl, line 876: How about using the same names that ``fio`` uses? E.g. ``runtime``, ``rw`` and have ``verify`` as an explicit option and not an io-pattern. E.g. one might do **verify** on sequential as well as random io-patterns. Actually... i do prefer the "iopattern" name over e.g. "rw"... and we won't be fio-compatible anyway... so I would instead just go with the names that you use, as they are nicer... with one exception: use ``--runtime`` instead of ``--time``, i find ``--time`` to be a bit too "wide".

And also, have an explicit ``--verify`` flag, such that verification is independent of iopattern. Is it actually possible to have verification independent of iopattern as is? Or would that require a lot of changes?

safl (changes requested):
Let me start by saying that this is awesome. Then, a couple of things inline.

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+	*seed = *seed * 1103515245 + 12345;
+	return (int)((*seed >> 16) & 0x7fff);
+}
+#endif
+
+static bool
+is_power_of_2(uint32_t x)
+{
+	return x > 0 && (x & (x - 1)) == 0;
+}
```
karlowich: We already have a function like this in `libxnvme_util.h`: `xnvme_is_pow2()`

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+{
+	return (uint64_t)ts->tv_sec * 1000000000ULL + (uint64_t)ts->tv_nsec;
+}
+
+static double
+elapsed_secs(struct timespec *start, struct timespec *end)
+{
+	return (double)(timespec_to_ns(end) - timespec_to_ns(start)) / 1e9;
+}
+
```
karlowich: Similarly, we have some time related functions in `libxnvme_util.h` perhaps you can use those instead?
Otherwise, if something is missing, I think it would be relevant to add what you need there.

Makefile
NaddiNadja: I was also unsure whether we need these in the makefile, I just wanted the commands written somewhere, but then I added docs later, so maybe I can completely remove the makefile commit?

lib/xnvme_cli.c
NaddiNadja: I used the IO patterns and naming from bdevperf, but later actually removed verify, because I have `xnvmeperf run` and `xnvmeperf verify` subcommands instead, so it's a mistake that verify is written in here as an iopattern

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+			err = submit_io(job, ctx);
+			if (err) {
+				xnvme_queue_put_cmd_ctx(job->queue, ctx);
+				break;
+			}
+		}
+	}
+
+	clock_gettime(CLOCK_MONOTONIC, &start);
```
karlowich: Should the initial filling of the queues be part of the time taking?

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+	struct xnvmeperf_thread *threads;
+	struct xnvme_dev **devs;
+	pthread_t *tids;
+	pthread_t *close_tids;
+	int err = 0;
+
+	// Pre-open all devices, as they can only be opened once.
+	devs = calloc(args->ndevs, sizeof(*devs));
+	if (!devs) {
+		err = -ENOMEM;
```
karlowich: You can set `err = errno` to catch if calloc fails for a different reason

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+			}
+			free(devs);
+			return err;
+		}
+	}
+
+	threads = calloc(args->ncpus, sizeof(*threads));
+	tids = calloc(args->ncpus, sizeof(*tids));
+	if (!threads || !tids) {
+		err = -ENOMEM;
```
karlowich, line 541: Same here, set `err = errno`

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+		struct xnvme_dev *dev;
+		struct xnvmeperf_job job = {0};
+		void *write_buf = NULL, *read_buf = NULL, *expect_buf = NULL;
+		uint64_t mismatches = 0;
+
+		dev = xnvme_dev_open(args->dev_uris[d], &args->opts);
+		if (!dev) {
+			fprintf(stderr, "Failed: xnvme_dev_open(%s): %s\n", args->dev_uris[d],
+				strerror(errno));
+			err = -errno;
```
karlowich, line 673: Set this before calling fprintf, as that can potentially overwrite `errno` (although unlikely)

tools/xnvmeperf.c
```diff
@@ -0,0 +1,975 @@
...
+		if (err) {
+			xnvme_cli_perr("Failed: setup_job()", err);
+			goto next_dev;
+		}
+
+		write_buf = xnvme_buf_alloc(dev, args->iosize);
+		read_buf = xnvme_buf_alloc(dev, args->iosize);
+		expect_buf = malloc(args->iosize);
+		if (!write_buf || !read_buf || !expect_buf) {
+			err = -ENOMEM;
```
karlowich: Set this to errno

karlowich (changes requested):
Great, here's second round of comments

tools/xnvmeperf.c
karlowich: Okay, I agree.

karlowich: I would still like to replace `memcmp` with `xnvme_diff`

Makefile
karlowich: I agree, let's just remove them from the Makefile

tools/xnvmeperf.c
NaddiNadja: ah yes of course!, i'll fix that part

Makefile
NaddiNadja: will do will do 👍

tools/xnvmeperf.c
NaddiNadja: omg, amazing

tools/xnvmeperf.c
NaddiNadja: great! much nicer

tools/xnvmeperf.c
NaddiNadja: yea, I feel like it should, I'll move it up and test it out

tools/xnvmeperf.c
NaddiNadja: I was unsure here, because if the first calloc sets errno to ENOMEM, but then the next calloc succeeds (is this even possible?), is errno then still set correctly?

karlowich: Well if you want to be on the safe side, you can't collapse the error handling.

NaddiNadja: yea but that's why like ..... isn't it simpler to just set it to the only errno that calloc sets?

meson.build
NaddiNadja: For xNVMe v0.7.3, we have it in the default_options (see [here](https://github.com/xnvme/xnvme/blob/v0.7.3/meson.build)), but commented out, so I assumed that was the reason it was not in the default options. Might have been fixed since we released xNVMe v0.7.3 though, but that was the reason it's been added further down in the file with the windows check.

tools/xnvmeperf.c
karlowich: For sure

NaddiNadja:
> A couple of things that would be great to add:
> 
>* A little intro to the use here similar to https://xnvme.io/tools/fio/index.html
>* Simplify usage; boil it down to cli-arguments (drop json/toml for now), this might require expanding the "xnvme_cli" module, but that would be great.
>* A test exercising the tool with using all available backends (note we have a bunch of changes coming in this area on backend-instrumentation on (#623)
> 
> Down the line:
> 
>* Extend the latency testing with numbers using xnvmeperf

I have added the intro, and removed the configuration file. Due to the changes in #623, I would suggest that we create a Github issue on extending the tests with one using all available backends, and extending the latency testing with xnvmeperf too - just so the scope of this PR doesn't expand much further. :)

meson.build
NaddiNadja: Okay, I just tried adding it, and it seems that there are still some errors on Windows (see [this run](https://github.com/xnvme/xnvme/actions/runs/22434213899/job/64960062494?pr=636)), so I will leave it as it is now. :)

tools/xnvmeperf.c
```diff
@@ -0,0 +1,976 @@
...
+		xnvme_cli_perr("Failed: calloc() for devs", err);
+		return err;
+	}
+
+	for (int i = 0; i < args->ndevs; i++) {
+		devs[i] = xnvme_dev_open(args->dev_uris[i], &args->opts);
+		if (!devs[i]) {
+			fprintf(stderr, "Failed: xnvme_dev_open(%s): %s\n", args->dev_uris[i],
+				strerror(errno));
+			err = -errno;
```
karlowich, line 518: This should come before `fprintf`

karlowich (changes requested):

tools/xnvmeperf.c
NaddiNadja: thought i had caught them all

karlowich (approved):
LGTM

safl (approved):
Truly great!

PR #638, opened 2026-02-26
Introduce CUDA memory compatibility

karlowich:
These commits changes the functions in the `xnvme_buf` API to be compatible with both device and host memory.
That is, if CUDA is available on the system and the meson feature is not disabled.

In preparation for this, I am introducing a new function `xnvme_buf_copy()` which acts like `memcpy()`, as well as changing our tools, tests, examples to use the `xnvme_buf` API instead of the C standard library equivalents.

Since, I'm already touching the `xnvme_buf_fill()` code path, I decided to give it a little extra love.

### DIFF
--- a/lib/xnvme_buf.c
+++ b/lib/xnvme_buf.c
@@ -8,6 +8,7 @@
 #include <libxnvme.h>
 #include <xnvme_dev.h>
 #include <xnvme_be.h>
+#include <xnvme_host_buf.h>
 
 void *
 xnvme_buf_virt_alloc(size_t alignment, size_t nbytes)
@@ -90,15 +91,13 @@ xnvme_buf_free(const struct xnvme_dev *dev, void *buf)
 int
 xnvme_buf_clear(void *buf, size_t nbytes)
 {
-	memset(buf, 0, nbytes);
-	return 0;
+	return xnvme_host_buf_clear(buf, nbytes);
 }
 
 int
 xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
 {
-	memcpy(dst, src, nbytes);
-	return 0;
+	return xnvme_host_buf_memcpy(dst, src, nbytes);
 }
 
 /**
@@ -182,84 +181,20 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
 int
 xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
 {
-	uint8_t *cbuf = buf;
-
-	if (strlen(content) == 1) {
-		memset(cbuf, content[0], nbytes);
-		return 0;
-	}
-
-	if (!strncmp(content, "anum", 4)) {
-		for (size_t i = 0; i < nbytes; ++i) {
-			cbuf[i] = (i % 26) + 65;
-		}
-
-		return 0;
-	}
-
-	if (!strncmp(content, "rand-t", 6)) {
-		srand(time(NULL));
-		for (size_t i = 0; i < nbytes; ++i) {
-			cbuf[i] = (rand() % 26) + 65;
-		}
-
-		return 0;
-	}
-
-	if (!strncmp(content, "rand-k", 6)) {
-		srand(0);
-		for (size_t i = 0; i < nbytes; ++i) {
-			cbuf[i] = (rand() % 26) + 65;
-		}
-
-		return 0;
-	}
-
-	if (!strncmp(content, "zero", 4)) {
-		return xnvme_buf_clear(buf, nbytes);
-	}
-
-	return xnvme_buf_from_file(buf, nbytes, content);
+	return xnvme_host_buf_fill(buf, nbytes, content);
 }
 
 int
 xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff)
 {
-	const uint8_t *exp = expected;
-	const uint8_t *act = actual;
-
-	for (size_t i = 0; i < nbytes; ++i) {
-		if (exp[i] == act[i]) {
-			continue;
-		}
-
-		*diff += 1;
-	}
-
-	return 0;
+	return xnvme_host_buf_diff(expected, actual, nbytes, diff, false);
 }
 
 int
 xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int XNVME_UNUSED(opts))
 {
-	const uint8_t *exp = expected;
-	const uint8_t *act = actual;
 	size_t diff = 0;
-
-	printf("comparison:\n");
-	printf("  diffs:\n");
-	for (size_t i = 0; i < nbytes; ++i) {
-		if (exp[i] == act[i]) {
-			continue;
-		}
-
-		++diff;
-		printf("    - {byte: '%06zu', expected: 0x%" PRIx8 ", actual: 0x%" PRIx8 ")\n", i,
-		       exp[i], act[i]);
-	}
-	printf("  nbytes: %zu\n", nbytes);
-	printf("  nbytes_diff: %zu\n", diff);
-	return 0;
+	return xnvme_host_buf_diff(expected, actual, nbytes, &diff, true);
 }
 
 int
--- a/lib/xnvme_buf.c
+++ b/lib/xnvme_buf.c
@@ -9,6 +9,7 @@
 #include <xnvme_dev.h>
 #include <xnvme_be.h>
 #include <xnvme_host_buf.h>
+#include <xnvme_cuda_buf.h>
 
 void *
 xnvme_buf_virt_alloc(size_t alignment, size_t nbytes)
@@ -91,12 +92,20 @@ xnvme_buf_free(const struct xnvme_dev *dev, void *buf)
 int
 xnvme_buf_clear(void *buf, size_t nbytes)
 {
+	if (xnvme_buf_is_cuda(buf)) {
+		return xnvme_cuda_buf_clear(buf, nbytes);
+	}
+
 	return xnvme_host_buf_clear(buf, nbytes);
 }
 
 int
 xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
 {
+	if (xnvme_buf_is_cuda(dst) || xnvme_buf_is_cuda(src)) {
+		return xnvme_cuda_buf_memcpy(dst, src, nbytes);
+	}
+
 	return xnvme_host_buf_memcpy(dst, src, nbytes);
 }
 
@@ -181,12 +190,20 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
 int
 xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
 {
+	if (xnvme_buf_is_cuda(buf)) {
+		return xnvme_cuda_buf_fill(buf, nbytes, content);
+	}
+
 	return xnvme_host_buf_fill(buf, nbytes, content);
 }
 
 int
 xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff)
 {
+	if (xnvme_buf_is_cuda(expected) || xnvme_buf_is_cuda(actual)) {
+		return xnvme_cuda_buf_diff(expected, actual, nbytes, diff, false);
+	}
+
 	return xnvme_host_buf_diff(expected, actual, nbytes, diff, false);
 }
 
@@ -194,6 +211,10 @@ int
 xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int XNVME_UNUSED(opts))
 {
 	size_t diff = 0;
+	if (xnvme_buf_is_cuda(expected) || xnvme_buf_is_cuda(actual)) {
+		return xnvme_cuda_buf_diff(expected, actual, nbytes, &diff, true);
+	}
+
 	return xnvme_host_buf_diff(expected, actual, nbytes, &diff, true);
 }
 
--- a/lib/xnvme_cuda_buf.c
+++ b/lib/xnvme_cuda_buf.c
@@ -1,0 +1,175 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifdef XNVME_CUDA_ENABLED
+#include <cuda_runtime.h>
+#include <libxnvme.h>
+#include <xnvme_host_buf.h>
+#include <errno.h>
+
+bool
+xnvme_buf_is_cuda(const void *buf)
+{
+	cudaError_t err;
+	struct cudaPointerAttributes attr;
+
+	err = cudaPointerGetAttributes(&attr, buf);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaPointerGetAttributes(buf), err: %s",
+			    cudaGetErrorString(err));
+		return false; // If it fails we assume it is not a CUDA buffer
+	}
+
+	return attr.type == cudaMemoryTypeDevice;
+}
+
+int
+xnvme_cuda_buf_clear(void *buf, size_t nbytes)
+{
+	cudaError_t err;
+
+	err = cudaMemset(buf, 0, nbytes);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaMemset, err: %s", cudaGetErrorString(err));
+		return -EIO;
+	}
+
+	return 0;
+}
+
+int
+xnvme_cuda_buf_memcpy(void *dst, const void *src, size_t nbytes)
+{
+	cudaError_t err;
+
+	err = cudaMemcpy(dst, src, nbytes, cudaMemcpyDefault);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaMemcpy, err: %s", cudaGetErrorString(err));
+		return -EIO;
+	}
+	return 0;
+}
+
+int
+xnvme_cuda_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff,
+		    bool print)
+{
+	void *exp, *act;
+	cudaError_t cu_err;
+	int err;
+
+	exp = malloc(nbytes);
+	if (!exp) {
+		err = -errno;
+		XNVME_DEBUG("FAILED: malloc(exp), errno: %d", err);
+		return err;
+	}
+
+	act = malloc(nbytes);
+	if (!act) {
+		err = -errno;
+		XNVME_DEBUG("FAILED: malloc(act), errno: %d", err);
+		goto exit;
+	}
+
+	cu_err = cudaMemcpy(exp, expected, nbytes, cudaMemcpyDefault);
+	if (cu_err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaMemcpy(expected), err: %s", cudaGetErrorString(cu_err));
+		err = -EIO;
+		goto exit;
+	}
+
+	cu_err = cudaMemcpy(act, actual, nbytes, cudaMemcpyDefault);
+	if (cu_err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaMemcpy(actual), err: %s", cudaGetErrorString(cu_err));
+		err = -EIO;
+		goto exit;
+	}
+
+	err = xnvme_host_buf_diff(exp, act, nbytes, diff, print);
+	if (err) {
+		XNVME_DEBUG("FAILED: xnvme_host_buf_diff(), err: %d", err);
+		goto exit;
+	}
+
+exit:
+	free(exp);
+	free(act);
+
+	return err;
+}
+
+int
+xnvme_cuda_buf_fill(void *buf, size_t nbytes, const char *content)
+{
+	cudaError_t cu_err;
+	int err;
+	uint8_t *cbuf;
+
+	cbuf = malloc(nbytes);
+	if (!cbuf) {
+		err = -errno;
+		XNVME_DEBUG("FAILED: malloc(cbuf), errno: %d", err);
+		return err;
+	}
+
+	err = xnvme_host_buf_fill(cbuf, nbytes, content);
+	if (err) {
+		XNVME_DEBUG("FAILED: xnvme_host_buf_fill(%s), err: %d", content, err);
+		goto exit;
+	}
+
+	cu_err = cudaMemcpy(buf, cbuf, nbytes, cudaMemcpyDefault);
+	if (cu_err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaMemcpy, err: %s", cudaGetErrorString(cu_err));
+		err = -EIO;
+		goto exit;
+	}
+exit:
+	free(cbuf);
+	return err;
+}
+#else
+
+#include <libxnvme.h>
+#include <errno.h>
+
+bool
+xnvme_buf_is_cuda(const void *XNVME_UNUSED(buf))
+{
+	return false;
+}
+
+int
+xnvme_cuda_buf_clear(void *XNVME_UNUSED(buf), size_t XNVME_UNUSED(nbytes))
+{
+	XNVME_DEBUG("FAILED: xNVMe was compiled without CUDA");
+	return -ENOTSUP;
+}
+
+int
+xnvme_cuda_buf_memcpy(void *XNVME_UNUSED(dst), const void *XNVME_UNUSED(src),
+		      size_t XNVME_UNUSED(nbytes))
+{
+	XNVME_DEBUG("FAILED: xNVMe was compiled without CUDA");
+	return -ENOTSUP;
+}
+
+int
+xnvme_cuda_buf_diff(const void *XNVME_UNUSED(expected), const void *XNVME_UNUSED(actual),
+		    size_t XNVME_UNUSED(nbytes), size_t *XNVME_UNUSED(diff),
+		    bool XNVME_UNUSED(print))
+{
+	XNVME_DEBUG("FAILED: xNVMe was compiled without CUDA");
+	return -ENOTSUP;
+}
+
+int
+xnvme_cuda_buf_fill(void *XNVME_UNUSED(buf), size_t XNVME_UNUSED(nbytes),
+		    const char *XNVME_UNUSED(content))
+{
+	XNVME_DEBUG("FAILED: xNVMe was compiled without CUDA");
+	return -ENOTSUP;
+}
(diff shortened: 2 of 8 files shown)


NaddiNadja (dismissed):
lgtm ⭐

safl (commented):
This is exciting work; however, I have a few concerns.

As I understand it, when xNVMe is linked with CUDA, the buffer helpers `xnvme_buf_*()` are unconditionally mapped to the CUDA variety `xnvme_cuda_buf_*()` at compile time.

One concern is that the rest of xNVMe uses runtime dispatch. Ideally, these helpers should delegate to the CUDA variants only when operating on CUDA buffers, rather than being statically bound at compile time.

Another concern is whether linking with CUDA limits the buffer helpers to CUDA buffers only. For example, when linked with CUDA, must users allocate all buffers using CUDA-managed memory?

karlowich:
@safl 

All of the `xnvme_cuda_buf_*()` functions are compatible with both device and host memory. The problem with runtime delegation is that only CUDA knows if it is a runtime buffer(1). I thought about naming the file and functions `xnvme_cuda_compat_buf_*()`, this would be more accurate, but also a bit of a mouthful. Alternatively, you could move the delegation code into `xnvme_buf.c`, but then you would have a significant amount of code behind macros. I like that the macros only switch which function to call.

As it stands right now, calling the `xnvme_buf_*()` functions is safe regardless of how you allocated the buffer (except for HIP/ROCm of course). As such, I think it just a question of naming or perhaps writing some docstrings (which I didn't do since these functions are only internal). 

(1) This is not strictly true, I think we could look up something about the address ranges used by the host and the device. However, this would have to be done for every single platform.

safl:
> @safl
> 
> All of the `xnvme_cuda_buf_*()` functions are compatible with both device and host memory. The problem with runtime delegation is that only CUDA knows if it is a runtime buffer(1).

With runtime delegation i meant, the decision about which function implements the behavior is delegated at runtime. Not that the decision is delegated to the CUDA runtime.

karlowich:
@safl I believe I have addressed your concerns now.

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+#include <xnvme_buf.h>
+#include <errno.h>
+
+bool
+xnvme_buf_is_cuda(void *buf)
+{
+	cudaError_t err;
+	struct cudaPointerAttributes attr;
+
+	cudaPointerGetAttributes(&attr, buf);
```
safl: Should this be ``err = cudaPointerGetAttributes(&attr, buf);``?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+	free((void *)exp);
+	free((void *)act);
+
+	return diff;
+}
+
+int
+xnvme_cuda_buf_fill(void *buf, size_t nbytes, const char *content)
+{
+	cudaError_t err = 0;
```
safl: Why zero-initialize her?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+	uint8_t *cbuf;
+
+	cbuf = malloc(nbytes);
```
safl, line 110: This seems to lack error-handling?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+		err = errno;
+		XNVME_DEBUG("FAILED: malloc(exp), errno: %d", err);
+		return err;
+	}
+
+	act = malloc(nbytes);
+	if (!act) {
+		err = errno;
+		XNVME_DEBUG("FAILED: malloc(act), errno: %d", err);
+		free((void *)act);
```
safl: This should de-allocate ``exp``, not ``act``?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+		free((void *)exp);
+		free((void *)act);
+		return err;
+	}
+
+	diff = _xnvme_buf_diff(exp, act, nbytes, print);
+
+	free((void *)exp);
+	free((void *)act);
+
```
safl, line 99: You could add an 'exit' label here to avoided the repeated error-handling when checking ``cudaMemCpy``.

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+	struct cudaPointerAttributes exp_attr, act_attr;
+	const uint8_t *exp, *act;
+	cudaError_t err;
+	size_t diff = 0;
+
+	err = cudaPointerGetAttributes(&exp_attr, expected);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaPointerGetAttributes(expected), err: %s",
+			    cudaGetErrorString(err));
+		return err;
```
safl: Is ``cudaError_t`` compatible with size_t / the expected behavior on the error-path? Perhaps the cuda-helper needs a different signature than ``xnvme_buf_diff()`` to do error-handling by the caller?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	err = cudaPointerGetAttributes(&act_attr, expected);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaPointerGetAttributes(actual), err: %s",
+			    cudaGetErrorString(err));
+		return err;
```
safl: Is ``cudaError_t`` compatible with size_t / the expected behavior on the error-path? Perhaps the cuda-helper needs a different signature than ``xnvme_buf_diff()`` to do error-handling by the caller?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,214 @@
...
+	if (exp_attr.type != act_attr.type) {
+		XNVME_DEBUG("FAILED: memory type for expected and actual doesn't match");
+		return -EINVAL;
+	}
+
+	exp = malloc(nbytes);
+	if (!exp) {
+		err = errno;
+		XNVME_DEBUG("FAILED: malloc(exp), errno: %d", err);
+		return err;
```
safl, line 66: This won't work... i mean if errno is ``ENOMEM`` then this function returns "12 bytes differed!"

safl (changes requested):
As you have mentioned as well, then this could be a smaller commit, e.g. the first 7 are very easily reviewed as their is not really much to address, only the last 2 commits carrying the changes that are worth discussing requiring re-iteration.
I have some stuff inline, and yeah, this is more along the lines of what I was looking for.

lib/xnvme_cuda_buf.c
karlowich: Absolutely

lib/xnvme_cuda_buf.c
karlowich: Definitely

lib/xnvme_cuda_buf.c
karlowich: You're right

lib/xnvme_cuda_buf.c
karlowich: Right, that's an oversight

lib/xnvme_cuda_buf.c
karlowich: Yes

karlowich:
@safl Thanks for the review. I resolved your comments as this PR will change direction now.

(As discussed) I will refactor our xnvme_buf API to use correct error handling that is changing the signature of the following functions:
* ```C
  *OLD* void * xnvme_buf_clear(void *buf, size_t nbytes) 
  *NEW* int xnvme_buf_clear(void *buf, size_t nbytes)
* ```C
  *OLD* void * xnvme_buf_copy(void *dst, void *src, size_t nbytes)
  *NEW* int xnvme_buf_copy(void *dst, void *src, size_t nbytes)
* ```C
  *OLD* size_t xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes)
  *NEW* int xnvme_buf_diff(const void *expected, const void *actual, size_t *diff, size_t nbytes)

The question is if I should give `xnvme_buf_diff_pr()` the same treatment or simply remove this function and add an argument to `xnvme_buf_diff()` controlling whether to print.

@safl @NaddiNadja What are your thoughts on this?

NaddiNadja:
> The question is if I should give `xnvme_buf_diff_pr()` the same treatment or simply remove this function and add an argument to `xnvme_buf_diff()` controlling whether to print.

I feel like the two situations, (a) check the diff and (b) print the diff to console, are very different from a user's perspective, so in the public API, I would keep these specifically separate. It could return an int (err), but it could also just print it to the console, given that it's a printing-function. I don't know what I prefer.

karlowich:
The refactor has been moved here: https://github.com/xnvme/xnvme/pull/643

karlowich:
Broke the copy part into a separate PR #644

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,218 @@
...
+
+	return false;
+}
+
+int
+xnvme_cuda_buf_clear(void *buf, size_t nbytes)
+{
+	cudaError_t err;
+
+	err = cudaMemset(buf, 0, sizeof(nbytes));
```
safl: This should be:
``err = cudaMemset(buf, 0, nbytes);``
?

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,218 @@
...
+	cudaError_t err;
+
+	err = cudaPointerGetAttributes(&exp_attr, expected);
+	if (err != cudaSuccess) {
+		XNVME_DEBUG("FAILED: cudaPointerGetAttributes(expected), err: %s",
+			    cudaGetErrorString(err));
+		return err;
+	}
+
+	err = cudaPointerGetAttributes(&act_attr, expected);
```
safl: I think this should be:
```
err = cudaPointerGetAttributes(&act_attr, actual);
```

otherwise... then further down (line 108) the check ``exp_attr.type != act_attr.type`` will always be true.

safl (changes requested):
A general issue is that the return values from the CUDA versions of the buffer helpers implemented here return ``cudaError_t`` (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038). Since the calling convention for the functions are to return negativ ``errno``, then it becomes confusing that suddenly they are returning positive values, values the user have no chance of knowing what means.

Perhaps just do a wildcard conversion to ``-EIO`` or something along those lines... this needs a bit of consideration.

Other specific things, see inline.

lib/xnvme_buf.c
```diff
@@ -87,18 +89,57 @@ xnvme_buf_free(const struct xnvme_dev *dev, void *buf)
...
 int
 xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
 {
-	memcpy(dst, src, nbytes);
-	return 0;
+	int dst_type = xnvme_buf_get_type(dst);
+	int src_type = xnvme_buf_get_type(src);
+
+	switch (dst_type) {
+	case XNVME_BUF_HOST:
```
safl: How about this instead:

```c
int
xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
{
  if (xnvme_buf_is_cuda(dst) || xnvme_buf_is_cuda(src)) {
    return xnvme_cuda_buf_memcpy(dst, src, nbytes);
  }
  memcpy(dst, src, nbytes);
  return 0;
}
```

and then leave the convoluted logic to the CUDA implementation.

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,218 @@
...
+		return err;
+	}
+
+	return 0;
+}
+
+int
+xnvme_cuda_buf_memcpy(void *dst, const void *src, size_t nbytes)
+{
+	struct cudaPointerAttributes src_attr, dst_attr;
```
safl: Could this be reduced to:

```c
int
xnvme_cuda_buf_memcpy(void *dst, const void *src, size_t nbytes)
{
  cudaError_t err;

  err = cudaMemcpy(dst, src, nbytes, cudaMemcpyDefault);
  if (err != cudaSuccess) {
    XNVME_DEBUG("FAILED: cudaMemcpy, err: %s", cudaGetErrorString(err));
    return -EIO;
  }

  return 0;
}
```

Form the docs it looks like it cudaMemcpy (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gc263dbe6574220cc776b45438fc351e8) does all the logic here expressed, e.g.

 "Passing **cudaMemcpyDefault** is recommended, in which case the type of transfer is inferred from the pointer values."

Or am i missing something fundamental here?

lib/xnvme_buf.c
```diff
@@ -180,7 +221,37 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
 }
 
 int
-xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
+_xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff, bool print)
```
safl: How about this signature instead.

```c
static int
host_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff, bool print);
```

Then ``xnvme_buf_diff()`` does the dispatch and the implementations it dispatches to follow the same naming pattern.

Since the host-implementation are not needed anywhere else, then they can be static functions instead of using prefixing with ``_``.

Same for _xnvme_buf_fille, do ``static int host_buf_fill(...)`` instead

safl (commented):
After the first review i found some additional thing, it is a bit scattered, i apologize for that.

lib/xnvme_cuda_buf.c
karlowich: Definitely

lib/xnvme_cuda_buf.c
karlowich: Right of course

lib/xnvme_cuda_buf.c
karlowich: You're completely right, I can't recall why I decided the verbose version was preferred.

lib/xnvme_buf.c
karlowich: I had this originally. I decided to change to the nested switch statements to avoid redoing the code if/when more memory types are introduced. I don't know what is best.

lib/xnvme_buf.c
karlowich: I am using the `_xnvme_buf_diff()` and `_xnvme_buf_fill()` from the CUDA code to avoid reimplementing the logic.

lib/xnvme_cuda_buf.c
karlowich: Looking at this code path - perhaps we don't have to fail if the memory types aren't matching? Then we could use `cudaMemcpyDefault` and avoid the `cudaPointerGetAttributes` dance.

karlowich:
> A general issue is that the return values from the CUDA versions of the buffer helpers implemented here return cudaError_t

@safl Yes, it is unclear to me how to go about this in the best way. As a start, I changed them to be returned as negative. Unfortunately, there's a lot of overlap with errno. Although all the errors I have ever gotten have been way outside the range of errno.

lib/xnvme_cuda_buf.c
karlowich: I think the simpler solution is nicer.

lib/xnvme_buf.c
karlowich: That being said, it might still be nicer with the `host_*` naming scheme. Even if they're still being called from inside `xnvme_buf_cuda.c`.

safl (changes requested):
The negative ``errno`` calling convention is still violated here.
For example, the CUDA function now return negative ``cudaError_t``.
Thus, when an error is returned like ``cudaErrorMemoryAllocation(2)`` then what the user will see is ``errno(2)`` aka ``ENOENT 2 No such file or directory``.

Thus, convert these to some generic errno like ``EIO 5 Input/output error``.

Replacing the "_" with "host_..." would be preferable.

lib/xnvme_cuda_buf.c
karlowich: Because of this line from the documentation:
"However, cudaMemcpyDefault is only allowed on systems that support unified virtual addressing."
I don't know if this is something we should protect against

safl: The other CUDA/GPU requirements of bar-space mapping, dma-buf support, GSP and open-source-kernel modules puts it in a territory where we can safely assume that UVA is also supported. If I recall correctly then UVA got introduce some 10+ years ago.

safl: Having said that, then one could go about and devise some nice CUDA compile-feature-checks in Meson to ensure that all the CUDA facing APIs that we utilize are tested for and produce nice error-messages at compile-time. However, we can add that kind of polish later.

lib/xnvme_buf.c
safl: Would be great to drop that ``_`` prefix and replacing it either with ``static`` where possible or the ``host_`` prefix.

lib/xnvme_cuda_buf.c
karlowich: Sounds good

lib/xnvme_buf.c
karlowich: Yes, I am cooking something up in this direction

safl:
> > A general issue is that the return values from the CUDA versions of the buffer helpers implemented here return cudaError_t
> 
> @safl Yes, it is unclear to me how to go about this in the best way. As a start, I changed them to be returned as negative. Unfortunately, there's a lot of overlap with errno. Although all the errors I have ever gotten have been way outside the range of errno.

I made another note on this in the most recent review. Let's just go with ``return -EIO;``

karlowich:
@safl I added a new commit which introduces a separate `xnvme_host_buf` API this way the roles of `xnvme_buf` and `xnvme_cuda_buf` hopefully become clearer.

safl (commented):
nit: the commit message mentions ``buf_copy`` which is now ``buf_memcpy``.

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,180 @@
...
+		return -EIO;
+	}
+	return 0;
+}
+
+int
+xnvme_cuda_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff,
+		    bool print)
+{
+	struct cudaPointerAttributes exp_attr, act_attr;
```
safl: These are unused.

safl (commented):
See inline... also... ~~i am brewing up something for the CUDA version check... [edit]~~

Future work:

- https://github.com/xnvme/xnvme/issues/647
- https://github.com/xnvme/xnvme/issues/648

include/xnvme_buf.h
```diff
@@ -0,0 +1,13 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef __INTERNAL_XNVME_BUF_H
+#define __INTERNAL_XNVME_BUF_H
+
+enum xnvme_buf_type {
```
safl: I thought you dropped this...

lib/xnvme_buf.c
```diff
@@ -88,16 +90,55 @@ xnvme_buf_free(const struct xnvme_dev *dev, void *buf)
...
+		return xnvme_host_buf_clear(buf, nbytes);
+	case XNVME_BUF_CUDA:
+		return xnvme_cuda_buf_clear(buf, nbytes);
+	default:
+		return -ENOTSUP;
+	}
 }
 
 int
 xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
```
safl, line 103:
```c
int
xnvme_buf_memcpy(void *dst, const void *src, size_t nbytes)
{
  if (xnvme_buf_is_cuda(dst) || xnvme_buf_is_cuda(src)) {
    return xnvme_cuda_buf_memcpy(dst, src, nbytes);
  }

  return xnvme_host_buf_memcpy(dst, src, nbytes);
}
```

lib/xnvme_buf.c
```diff
@@ -88,16 +90,55 @@ xnvme_buf_free(const struct xnvme_dev *dev, void *buf)
...
+xnvme_buf_get_type(const void *buf)
+{
+	if (xnvme_buf_is_cuda(buf)) {
+		return XNVME_BUF_CUDA;
+	}
+	return XNVME_BUF_HOST;
+}
+
 int
 xnvme_buf_clear(void *buf, size_t nbytes)
```
safl, line 93:
```c
int
xnvme_buf_clear(void *buf, size_t nbytes)
{
  if (xnvme_buf_is_cuda(buf)) {
    return xnvme_cuda_buf_clear(buf, nbytes);
  }
  return xnvme_host_buf_clear(buf, nbytes);
}
```

lib/xnvme_buf.c
```diff
@@ -179,22 +220,73 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
 }
 
 int
-xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
+xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff)
```
safl:
```c
int
xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff)
{
  if (xnvme_buf_is_cuda(expected) || xnvme_buf_is_cuda(actual)) {
    return xnvme_cuda_buf_diff(expected, actual, nbytes, diff, false);
  }
  return xnvme_host_buf_diff(expected, actual, nbytes, diff, false);
}
```

lib/xnvme_buf.c
```diff
@@ -179,22 +220,73 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
...
+			return -ENOTSUP;
+		}
+	default:
+		return -ENOTSUP;
+	}
 }
 
 int
-xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff)
+xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int XNVME_UNUSED(opts))
```
safl:
```c
int
xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int XNVME_UNUSED(opts))
{
  size_t diff = 0;
  if (xnvme_buf_is_cuda(expected) || xnvme_buf_is_cuda(actual)) {
    return xnvme_cuda_buf_diff(expected, actual, nbytes, &diff, true);
  }
  return xnvme_host_buf_diff(expected, actual, nbytes, &diff, true);
}
```

lib/xnvme_buf.c
```diff
@@ -179,22 +220,73 @@ xnvme_buf_to_file(void *buf, size_t nbytes, const char *path)
...
+			return -ENOTSUP;
+		}
+	default:
+		return -ENOTSUP;
+	}
 }
 
 int
-xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int XNVME_UNUSED(opts))
+xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
```
safl:
```c
int
xnvme_buf_fill(void *buf, size_t nbytes, const char *content)
{
  if (xnvme_buf_is_cuda(buf)) {
    return xnvme_cuda_buf_fill(buf, nbytes, content);
  }
  return xnvme_host_buf_fill(buf, nbytes, content);
}
```

safl (changes requested):
I thought that when you said "simpler is better" that you went along the way of the suggested changes. I listed them again, I hope you agree that this is more readable and simpler.

With these suggestions we get to drop enum, drop ``xnvme_buf_get_type()``, drops ``include/xnvme_buf.h`` entirely. Simple if/else, host is the default path.

lib/xnvme_cuda_buf.c
```diff
@@ -0,0 +1,180 @@
...
+
+#ifdef XNVME_CUDA_ENABLED
+#include <cuda_runtime.h>
+#include <libxnvme.h>
+#include <xnvme_host_buf.h>
+#include <errno.h>
+
+bool
+xnvme_buf_is_cuda(const void *buf)
+{
```
safl, line 13: Could be simpler...

```c
bool
xnvme_buf_is_cuda(const void *buf)
{
  struct cudaPointerAttributes attr;
  cudaError_t err;

  err = cudaPointerGetAttributes(&attr, buf);
  if (err != cudaSuccess) {
    XNVME_DEBUG("FAILED: cudaPointerGetAttributes(buf), err: %s", cudaGetErrorString(err));
    return false;
  }
  return attr.type == cudaMemoryTypeDevice;
}
```

safl: It is this part that hurt my eyes:

```c
if (attr.type == cudaMemoryTypeDevice) {
  return true;
}

return false;
```

karlowich:
> Future work: #647 #648

Nice.

I will address your concerns. The enum and memory types were added to handle more memory types in the future, but I guess we can refactor as needed then.

lib/xnvme_cuda_buf.c
karlowich: Yes, that is an atrocity

lib/xnvme_cuda_buf.c
karlowich: Yes

include/xnvme_buf.h
karlowich: Dropped it now

lib/xnvme_buf.c
karlowich: yes

lib/xnvme_buf.c
karlowich: yes

lib/xnvme_buf.c
karlowich: yes

lib/xnvme_buf.c
karlowich: yes

lib/xnvme_buf.c
karlowich: yes

karlowich:
@safl Now it should be squeaky clean. I apologize for the **messy** code

safl (approved):
lgtm.

NaddiNadja (approved):
looks good to me too now :) also, looks like the remaining comments have all been correctly addressed, so they can just be resolved.

PR #641, opened 2026-03-02
Simple fixes to xnvme/buf

karlowich:
These are changes that are mostly unrelated to #638. Thus they will be available here as to not muddy the point of the other PR.


NaddiNadja (approved):
Looks good to me, and good to get out in a separate PR :)

safl (approved):
LGTM. For future reference, then it would be nice with a description why this is needed, not just that it is in preparation for something in the future.

karlowich:
Right, I see I skipped right past any justification

PR #643, opened 2026-03-03
Refactor xNVMe BUF API to have error handling

karlowich:
This refactors the following functions to return an error code:
* ```C
  *OLD* void * xnvme_buf_clear(void *buf, size_t nbytes) 
  *NEW* int xnvme_buf_clear(void *buf, size_t nbytes)
* ```C
  *OLD* void * xnvme_buf_copy(void *dst, void *src, size_t nbytes)
  *NEW* int xnvme_buf_copy(void *dst, void *src, size_t nbytes)
* ```C
  *OLD* size_t xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes)
  *NEW* int xnvme_buf_diff(const void *expected, const void *actual, size_t *diff, size_t nbytes)

Additionally, it adds error handling to every path using `xnvme_buf_fill()`.

Currently, none of these functions can fail. However, the intention is that they should handle different memory types in the future. As such, they need proper error handling.

NOTE: this commit `fix(platform): rename variables from ctrl to ctrlr` is unrelated - but I found it too frustrating to ignore

### DIFF
--- a/include/libxnvme_buf.h
+++ b/include/libxnvme_buf.h
@@ -169,9 +169,9 @@ xnvme_buf_fill(void *buf, size_t nbytes, const char *content);
  * @param buf Pointer to the buffer to fill with zeroes
  * @param nbytes Amount of bytes to fill with zeroes in buf
  *
- * @return Returns the first argument.
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
  */
-void *
+int
 xnvme_buf_clear(void *buf, size_t nbytes);
 
 /**
--- a/include/libxnvme_buf.h
+++ b/include/libxnvme_buf.h
@@ -175,16 +175,17 @@ int
 xnvme_buf_clear(void *buf, size_t nbytes);
 
 /**
- * Returns the number of bytes where expected is different from actual
+ * Compare two buffers and count the number of differing bytes.
  *
- * @param expected Pointer to buffer of "expected" content
- * @param actual Pointer to buffer to compare to "expected"
- * @param nbytes Amount of bytes to compare
+ * @param expected Pointer to buffer containing the expected data
+ * @param actual Pointer to buffer to compare against the expected data
+ * @param nbytes Number of bytes to compare
+ * @param diff Output pointer storing the number of differing bytes
  *
- * @return On success, returns number of bytes that differ
+ * @return 0 on success, negative `errno` on error.
  */
-size_t
-xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes);
+int
+xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff);
 
 /**
  * Prints the number and value of bytes where expected is different from actual
@@ -193,8 +194,10 @@ xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes);
  * @param actual Pointer to buffer to compare to "expected"
  * @param nbytes Amount of bytes to compare
  * @param opts printer options, see ::xnvme_pr
+ *
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
  */
-void
+int
 xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int opts);
 
 /**
--- a/lib/xnvme_dev.c
+++ b/lib/xnvme_dev.c
@@ -209,7 +209,12 @@ _dev_idfy_csi(struct xnvme_dev *dev, struct xnvme_spec_idfy *idfy_ns,
 
 	struct xnvme_spec_znd_idfy_ns *zns = (void *)idfy_ns;
 
-	xnvme_buf_clear(idfy_ctrlr, sizeof(*idfy_ctrlr));
+	err = xnvme_buf_clear(idfy_ctrlr, sizeof(*idfy_ctrlr));
+	if (err) {
+		XNVME_DEBUG("FAILED: xnvme_buf_clear(idfy_ctrlr), err: %d", err);
+		return err;
+	}
+
 	ctx = xnvme_cmd_ctx_from_dev(dev);
 	err = xnvme_adm_idfy_ctrlr_csi(&ctx, XNVME_SPEC_CSI_ZONED, idfy_ctrlr);
 	if (err || xnvme_cmd_ctx_cpl_status(&ctx)) {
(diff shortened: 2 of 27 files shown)


lib/xnvme_dev.c
```diff
@@ -280,15 +296,23 @@ _dev_idfy_csi(struct xnvme_dev *dev, struct xnvme_spec_idfy *idfy_ns,
 
 	// Attempt to identify NVM
 
-	xnvme_buf_clear(idfy_ctrlr, sizeof(*idfy_ctrlr));
+	err = xnvme_buf_clear(idfy_ctrlr, sizeof(*idfy_ctrlr));
+	if (err) {
+		XNVME_DEBUG("FAILED: xnvme_buf_clear(idfy_ctrlr), err: %d", err);
+		return err;
+	}
 	ctx = xnvme_cmd_ctx_from_dev(dev);
```
NaddiNadja, line 309: I might be annoying here, but I would want an empty line after an `if` block. This is not the only place this happens, but I didn't want to spam you with comments about this :)

include/libxnvme_buf.h
```diff
@@ -179,23 +179,27 @@ xnvme_buf_clear(void *buf, size_t nbytes);
...
  * @param nbytes Amount of bytes to compare
+ * @param diff Pointer to number of bytes that differ
  * @param opts printer options, see ::xnvme_pr
+ *
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
  */
-void
-xnvme_buf_diff_pr(const void *expected, const void *actual, size_t nbytes, int opts);
+int
+xnvme_buf_diff_pr(const void *expected, const void *actual, size_t *diff, size_t nbytes, int opts);
```
NaddiNadja: Given that he previous print-function didn't return the `size_t` difference, does it need to be part of the parameters here?

NaddiNadja (changes requested):
Only two things here :)

lib/xnvme_dev.c
karlowich: I have fixed this everywhere

include/libxnvme_buf.h
karlowich: Right, I changed the printing version to not take this argument

tools/xnvmeperf.c
```diff
@@ -747,7 +747,11 @@ xnvmeperf_verify(struct xnvmeperf_args *args)
 		for (int i = 0; i < nios; i++) {
 			uint64_t slba;
 
-			xnvme_buf_clear(read_buf, args->iosize);
+			err = xnvme_buf_clear(read_buf, args->iosize);
+			if (err) {
+				fprintf(stderr, "Failed: xnvme_buf_clear() at IO %d, err: %d\n", i,
+					err);
```
safl, line 770: Add a break here?

tests/znd_append.c
```diff
@@ -165,12 +165,20 @@ cmd_verify(struct xnvme_cli *cli)
...
+		if (err) {
+			xnvme_cli_perr("xnvme_buf_diff()", err);
+			goto exit;
+		}
 		if (diff) {
 			xnvme_cli_pinf("verification failed, diff: %zu", diff);
-			xnvme_buf_diff_pr(dbuf, vbuf, buf_nbytes, XNVME_PR_DEF);
+			err = xnvme_buf_diff_pr(dbuf, vbuf, buf_nbytes, XNVME_PR_DEF);
+			if (err) {
+				xnvme_cli_pinf("xnvme_buf_diff_pr()", err);
```
safl: Use ``xnvme_cli_perr`` here?

tools/xnvmeperf.c
```diff
@@ -771,8 +791,19 @@ xnvmeperf_verify(struct xnvmeperf_args *args)
 				xnvme_queue_poke(job.queue, 0);
 			}
 
-			fill_pattern(expect_buf, args->iosize, slba, job.nlb);
-			if (xnvme_buf_diff(expect_buf, read_buf, args->iosize) != 0) {
+			err = fill_pattern(expect_buf, args->iosize, slba, job.nlb);
+			if (err) {
+				fprintf(stderr, "Failed: fill_parttern() at IO %d, err: %d\n", i,
```
safl: ``fill_parttern --> fill_pattern``

include/libxnvme_buf.h
```diff
@@ -169,22 +169,23 @@ xnvme_buf_fill(void *buf, size_t nbytes, const char *content);
...
+ * @param diff Pointer to number of bytes that differ
  * @param nbytes Amount of bytes to compare
  *
- * @return On success, returns number of bytes that differ
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
  */
-size_t
-xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes);
+int
+xnvme_buf_diff(const void *expected, const void *actual, size_t *diff, size_t nbytes);
```
safl: It would be great to have the **output var** as the last variable. E.g. inputs first, then outputs.
E.g. ``int xnvme_buf_diff(const void *expected, const void *actual, size_t nbytes, size_t *diff);``

include/libxnvme_buf.h
```diff
@@ -169,22 +169,23 @@ xnvme_buf_fill(void *buf, size_t nbytes, const char *content);
...
  * @param nbytes Amount of bytes to fill with zeroes in buf
  *
- * @return Returns the first argument.
+ * @return On success, 0 is returned. On error, negative `errno` is returned.
  */
-void *
+int
 xnvme_buf_clear(void *buf, size_t nbytes);
 
 /**
```
safl, line 177: The docstring describes that it **returns** which does not match the new convention.

I would go with something like below (which also includes a re-ordering of input-arguments):

```
/**
 * Compare two buffers and count the number of differing bytes.
 *
 * @param expected Pointer to buffer containing the expected data
 * @param actual Pointer to buffer to compare against the expected data
 * @param nbytes Number of bytes to compare
 * @param diff Output pointer storing the number of differing bytes
 *
 * @return 0 on success, negative `errno` on error.
 */
```

safl (commented):
A few minor comments and suggestions below.

Also suggest updating `xnvme_buf_diff()` to follow the convention of input parameters first and output parameters last.

tools/xnvmeperf.c
karlowich: Yes

tests/znd_append.c
karlowich: Yes

include/libxnvme_buf.h
karlowich: done

include/libxnvme_buf.h
karlowich: yes

tools/xnvmeperf.c
karlowich: Yes

safl (dismissed):
Some format issue but otherwise LGTM.

safl (approved):
lgtm

NaddiNadja (approved):
LGTM

PR #645, opened 2026-03-03
More developer tooling

safl:
Replaces the scattered `cijoe-*` Makefile targets with a composable `guest-*` workflow. A sticky guest selection (`make guest-select`) persists across targets, so `guest-up`, `guest-provision`, `guest-test`, and `guest-stop` all operate on the same QEMU guest without repeating configuration. `make verify-guest` chains these into a single end-to-end run and fails fast when the source tarball is missing.

The meson dist-archive fixer is extended to purge SPDK build artifacts that embed host-specific paths, preventing stale configuration from breaking builds on the target system.

## Test plan

- [x] `make gen-artifacts ALLOW_DIRTY=1 && make verify-guest` completes successfully
- [x] CI workflow passes with updated target names

### DIFF
--- a/cijoe/scripts/xnvme_bindings_py_build.py
+++ b/cijoe/scripts/xnvme_bindings_py_build.py
@@ -8,6 +8,7 @@ Retargetable: True
 """
 import errno
 from argparse import ArgumentParser
+from pathlib import Path
 
 
 def add_args(parser: ArgumentParser):
@@ -42,7 +43,7 @@ def main(args, cijoe):
         "make test",
     ]
     for cmd in commands:
-        err, _ = cijoe.run(cmd, cwd=xnvme_source / "python" / "bindings")
+        err, _ = cijoe.run(cmd, cwd=str(Path(xnvme_source) / "python" / "bindings"))
         if err:
             return err
 
--- a/cijoe/scripts/xnvme_bindings_py_test.py
+++ b/cijoe/scripts/xnvme_bindings_py_test.py
@@ -8,6 +8,7 @@ Retargetable: True
 """
 import errno
 from argparse import ArgumentParser
+from pathlib import Path
 
 
 def add_args(parser: ArgumentParser):
@@ -26,7 +27,9 @@ def main(args, cijoe):
     if not xnvme_source:
         return errno.EINVAL
 
-    err, _ = cijoe.run("pytest test_buf.py", cwd=xnvme_source / "python" / "tests")
+    err, _ = cijoe.run(
+        "pytest test_buf.py", cwd=str(Path(xnvme_source) / "python" / "tests")
+    )
     if err:
         return err
 
--- a/toolbox/dev_setup_check.py
+++ b/toolbox/dev_setup_check.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+"""
+Check QEMU development environment readiness for xNVMe guest workflows.
+
+Checks:
+- qemu-system-x86_64 binary available
+- /dev/kvm accessible
+"""
+import os
+import shutil
+import subprocess
+import sys
+
+
+def is_custom_qemu(qemu_bin):
+    """Check if the QEMU binary is the custom xNVMe fork (supports KVS, mdts=0)."""
+
+    try:
+        result = subprocess.run(
+            [qemu_bin, "-device", "nvme-ns,help"],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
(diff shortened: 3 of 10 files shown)


karlowich (dismissed):
LGTM

toolbox/guest_select.py
```diff
@@ -0,0 +1,97 @@
...
+            continue
+
+        config_path = os.path.join(configs_dir, f"{guest}.toml")
+        if os.path.isfile(config_path):
+            guests.append(guest)
+
+    return guests
+
+
+def main():
```
NaddiNadja, line 39: This script's docstring says this is for cijoe workflows, but then the `main()` function need to have `args`, and `cijoe` as parameters. So `main(args, cijoe)`. 
You might want to do `main(args=None, cijoe=None)` just to ensure that you can call `main()` in the bottom of the file, OR call `main(None, None)` in the bottom.

toolbox/dev_setup_check.py
```diff
@@ -0,0 +1,82 @@
...
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
+        return "kv=" in (result.stdout + result.stderr)
+    except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
+        return False
+
+
+def main():
```
NaddiNadja, line 30: If this script should be compatible with cijoe workflows (just like guest_select.py), then same comment holds here :)

cijoe/scripts/xnvme_bindings_py_build.py
```diff
@@ -42,7 +43,7 @@ def main(args, cijoe):
         "make test",
     ]
     for cmd in commands:
-        err, _ = cijoe.run(cmd, cwd=xnvme_source / "python" / "bindings")
+        err, _ = cijoe.run(cmd, cwd=str(Path(xnvme_source) / "python" / "bindings"))
```
NaddiNadja, line 46: You don't need to cast to a string here. In cijoe, we just pass this argument either to `subprocess.Popen()` (with transport local), which can handle a Path object, or into an f-string (with transport ssh), which is also fine.

cijoe/scripts/xnvme_bindings_py_test.py
```diff
@@ -26,7 +27,9 @@ def main(args, cijoe):
     if not xnvme_source:
         return errno.EINVAL
 
-    err, _ = cijoe.run("pytest test_buf.py", cwd=xnvme_source / "python" / "tests")
+    err, _ = cijoe.run(
+        "pytest test_buf.py", cwd=str(Path(xnvme_source) / "python" / "tests")
+    )
```
NaddiNadja, line 32: You don't need to cast to a string here. In cijoe, we just pass this argument either to `subprocess.Popen()` (with transport local), which can handle a Path object, or into an f-string (with transport ssh), which is also fine.

NaddiNadja (changes requested):
A couple things here

toolbox/guest_select.py
safl: It is for CIJOE workflows in the sense that it creates a ``cijoes/guest.current`` such that the Makefile can us to to choose which config to use. It is not for cijoe workflows as in a cijoe-script following the CIJOE script conventions.

toolbox/dev_setup_check.py
safl: No need, this is just a helper-script for the ``Makefile`` not a CIJOE-script.

cijoe/scripts/xnvme_bindings_py_test.py
safl: When running locally I kept getting a bunch of errors thus I changed it to this. In case you don't find this harmful, how about we just keep it for now?

cijoe/scripts/xnvme_bindings_py_build.py
safl: When running locally I kept getting a bunch of errors thus I changed it to this. In case you don't find this harmful, how about we just keep it for now?

toolbox/guest_select.py
NaddiNadja: ohhhh

cijoe/scripts/xnvme_bindings_py_build.py
NaddiNadja: all good

NaddiNadja (approved):
All good then! ⭐

safl:
> LGTM

For proper process (+2) then use the review-function :)

karlowich (approved):

PR #651, opened 2026-03-12
CPU mask fix and printing args in xnvmeperf

NaddiNadja:


karlowich (dismissed):
LGTM

One thought: Don't we have a general purpose hex-parser in the `xnvme_cli.c`.
Perhaps it makes more sense to fix it there? Then we won't have the same problem in some other CLI tool in the future.

NaddiNadja:
> One thought: Don't we have a general purpose hex-parser in the `xnvme_cli.c`. Perhaps it makes more sense to fix it there? Then we won't have the same problem in some other CLI tool in the future.

Yes however, the problem still stands: I need to represent 96 CPUs in a mask, but the uint64_t variable in the xnvme_cli parsing only has 64 bytes to do so :(

karlowich:
> Yes however, the problem still stands

I see, well then this is all good

tools/xnvmeperf.c
```diff
@@ -96,32 +96,63 @@ pin_to_cpu(int cpu)
...
+		char c = hex[i];
+		int value = (c >= '0' && c <= '9')   ? c - '0'
+			    : (c >= 'a' && c <= 'f') ? c - 'a' + 10
+						     : c - 'A' + 10;
+		for (int b = 0; b < 4; b++) {
+			count += (value >> b) & 1;
+		}
 	}
 
 	cpus = malloc(sizeof(int) * count);
```
safl, line 155: When are these de-allocated?

safl (commented):
Does this replace any general infra in the cli-parser? Could/should it?

tools/xnvmeperf.c
NaddiNadja: in line 985 - but this is not part of this PR

NaddiNadja:
> Does this replace any general infra in the cli-parser? Could/should it?

I don't know if it's necessary to replace the general hex-parser with this one, as this one is specifically for very very high numbers, which I don't think we see in other cases than for CPU masks, right?

tools/xnvmeperf.c
safl: Ahh my bad.

tools/xnvmeperf.c
```diff
@@ -891,6 +890,8 @@ sub_run(struct xnvme_cli *cli)
 	struct xnvmeperf_args args = {0};
 	int err = 0;
 
+	printf("Running xnvmeperf with arguments:\n");
```
safl: How about moving this to after input-validation? Such that in case of error it is not just outputting an empty: ``Running xnvmeperf with arguments`` and then just "crickets".

cijoe/tests/tools/test_xnvmeperf.py
```diff
@@ -5,7 +5,7 @@
 def test_run_read(cijoe, device, be_opts, cli_args):
     err, _ = cijoe.run(
         f"xnvmeperf run --iopattern read --qdepth 32 --iosize 4096 --runtime 3"
-        f" --cpumask 0x1 --be {be_opts['be']} {device['uri']}"
+        f" --cpumask 1 --be {be_opts['be']} {device['uri']}"
```
safl: The ``cpumask`` is documented as a hex-value, seems like passing e.g. ``0xFF`` or ``0x1`` is no longer supported?

safl (commented):
I read it as it no longer supports passing ``--cpumask 0x1`` or ``--cpumask 0xFF``, is that correct?

tools/xnvmeperf.c
NaddiNadja: fixed

cijoe/tests/tools/test_xnvmeperf.py
NaddiNadja: Correct. I took a look at the implementation for taskset, and they jsut skip the first two characters, if they are "0x", so I implemented the same. I think the tool would also benefit from allowing the user to give a list of CPUs instead (comma separated, potentially supporting ranges, so e.g. `0-2,5,9,12-13`), but I might just add that as an issue here on github, so that it doesn't inflate this PR.

tools/xnvmeperf.c
```diff
@@ -96,32 +96,69 @@ pin_to_cpu(int cpu)
...
-	int count = 0, err;
+	int len, count = 0, err;
 	int *cpus;
+	bool nonzero = false;
 
-	mask = strtoul(hex, &endptr, 16);
-	if (*endptr != '\0' || endptr == hex || mask == 0) {
+	len = strlen(hex);
+
+	if (len > 1 && !memcmp(hex, "0x", 2L)) {
```
safl:
```suggestion
	if (len > 1 && (!memcmp(hex, "0x", 2L) || !memcmp(hex, "0X", 2L))) {
```

safl (changes requested):
One last thing ``strtoul`` handles ``0x`` as well as ``0X`` this ought to as well.

tools/xnvmeperf.c
NaddiNadja: ah yes of course! added it :)

safl (approved):
There are some things I would do different, but the important thing is getting that cpumask fix in. So, let's go with it as is.

karlowich (approved):
LGTM

PR #658, opened 2026-04-08
docs(tooling): bump dependency versions

safl:
Prompted by https://github.com/xnvme/xnvme/pull/657 i took a look, realizing that xNVMe no longer uses the thing dependabot found an issue with. However, since I was there anyway, here is a soft bump of the documentation dependencies.


NaddiNadja (dismissed):
nice!

karlowich (dismissed):
LGTM

PR #659, opened 2026-04-21
feat(be/upcie-cuda): implement mem_map and mem_unmap

fmma:
Wire the upcie-cuda backend's .mem_map and .mem_unmap hooks, previously stubbed with nosys, so callers can register externally-allocated CUDA buffers (cuMemAlloc, cuMemCreate/cuMemMap, cudaMalloc) for NVMe DMA in addition to the pre-built heap used by xnvme_buf_alloc.

The new mem_map path calls cuMemGetHandleForAddressRange to obtain a dma-buf fd for the registered range, attaches it, and builds a physical address LUT at device_pagesize granularity stored in a new cudamem_mapping entry on a per-runtime linked list. mem_unmap detaches the dma-buf and removes the entry.

The PRP builders in nvme_request_cuda.h no longer assume the payload is physically contiguous: every page is resolved through the new cudamem_vtp helper, which checks the heap first and falls back to the mapping list. This is required for registered buffers, whose physical pages can be non-contiguous.

vaddr and nbytes passed to mem_map must be aligned to device_pagesize; unaligned inputs return -EINVAL. mem_unmap matches by the exact vaddr that was passed to mem_map.

### DIFF
--- a/lib/upcie/xnvme_be_upcie_cuda.h
+++ b/lib/upcie/xnvme_be_upcie_cuda.h
@@ -18,6 +18,7 @@ struct xnvme_be_upcie_cuda_rte {
 	struct cudamem_config cuda_config;
 	struct cudamem_heap cuda_heap;
 	int is_initialized;
+	struct cudamem_mapping *mappings;
 };
 
 extern struct xnvme_be_upcie_cuda_rte g_upcie_cuda_rte;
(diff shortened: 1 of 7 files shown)


fmma:
Closed because PR was created by accident. Not ready yet.

PR #656, opened 2026-03-26
feat(cijoe): add upcie-cuda GPU test workflow and fix CUDA memory compatibility

karlowich:
Adds cijoe test infrastructure for the uPCIe CUDA P2P backend and fixes two
  bugs where CPU code inadvertently accessed GPU device memory.

  The new `test-gpu.yaml` workflow and `upcie-cuda` backend combination entry
  run the test suite against GPU-equipped machines using the `cuda` device label.

  Two segfaults are fixed: `test_ioworker` was allocating the `struct iovec[]`
  array via `xnvme_buf_alloc()` (which returns GPU memory under upcie-cuda), and
  `enumerate_controller()` was walking the Identify NS List response on the CPU
  without first copying it out of the DMA buffer. Both are fixed by using host
  memory for CPU-side access.

  Tests structurally incompatible with CUDA device memory are skipped: admin
  command tools (`xnvme idfy`, `xnvme log-*`, `xnvme dsm`, `lblk idfy`,
  `lblk compare`), fio (ioengine does not support CUDA buffers), and
  `test_init_term` (queue depth sweep exceeds hugepage heap capacity).

PR #660, opened 2026-04-21
feat(upcie/cuda): add GPU-resident NVMe IO queue API (libxnvme_gpu)

karlowich:
Adds a GPU-resident NVMe IO queue API (`libxnvme_gpu`) that enables
  dispatching NVMe commands directly from CUDA kernels without CPU
  involvement, built on top of the uPCIe CUDA backend.

  - **thirdparty**: import `nvme_qpair_cuda`, `nvme_controller_cuda`, and
    `nvme_request_cuda` headers from the uPCIe `gpu-queue` branch
  - **libxnvme_gpu**: add `xnvme_gpu_queue_create/destroy()` for managing
    GPU-resident NVMe IO queues and `xnvme_gpu_cmd_io()` inline helper for
    kernel-side dispatch via `nvme_qpair_cuda_io()`
  - **format**: extend clang-format pre-commit hook to cover `.cu` files
  - **examples**: add `xnvme_gpu.cu` demonstrating host-side queue setup,
    PRP address resolution via `xnvme_buf_vtophys()`, and kernel-side IO
    dispatch
  - **docs**: add `docs/api/c/gpu/` API reference with host-setup and
    device-dispatch sections backed by code snippets; wire into Doxygen
    with `PREDEFINED` CUDA guards


include/libxnvme_gpu.h
```diff
@@ -0,0 +1,94 @@
...
+extern "C" {
+#endif
+
+/**
+ * GPU-resident NVMe queue pair
+ *
+ * Instances live in CUDA device memory. Create one via xnvme_gpu_queue_create()
+ * and pass the returned pointer as a kernel argument.
+ */
+typedef struct nvme_qpair_cuda xnvme_gpu_queue_t;
```
safl: xNVMe only uses typedefs for function-pointers.

safl (changes requested):
The functionality here is strong and already quite useful. However, the current encapsulation and representation of accelerator interactions and their exposed APIs are not where they need to be long term.

Getting this right will require a more substantial redesign. In particular, we need clearer abstractions for different device classes such as storage vs compute, as well as a proper separation of memory representation. That work is non-trivial and will require broader discussion.

To avoid blocking progress, and to make the current functionality available for experimentation, I suggest we proceed with the following adjustments:

* Rename `xnvme_gpu` to `xnvme_cuda`, including headers. The current implementation is CUDA-specific, and generalizing the interface prematurely will likely lead to the wrong abstractions.
* Remove the `typedefs` to simplify and make the code more explicit.
* Add clear documentation caveats wherever the CUDA backend and related APIs are described, explicitly stating that the API is experimental and expected to change before the next xNVMe release.
* Clearly document that this functionality depends on an out-of-tree Linux kernel patch.

With these changes, users can start experimenting with the functionality while being fully aware of the API instability and kernel requirements, and we retain the flexibility to redesign the abstractions with greater care in a later iteration.

include/libxnvme_gpu.h
karlowich: Yes, I know. I chose this path to not "leak" uPCIe symbols to an xNVMe user. I don't think it is nice that a user of xNVMe needs to know about uPCIe specifics, as such I'm "hiding" it behind this typedef.
Of course it can be removed, but at that point it brings into the question if we even need `xnvme_gpu_cmd_io()` as it has the same purpose of "hiding" uPCIe symbols.

examples/xnvme_gpu.cu
```diff
@@ -0,0 +1,143 @@
...
+	/*
+	 * qdepth is both the CUDA block dimension and the NVMe queue depth.
+	 * Pass it as batch_size to the kernel to make all threads active.
+	 * To submit a partial batch, pass a smaller value as batch_size;
+	 * threads with tid >= batch_size participate in barriers but do not
+	 * submit or reap a command.
+	 */
+	size_t qdepth = 64;
+	size_t lba_nbytes = xnvme_dev_get_geo(dev)->lba_nbytes;
+	uint32_t nsid = xnvme_dev_get_nsid(dev);
```
NaddiNadja: these are defined in the middle of the function, is that ok?

karlowich:
> Rename xnvme_gpu to xnvme_cuda, including headers. The current implementation is CUDA-specific, and generalizing the interface prematurely will likely lead to the wrong abstractions.

Yes, that makes sense. I find xnvme_cuda to be too broad, since this is only a queue interface.
Looking at the code, these abstractions could easily fit when adding a ROCm queue. Maybe I can generalize the code a bit more and then it will live up to the `xnvme_gpu` name?

> Add clear documentation caveats wherever the CUDA backend and related APIs are described, explicitly stating that the API is experimental and expected to change before the next xNVMe release.

I can definitely do this. This only goes for the API currently named `xnvme_gpu` right? Every other part of CUDA in xNVMe is hidden behind established abstractions.

> Clearly document that this functionality depends on an out-of-tree Linux kernel patch.

This is already covered in docs: https://xnvme.io/en/next/background/backends/upcie/cuda.html#kernel-requirement

examples/xnvme_gpu.cu
```diff
@@ -0,0 +1,143 @@
...
+
+		cmds[i].common.opcode = XNVME_SPEC_NVM_OPC_READ;
+		cmds[i].nvm.slba = i;
+		cmds[i].common.nsid = nsid;
+		cmds[i].common.dptr.prp.prp1 = phys;
+		cmds[i].nvm.nlb = 0; /* nlb is zero-based; 0 means 1 LBA */
+	}
+
+	/* Copy commands to device memory before passing to the kernel */
+	if (cudaMalloc((void **)&cmds_dev, qdepth * sizeof(*cmds_dev)) != cudaSuccess) {
```
NaddiNadja: can we get cudaMalloc out of the if statement? also in the next if statement

naddinadja (commented):
These two comments are about the style, and I know it's in a `.cu` file, but since it looks so much like C, I think we could adhere to our style guides as much as possible :)

examples/xnvme_gpu.cu
karlowich: Should be cleaned up of course

examples/xnvme_gpu.cu
karlowich: Should be cleaned up of course

include/libxnvme_gpu.h
safl: As you know, then the leaking and pollution happens here:
```
#ifdef __CUDACC__
#include <errno.h>
#include <upcie/nvme/nvme_command.h>
#include <upcie/nvme/nvme_qpair_cuda.h>
#else
struct nvme_qpair_cuda;
#endif
```

The ``typedef`` does not improve upon that. It just confuses the subject, as typedefs do.

safl:
> > Rename xnvme_gpu to xnvme_cuda, including headers. The current implementation is CUDA-specific, and generalizing the interface prematurely will likely lead to the wrong abstractions.
> 
> Yes, that makes sense. I find xnvme_cuda to be too broad, since this is only a queue interface. Looking at the code, these abstractions could easily fit when adding a ROCm queue. Maybe I can generalize the code a bit more and then it will live up to the `xnvme_gpu` name?

Well, my point is that we should not spend time on trying to abstract and encapsulate now, we leave that for later. ROCm might easily fit here as well, that is great, but what if the abstraction should be ``xnvme_axx`` or embedded in ``xnvme_dev``? These are the types of discussions that he API needs when refactoring it, but now is not the time for it, so let's not try and half-bake it, just call it what it currently is, and then we make it **very** clear, that these APIs are experimental. That is, directly write in the top-level docsting ``EXPERIMENTAL``.

> 
> > Add clear documentation caveats wherever the CUDA backend and related APIs are described, explicitly stating that the API is experimental and expected to change before the next xNVMe release.
> 
> I can definitely do this. This only goes for the API currently named `xnvme_gpu` right? Every other part of CUDA in xNVMe is hidden behind established abstractions.

Yes.

> 
> > Clearly document that this functionality depends on an out-of-tree Linux kernel patch.
> 
> This is already covered in docs: https://xnvme.io/en/next/background/backends/upcie/cuda.html#kernel-requirement

This is easy to overlook, emphasize it clearly at the top paragraph with ```{warning}....``` as a user it is much nicer to know ahead of time.

karlowich:
@safl I have addressed everything. Please see attached screenshots for docs warnings.
Screenshot 2026-04-21 at 15 24 25
Screenshot 2026-04-21 at 15 25 36

safl (dismissed):
Great!

safl:
@karlowich I rebased this after #656 landed in ``next`` so CI can run now, saving you the wait later. It should be ready to merge when you’re back, pending ACK from @NaddiNadja.

safl (approved):
great!

docs/api/index.rst
```diff
@@ -53,6 +53,11 @@ practical :ref:`sec-api-c-examples`.
   and convenience functions for general applicability, such as ``XNVME_DEBUG``,
   wall-clock timers, library introspection for version and capabilities, etc.
 
+:ref:`sec-api-c-gpu`
+  Host-side queue management and device-side helpers for submitting NVMe
+  commands directly from CUDA kernels, available with the ``upcie-cuda``
```
naddinadja, line 58: this shouldn't be blocking, but do we write \`\`upcie-cuda\`\` or \*\*upcie-cuda\*\*?

naddinadja (approved):
I'll approve this, but I do have a comment. I don't wanna block this further because of such a small thing :)

docs/api/index.rst
karlowich: I see it's a little inconsistent. But looks like it is only \*\*upcie-cuda\*\* in the upcie-cuda docs page. Every where else we use \`\`upcie-cuda\`\`

PR #662, opened 2026-04-22
feat(tools/xnvmeperf): add GPU benchmark subcommands (gpu-run, gpu-verify)

karlowich:
Adds GPU-direct NVMe benchmarking to `xnvmeperf` via two new CUDA-only subcommands, along with refactoring work to support the new GPU path cleanly.

  ### New features
  - **`gpu-run`**: time-bounded GPU NVMe benchmark using CUDA kernels (sequential and random), issuing commands via `xnvme_gpu_cmd_io()`. Command templates are built host-side with `xnvme_buf_vtophys()` for PRP entries (1-page, 2-page, PRP-list), then copied to device memory before kernel launch. Reports IOPS and MiB/s.
  - **`gpu-verify`**: writes an LBA-stamped pattern to each device through GPU queues and reads it back to verify data integrity. Uses the same queue topology as `gpu-run`, providing  confidence that benchmark results are correct.
  - **`--queues`**: new CLI option to control the number of NVMe queues per device. Total job slots = devices × queues, distributed evenly across threads.

  ### Refactoring (preparatory)
  - Moved `xnvmeperf.c` into `tools/xnvmeperf/` subdirectory with its own `meson.build`
  - Extracted `xnvmeperf.h` with shared types (`enum iopattern`, `struct xnvmeperf_args`) used by both CPU and GPU paths
  - Extracted `parse_common_args()` / `parse_run_args()` helpers to remove duplicated argument parsing across subcommands
  - Extracted `xnvmeperf_open_devs()` helper to consolidate the repeated open-all-devices loop
  - Extracted `print_run_args()` and `print_perf_results()` helpers for consistent output formatting
  - `meson.build` adds the CUDA language and links `xnvmeperf_gpu.cu` when `XNVME_BE_UPCIE_CUDA_ENABLED`

  ### Docs
  - Added `gpu-run` and `gpu-verify` usage docs with `.cmd`/`.out` files
  - Cross-referenced the `upcie-cuda` backend and `libxnvme_cuda` API sections in `docs/tools/xnvmeperf/index.rst`

### DIFF
--- a/docs/tools/xnvmeperf/index.rst
+++ b/docs/tools/xnvmeperf/index.rst
@@ -5,8 +5,13 @@ xnvmeperf
 
 **xnvmeperf** is a multi-threaded async I/O benchmark for NVMe devices. It
 runs a time-bounded async I/O loop and reports per-device and aggregate IOPS
-and throughput. A ``verify`` subcommand is also provided to check data integrity
-by writing a known per-LBA pattern and reading it back.
+and throughput. A ``verify`` subcommand checks data integrity by writing a known
+per-LBA pattern and reading it back.
+
+When built with CUDA support (``XNVME_BE_UPCIE_CUDA_ENABLED``), two additional
+subcommands are available: ``cuda-run`` and ``cuda-verify``. These run NVMe I/O
+directly from CUDA kernels via the :ref:`sec-backends-upcie-cuda` backend, bypassing
+the host I/O path entirely.
 
 .. literalinclude:: xnvmeperf_usage.out
    :language: bash
@@ -48,3 +53,51 @@ mismatches and I/O errors per device.
 Example::
 
    xnvmeperf verify --iosize 4096 --count 256 /dev/nvme0n1
+
+``cuda-run`` — GPU benchmark
+============================
+
+.. seealso::
+
+   :ref:`sec-backends-upcie-cuda`
+      Backend setup, system configuration, and memory architecture.
+
+   :ref:`sec-api-c-gpu`
+      The ``libxnvme_cuda`` API used to create queues and dispatch commands
+      from CUDA kernels.
+
+Requires the ``upcie-cuda`` backend. All queues across all devices are driven
+by a single CUDA kernel: each CUDA block owns one NVMe queue and each thread
+within the block owns one queue slot, so ``--qdepth`` threads submit and reap
+commands in lock-step. The grid has ``ndevs × --queues`` blocks in total.
+
+Both ``--qdepth`` and ``--iosize`` must be powers of 2. Supported patterns are
+``read``, ``write``, ``randread``, and ``randwrite``.
+
+.. literalinclude:: xnvmeperf_cuda_run_usage.out
+   :language: bash
+
+Example — sequential read, four queues of depth 32 on two devices::
+
+   xnvmeperf cuda-run --iopattern read --queues 4 --qdepth 32 --iosize 4096 \
+       --runtime 10 --be upcie-cuda 0000:01:00.0 0000:02:00.0
+
+Example — random write, single queue::
+
+   xnvmeperf cuda-run --iopattern randwrite --qdepth 64 --iosize 4096 \
+       --runtime 10 --be upcie-cuda 0000:01:00.0
+
+``cuda-verify`` — GPU data integrity check
+==========================================
+
+Uses the same queue topology as ``cuda-run``. For each queue slot the kernel
+writes a unique LBA-stamped pattern, then reads it back on the host and
+compares against the expected data. This is intended to confirm that
+``cuda-run`` results reflect correct I/O rather than silent data corruption.
+
+.. literalinclude:: xnvmeperf_cuda_verify_usage.out
+   :language: bash
+
+Example::
+
+   xnvmeperf cuda-verify --iosize 4096 --qdepth 32 --be upcie-cuda 0000:01:00.0
--- a/lib/xnvme_cli.c
+++ b/lib/xnvme_cli.c
@@ -883,6 +883,12 @@ static struct xnvme_cli_opt_attr xnvme_cli_opts[] = {
 		.descr = "Hex CPU bitmask for thread pinning (e.g. 0x3)",
 	},
 	{
+		.opt = XNVME_CLI_OPT_NQUEUES,
+		.vtype = XNVME_CLI_OPT_VTYPE_NUM,
+		.name = "nqueues",
+		.descr = "Number of queues per device",
+	},
+	{
 		.opt = XNVME_CLI_OPT_END,
 		.vtype = XNVME_CLI_OPT_VTYPE_NUM,
 		.name = "",
@@ -1563,6 +1569,9 @@ xnvme_cli_assign_arg(struct xnvme_cli *cli, struct xnvme_cli_opt_attr *opt_attr,
 		break;
 	case XNVME_CLI_OPT_CPUMASK:
 		args->cpumask = arg ? arg : "INVALID_INPUT";
+		break;
+	case XNVME_CLI_OPT_NQUEUES:
+		args->nqueues = num;
 		break;
 	case XNVME_CLI_OPT_POSA_TITLE:
 	case XNVME_CLI_OPT_NON_POSA_TITLE:
--- a/tools/xnvmeperf.c
+++ b/tools/xnvmeperf.c
@@ -473,10 +473,98 @@ dev_close_fn(void *arg)
 }
 
 static void
+print_run_args(struct xnvmeperf_args *args, const char *pattern)
+{
+	printf("Running xnvmeperf with arguments:\n");
+	printf("- Devices: [");
+	for (int i = 0; i < args->ndevs; i++) {
+		if (i) {
+			printf(", ");
+		}
+		printf("%s", args->dev_uris[i]);
+	}
+	printf("] (total: %d)\n", args->ndevs);
+	printf("- io pattern: %s\n", pattern);
+	printf("- queues per device: %u\n", args->nqueues);
+	printf("- queue depth: %u\n", args->qdepth);
+	printf("- io size: %u\n", args->iosize);
+	printf("- runtime: %u\n", args->time);
+
+	if (args->ncpus) {
+		printf("- CPUs: [");
+		for (int i = 0; i < args->ncpus; i++) {
+			if (i) {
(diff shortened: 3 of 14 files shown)


lib/xnvme_cli.c
```diff
@@ -882,6 +882,12 @@ static struct xnvme_cli_opt_attr xnvme_cli_opts[] = {
 		.name = "cpumask",
 		.descr = "Hex CPU bitmask for thread pinning (e.g. 0x3)",
 	},
+	{
+		.opt = XNVME_CLI_OPT_QUEUES,
+		.vtype = XNVME_CLI_OPT_VTYPE_NUM,
+		.name = "queues",
```
safl: In other places then we refer to the number of queues as ``nqueues``.

tools/xnvmeperf/xnvmeperf.h
```diff
@@ -35,4 +35,10 @@ struct xnvmeperf_args {
 int
 fill_pattern(void *buf, size_t nbytes, uint64_t slba, uint16_t nlb);
 
+#ifdef XNVME_BE_UPCIE_CUDA_ENABLED
+int
+xnvmeperf_gpu_run(struct xnvme_dev **devs, const struct xnvmeperf_args *args,
```
safl: How about ``xnvmeperf_cuda_run``?

tools/xnvmeperf/xnvmeperf_cuda.cu
```diff
@@ -0,0 +1,653 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
safl, line 3: Since the file-type ``.cu`` already tells one that this is CUDA, then how about just naming it ``xnvmeperf.cu`` and also drop the "_gpu" 
on the helpers.

Or, name it ``xnvmeperf_cuda.cu`` and rename the helpers to ``xnvmeperf_cuda_*()``.

docs/tools/xnvmeperf/index.rst
```diff
@@ -5,8 +5,13 @@ xnvmeperf
...
 runs a time-bounded async I/O loop and reports per-device and aggregate IOPS
-and throughput. A ``verify`` subcommand is also provided to check data integrity
-by writing a known per-LBA pattern and reading it back.
+and throughput. A ``verify`` subcommand checks data integrity by writing a known
+per-LBA pattern and reading it back.
+
+When built with CUDA support (``XNVME_BE_UPCIE_CUDA_ENABLED``), two additional
+subcommands are available: ``gpu-run`` and ``gpu-verify``. These run NVMe I/O
+directly from CUDA kernels via the ``upcie-cuda`` backend, bypassing the host
+I/O path entirely.
```
safl: Add a reference to the upcie-cuda docs here, then one gets the notice and warnings etc.

safl (changes requested):
Awesome. One nitpick and then request to go with the non-encapsulated naming that we went with for https://github.com/xnvme/xnvme/pull/660 provided some examples inline.

lib/xnvme_cli.c
karlowich: Right of course

tools/xnvmeperf/xnvmeperf.h
karlowich: Sure

tools/xnvmeperf/xnvmeperf_cuda.cu
karlowich: Let's go with `xnvmeperf_cuda.cu` (I think it's nice to have the postfix so you don't mistake the files visually) with function names `xnvmperf_cuda_*()`

docs/tools/xnvmeperf/index.rst
karlowich: Sure

safl (dismissed):
lgtm.

tools/xnvmeperf.c
```diff
@@ -1027,32 +1060,19 @@ sub_verify(struct xnvme_cli *cli)
...
+	err = parse_common_args(cli, &args);
+	if (err) {
 		return err;
 	}
 
 	if (!cli->args.count) {
-		err = -EINVAL;
-		xnvme_cli_perr("Error: --count must be a positive integer", err);
-		return err;
+		fprintf(stderr, "Error: --count must be a positive integer\n");
```
naddinadja: use `xnvme_cli_perr` here

tools/xnvmeperf.c
```diff
@@ -25,6 +25,7 @@ struct xnvmeperf_args {
 	uint32_t iosize;
 	uint32_t time;
 	uint32_t count;
+	uint32_t nqueues;
```
naddinadja: How about saving this as an int instead of having to cast it to int every time we use it?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -1145,6 +1145,76 @@ sub_verify(struct xnvme_cli *cli)
...
+	failed_per_dev = calloc(args.ndevs, sizeof(*failed_per_dev));
+
+	if (!rounds_per_dev || !failed_per_dev) {
+		free(rounds_per_dev);
+		free(failed_per_dev);
+		err = -ENOMEM;
+		goto close_devs;
+	}
+
+	err = xnvmeperf_cuda_run(devs, &args, rounds_per_dev, failed_per_dev, &elapsed_ms);
```
naddinadja: Could the opening of the devices etc be moved into `xnvmeperf_cuda_run()`, such that the structure is the same across all three subcommands?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -1145,6 +1145,76 @@ sub_verify(struct xnvme_cli *cli)
...
+	if (err) {
+		fprintf(stderr, "Failed: xnvmeperf_cuda_run(): err(%d)\n", err);
```
naddinadja: use `xnvme_cli_perr()`

tools/xnvmeperf/xnvmeperf_cuda.cu
naddinadja, line 1: Here we don't print on most of the error paths. Can we get some `xnvme_cli_perr` to explain what happened?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -1215,6 +1215,55 @@ sub_cuda_run(struct xnvme_cli *cli)
 	return err;
 }
 
+static int
+sub_cuda_verify(struct xnvme_cli *cli)
```
naddinadja, line 1286: Can you move the opening of devices and printing to inside `xnvmeperf_cuda_verify()` such that the structure is the same across all four subcommands?

docs/tools/xnvmeperf/index.rst
```diff
@@ -48,3 +53,51 @@ mismatches and I/O errors per device.
...
+Both ``--qdepth`` and ``--iosize`` must be powers of 2. Supported patterns are
+``read``, ``write``, ``randread``, and ``randwrite``.
+
+.. literalinclude:: xnvmeperf_cuda_run_usage.out
+   :language: bash
+
+Example — sequential read, four queues of depth 32 on two devices::
+
+   xnvmeperf cuda-run --iopattern read --queues 4 --qdepth 32 --iosize 4096 \
+       --runtime 10 --be upcie-cuda 0000:01:00.0 0000:02:00.0
```
naddinadja, line 83: What happens if you run `xnvmeperf cuda-run` without `--be upcie-cuda`? If the backend MUST be upcie-cuda, is there value in defaulting to this if the backend argument is not given?

naddinadja (changes requested):
I have some comments inline

tools/xnvmeperf.c
karlowich: I see the printing is a bit all over the place. Do you want this format everywhere possible?
```
err = -EINVAL;
xnvme_cli_perr("Error: --count must be a positive integer", err);
return err;
```

tools/xnvmeperf/xnvmeperf.c
karlowich: This would mean creating a function `xnvmeperf_cuda_run()` in `xnvmeperf.c` and renaming the one in `xnvmeperf_cuda.cu`. I want as little in the cuda file as possible, to avoid c++-isms as much as possible

tools/xnvmeperf.c
karlowich: I will consider how it looks in the cuda file. But in xnvmeperf.c the casts should just be removed, they are c++-isms.

tools/xnvmeperf/xnvmeperf.c
karlowich: Yes

tools/xnvmeperf/xnvmeperf_cuda.cu
karlowich: Yes, that seems appropriate

docs/tools/xnvmeperf/index.rst
karlowich: You're right, this should be enforced

tools/xnvmeperf/xnvmeperf.c
karlowich: I will take this approach

tools/xnvmeperf/xnvmeperf.c
karlowich: I will take same approach as for `xnvmeperf_cuda_run()`

docs/tools/xnvmeperf/index.rst
karlowich: Oh I don't know about defaulting to it. Probably yes, I mean why not, it is essentially a dead argument for now.

tools/xnvmeperf.c
karlowich: I will implement this pattern. It seems the best

tools/xnvmeperf.c
karlowich: Changing this would create a mess in the CUDA file, which uses nqueues all the time. I would rather keep the 3 casts we have in xnvmeperf.c

karlowich:
@NaddiNadja Addressed all your feedback now

safl (approved):
lgtm

naddinadja (approved):
lgtm :)

PR #663, opened 2026-04-27
feat(be/upcie-cuda): implement mem_map and mem_unmap

fmma:
Depends on https://github.com/safl/upcie/pull/25.

Wire the upcie-cuda backend's `.mem_map` and `.mem_unmap` hooks, previously stubbed with nosys, so callers can register externally-allocated CUDA buffers (`cuMemAlloc`, `cuMemCreate`/`cuMemMap`, `cudaMalloc`) for NVMe DMA in addition to the pre-built heap used by `xnvme_buf_alloc`.

Mappings are tracked on a per-runtime linked list of `cudamem_mapping` entries; `mem_map` appends one via `cudamem_mapping_add()` and `mem_unmap` removes the matching entry via `cudamem_mapping_remove()`. The add path obtains a dma-buf fd through `cuMemGetHandleForAddressRange`, attaches it, and builds a per-device-page LUT. `_cuda_rte_term()` drains any mappings still alive via `cudamem_mapping_clear()` when the last controller is closed.

`buf_vtophys` and the per-command PRP build paths dispatch on buffer provenance via `cudamem_heap_contains()`: heap-backed buffers resolve through `cudamem_heap_block_virt_to_phys`, mapped buffers through the list-walking `cudamem_mapping_virt_to_phys`. Command paths use the unsuffixed `nvme_request_prep_command_prps_{contig,iov}_cuda` dispatcher which routes to the heap or mapped variant. A resolution failure on a mapped buffer is reported as `-EINVAL` up to the cmd dispatchers rather than silently producing a PRP pointing at physical address 0.

`vaddr` and `nbytes` passed to `mem_map` must be aligned to `device_pagesize`; unaligned inputs return `-EINVAL`. `mem_unmap` matches by the exact `vaddr` that was passed to `mem_map`.

An accompanying example (`examples/xnvme_cuda_mem_map.cu`) demonstrates `cudaMalloc` + `xnvme_mem_map` + GPU-resident NVMe IO end to end.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -37,4 +37,14 @@ if conf_data.get('XNVME_BE_UPCIE_CUDA_ENABLED')
     install_rpath: xnvmelib_rpath,
     install: true,
   )
+  executable(
+    'xnvme_cuda_mem_map',
+    'xnvme_cuda_mem_map.cu',
+    include_directories: [conf_inc, xnvme_inc],
+    link_with: xnvmelib,
+    dependencies: [cuda_dep],
+    link_args: link_args_hardening,
+    install_rpath: xnvmelib_rpath,
+    install: true,
+  )
 endif
--- a/examples/xnvme_cuda_mem_map.cu
+++ b/examples/xnvme_cuda_mem_map.cu
@@ -1,0 +1,179 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <errno.h>
+#include <stdint.h>
+#include <libxnvme.h>
+
+/**
+ * GPU-resident NVMe IO into a cudaMalloc'd, mem_map'd buffer
+ *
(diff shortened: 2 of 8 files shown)


safl (approved):
I just did a quick test integrating AiSIO with **llama.cpp**, I used this as the base for inter-operating with the existing memory allocated and managed llama.cpp, i do have some changes on top, doing the mapping at larger granularity, but this it good as is. Thanks!

karlowich:
This looks good, but I would like it to be in a different shape. Let's create a new xNVMe backend `upcie-cuda-map` (or something like that), instead of extending `upcie-cuda`.
This will save us the branching on the IO path.
The new xNVMe backend should only support map/unmap in the memory interface.
I think this is the cleanest version until #707 changes how uPCIe handles memory.

karlowich (changes requested):
Accidentally clicked "Comment" instead of "Request changes"

safl:
Closing this as it was superseded by https://github.com/xnvme/xnvme/pull/745

PR #665, opened 2026-05-04
feat(be): allow upcie and upcie-cuda to share an NVMe controller simultaneously

karlowich:
This series allows the `upcie` backend variants (`upcie` and `upcie-cuda`) to
share the same NVMe controller at the same time, enabling dual-backend use
cases such as opening a single device with both `upcie` and `upcie-cuda`
concurrently.

### Mechanism

Controller state in the `cref` table is keyed by backend name, so a second
backend config opening an already-open URI is normally rejected with `-EBUSY`.
This is relaxed for the `upcie` variants: because they share the same
underlying NVMe controller infrastructure, they accept each other's crefs
despite the inexact name match.

### Changes

- **`feat(be)`** — Relax the cref name-match check so `upcie` and `upcie-cuda`
  reuse the same controller for a given URI instead of conflicting.
- **`feat(upcie)`** — Move the CUDA runtime lifecycle (`_cuda_rte_init` /
  `_cuda_rte_term`) out of `ctrlr_init` / `ctrlr_term` and into `dev_open` /
  `dev_close`, so CUDA is initialised for every `upcie-cuda` handle regardless
  of which backend performed the primary controller open. The driver check
  (`uio_pci_generic` required, `vfio-pci` rejected) is kept in both
  `ctrlr_init` (primary path) and `dev_open` (secondary path, where
  `ctrlr_init` is skipped).
- **`feat(cli)`** — Add `XNVME_CLI_OPT_ALT_BE` / `--alt-be` (a second backend
  name independent of `--be`). This is purely a CLI/test concern, used so tests
  can open the same device with two different backends; there is no
  primary/secondary hierarchy and the order is irrelevant.
- **`test(dual-be)`** — New `dual-be` test tool with two subcommands:
  - `io-cmp`: write a known pattern via `--be`, read it back via `--alt-be`,
    and compare.
  - `idfy-cmp`: issue Identify Controller via both backends and byte-compare
    the responses.
- **`test(cijoe)`** — pytest-based integration tests:
  - `test_open_dev_twice`: opens the same device twice with identical backends
    to verify same-name cref reuse.
  - `test_io_cmp` / `test_idfy_cmp`: exercise the dual-backend path for
    `upcie` / `upcie-cuda` pairs (skipped for backends without an alternate).
- **`docs(upcie)`** — Document dual-backend operation for `upcie` and
  `upcie-cuda`.

### User-facing impact

Minimal and intentionally so. Today the only supported combination is mixing
`upcie` and `upcie-cuda`. Just as a device URI like `0000:01:00.0` can already
be opened multiple times with the same backend, it can now also be opened with
these two different backends at once. No new concepts are exposed to ordinary
users.

**OLD Description for context:**

  This series allows compatible xNVMe backends to share the same NVMe controller

  simultaneously, enabling dual-backend use cases such as opening a device with
  both `upcie` and `upcie-cuda` at the same time.

  - **Backend family bitmask** (`xnvme_be_family`, `uint32_t family` in
    `xnvme_be_attr`): backends in the same family can reuse a live controller
    reference instead of conflicting on the URI. Replaces four reserved bytes —
    ABI size unchanged.
  - **Compat cref path** (`xnvme_be_cref_lookup_compat`): when opening a device
    URI, if a compatible-family controller is already open, its cref is reused
    and only per-handle init (`dev_open`) is called. Fixes a pre-existing bug
    where same-name cref hits incorrectly skipped `dev_open`.
  - **`--alt-be` CLI option** (`XNVME_CLI_OPT_ALT_BE`): allows tools to accept a
    second backend name for simultaneous dual-backend device access.
  - **`upcie` + `upcie-cuda` coexistence**: both backends are registered in
    `XNVME_BE_FAMILY_UPCIE`. CUDA runtime lifecycle (`_cuda_rte_init` /
    `_cuda_rte_term`) is moved from `ctrlr_init` / `ctrlr_term` into
    `dev_open` / `dev_close` so CUDA handles initialise correctly on the compat
    secondary path.
  - **`be-families` test tool** with two subcommands: `io-cmp` (write via `--be`,
    read back via `--alt-be`, compare) and `idfy-cmp` (Identify Controller via
    both backends, byte-compare responses).
- **cijoe integration tests** (`test_xnvme_tests_be_families`): `test_open_dev_twice`
    verifies same-name cref reuse; `test_io_cmp` and `test_idfy_cmp` exercise the
    compat-family path for `upcie` / `upcie-cuda` pairs.

### DIFF
--- a/cijoe/tests/logic/test_xnvme_tests_dual_be.py
+++ b/cijoe/tests/logic/test_xnvme_tests_dual_be.py
@@ -1,0 +1,43 @@
+import pytest
+
+from ..conftest import xnvme_parametrize
+
+
+def get_alt_be(be: str):
+    if be == "upcie":
+        return "upcie-cuda"
+    elif be == "upcie-cuda":
+        return "upcie"
+    else:
+        return None
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be"])
+def test_open_dev_twice(cijoe, device, be_opts, cli_args):
+    # We are repurposing "xnvme_tests_dual_be idfy-cmp" to verify that opening the same device twice with the same backend works
+    err, _ = cijoe.run(
+        f"xnvme_tests_dual_be idfy-cmp {device['uri']} --be {be_opts['be']} --alt-be {be_opts['be']} --dev-nsid {device['nsid']}"
+    )
+    assert not err
+
+
+@xnvme_parametrize(labels=["dev", "cuda"], opts=["be"])
+def test_io_cmp(cijoe, device, be_opts, cli_args):
+    alt_be = get_alt_be(be_opts["be"])
+    if not alt_be:
+        pytest.skip(reason=f"[be={be_opts['be']}] does not have an alternate backend")
+    err, _ = cijoe.run(
+        f"xnvme_tests_dual_be io-cmp {device['uri']} --be {be_opts['be']} --alt-be {alt_be} --dev-nsid {device['nsid']}"
+    )
+    assert not err
+
+
+@xnvme_parametrize(labels=["dev", "cuda"], opts=["be"])
+def test_idfy_cmp(cijoe, device, be_opts, cli_args):
+    alt_be = get_alt_be(be_opts["be"])
+    if not alt_be:
+        pytest.skip(reason=f"[be={be_opts['be']}] does not have an alternate backend")
+    err, _ = cijoe.run(
+        f"xnvme_tests_dual_be idfy-cmp {device['uri']} --be {be_opts['be']} --alt-be {alt_be} --dev-nsid {device['nsid']}"
+    )
+    assert not err
--- a/tests/dual_be.c
+++ b/tests/dual_be.c
@@ -1,0 +1,243 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <errno.h>
+#include <stdlib.h>
+
+#include <libxnvme.h>
+
+static int
+open_alt(struct xnvme_cli *cli, struct xnvme_dev **alt_dev)
+{
+	struct xnvme_opts opts = xnvme_opts_default();
+	const char *uri = cli->args.uri;
+	int err;
+
+	opts.be = cli->args.alt_be;
+	opts.nsid = cli->args.dev_nsid;
+
+	*alt_dev = xnvme_dev_open(uri, &opts);
+	if (!(*alt_dev)) {
+		err = -errno;
+		xnvme_cli_perr("xnvme_dev_open(alt_dev)", err);
+		return err;
+	}
+
+	return 0;
+}
+
+static int
+sub_io_cmp(struct xnvme_cli *cli)
+{
+	struct xnvme_dev *alt_dev, *dev = cli->args.dev;
+	const struct xnvme_geo *geo;
+	struct xnvme_cmd_ctx ctx;
+	void *write_buf = NULL, *read_buf = NULL;
+	uint32_t iosize, nsid;
+	uint16_t nlb;
+	size_t diff = 0;
+	int err;
+
+	nsid = cli->args.dev_nsid;
+
+	err = open_alt(cli, &alt_dev);
+	if (err) {
+		return err;
+	}
+
+	geo = xnvme_dev_get_geo(dev);
+	if (!geo->lba_nbytes) {
+		err = -EINVAL;
+		xnvme_cli_perr("Error: LBA size == 0", err);
+		goto exit;
+	}
+
+	iosize = geo->lba_nbytes;
+	nlb = 0;
+
+	write_buf = xnvme_buf_alloc(dev, iosize);
+	if (!write_buf) {
+		err = -errno;
+		xnvme_cli_perr("xnvme_buf_alloc(write)", err);
+		goto exit;
+	}
+
+	err = xnvme_buf_fill(write_buf, iosize, "anum");
+	if (err) {
+		xnvme_cli_perr("xnvme_buf_fill(write)", err);
+		goto exit;
+	}
+
+	ctx = xnvme_cmd_ctx_from_dev(dev);
+	err = xnvme_nvm_write(&ctx, nsid, 0, nlb, write_buf, NULL);
+	if (err || xnvme_cmd_ctx_cpl_status(&ctx)) {
+		xnvme_cli_perr("xnvme_nvm_write()", err);
+		xnvme_cmd_ctx_pr(&ctx, XNVME_PR_DEF);
+		err = err ? err : -EIO;
+		goto exit;
+	}
+
+	read_buf = xnvme_buf_alloc(alt_dev, iosize);
+	if (!read_buf) {
+		err = -errno;
+		xnvme_cli_perr("xnvme_buf_alloc(read)", err);
+		goto exit;
+	}
+
+	err = xnvme_buf_fill(read_buf, iosize, "zero");
+	if (err) {
+		xnvme_cli_perr("xnvme_buf_fill(read)", err);
+		goto exit;
+	}
+
+	ctx = xnvme_cmd_ctx_from_dev(alt_dev);
+	err = xnvme_nvm_read(&ctx, nsid, 0, nlb, read_buf, NULL);
+	if (err || xnvme_cmd_ctx_cpl_status(&ctx)) {
+		xnvme_cli_perr("xnvme_nvm_read()", err);
+		xnvme_cmd_ctx_pr(&ctx, XNVME_PR_DEF);
+		err = err ? err : -EIO;
+		goto exit;
+	}
+
+	err = xnvme_buf_diff(write_buf, read_buf, iosize, &diff);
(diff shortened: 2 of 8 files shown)


safl (commented):
Nice catch on the cref reuse failure path.

For sequencing, I would prefer that fix is not mixed with the refactor/concept
introduction. Keeping the bug fix separate avoids conflating a correctness fix
with a feature/design change, and prevents either one from blocking the other
from progressing.

For the refactor, I think `xnvme_be_family` is too broad. The backend is not
what is shared; the shared resource is the controller-state object stored in
`dev->be.state[0]`.

Could we model this as something like `ctrlr_state_id` instead?

- `0` means exclusive / not shareable.
- Same non-zero `ctrlr_state_id` means configs can share an existing controller
  handle for the URI.
- `upcie` and `upcie-cuda` would both use something like
  `XNVME_CTRLR_STATE_UPCIE`.

I am leaning away from expanding the `cref` helper API here. The recent
backend/controller-ref refactoring and platform introduction were about
simplifying things, and growing the controller reference-counting surface risks
adding back complexity. I would prefer we instead simplify further: make `cref`
lookup resolve only by URI and leave compatibility decisions to the platform.

Helpers are fine, but they should be platform helpers rather than `cref`
helpers. That keeps `cref` simple: track URI -> controller/refcount and return
the existing controller plus its controller-state id, or enough metadata for the
platform to determine it.

With `ctrlr_state_id`, the lookup/ref part and the compatibility decision become
neatly separated:

- `cref` lookup is only by URI.
- `xnvme_platform.c` compares the existing controller-state id with the requested
  backend config's `ctrlr_state_id` and decides whether reuse is allowed.

That keeps the compatibility rule close to backend selection, and avoids
teaching `cref` about backend families or compatible backend matching.

karlowich:
This `cref` code is a can of worms. I will open a new PR with refactoring/simplifaction before working more on this PR.

safl (commented):
I cannot follow the concepts here, i see the following terms introduced without any definition of what they are:

* Backend family for shared NVMe controllers
  * What constitutes a backend family? What is shared, who shares, and how? Where is the backend family defined, what can it do or serve, what is it?

* Allow compatible backends to share a device URI
  * What does it mean to share a "device URI", is the device URI truly the thing being shared?
  * The commit message says something about a bitmask, and it is used somehow.. why is it a bitmask and not just a value? Seems like there is only one value defined in this PR and it looks more like an identifier than something that needs bitmask.
  * When introducing a compatibility flag why is the name still used as a means to detect compatibility, why not use the infrastructure introduced here with the compatibility bitmask, if that is what it is? E.g. why have more ways of determining compatibility when you a introducing infrastructure specifically to determine compatibility?

* feat(upcie): allow upcie and upcie-cuda to coexist on the same device
  * Looking at this, then the "compat-bitmask" looks even more like an identifier than anything else, like it declares "i use this value to identity the state i need". Also, this seems to be bound on the implementation of the ctrlr-state, thus, it would seem more natural that a controller-state was defined somewhere and it had the identifier value, then some consumer like the backend here says: "this is the controller-state i use", and then when doing looksup in cref, then it is looking up on the device-uri and then checking compatibility by checking the controller-format/state identifer.

* alt_be and dual-backend
  * When you call something "alt-be" my mind goes to "alternative backend", is that what this is? It also looks like it is solely a cli thing? No mentions in the backend? Is it solely a cli/test concern?
  * In case that this is an "alternative" backend, what does that mean, when is it in effect? What is the relation primary/secondary or clients/server, what is going on her?
  * IN the docs something is described as dual-backend but it does not really shed a light on this.

In #686 also does sharing... how do these PRs relate? Are they, or could they, fundamentally solve the same problem?

All of this terminology and concepts are bleeding through to the user, is that really needed? Could it be encapsulated? How does one know what to combine as one or the other? 

P.S. The PR description seems to be very outdated, reading it makes is even harder to try and understand what goes on there. It mentions ``xnvme_be_family`` but ``xnvme_be_family`` is nowhere to be found.

karlowich:
@safl I see now, I should have made a comment describing the new state of the PR, since the original PR is old and the description is stale.

The headline from the description is still correct:

"This series allows compatible xNVMe backends to share the same NVMe controller, enabling dual-backend use cases such as opening a device with both upcie and upcie-cuda at the same time."

> What constitutes a backend family?

That term is from the stale PR Description. Your first review didn't like the "backend family" term, so now it is simply compatibility bitmask.

> What does it mean to share a "device URI", is the device URI truly the thing being shared

No, they share the `ctrlr` and thus the underlying RTE etc.

>  why is it a bitmask and not just a value?

This is to support one backend being compatible with multiple others. I don't know if we need this, but it seemed like a low complexity way to support the case where:
A is compatible with B and C, but B and C aren't compatible.

> When introducing a compatibility flag why is the name still used as a means to detect compatibility

This is to avoid defining a mask for every single backend, this way we only need to add it for the backends that are actually compatible with others.

> "this is the controller-state i use", and then when doing looksup in cref, then it is looking up on the device-uri and then checking compatibility by checking the controller-format/state identifer.

Yes, this is pretty much exactly what I implemented. "Backend compatibility bitmask" I think this covers what you're describing?

> When you call something "alt-be" my mind goes to "alternative backend", is that what this is? It also looks like it is solely a cli thing? No mentions in the backend? Is it solely a cli/test concern?

Yes. To test this my test needs to know which alternative backend to test against, thus the cli-flag.

> In case that this is an "alternative" backend, what does that mean, when is it in effect? What is the relation primary/secondary or clients/server, what is going on her?

The test just opens the `--be` and the `--alt-be` and carries out the test. There's no hierarchy and the order doesn't matter, it is all handled by the cref infra.

> In the docs something is described as dual-backend but it does not really shed a light on this.

This is on purpose, since the concept is not supposed to change anything by the user. We already support opening `0000:01:00.0` multiple times with the same backend, now we also support opening `0000:01:00.0` with two different backend.

> All of this terminology and concepts are bleeding through to the user, is that really needed? Could it be encapsulated? How does one know what to combine as one or the other?

I don't think that is true, as you say in the comment above, the docs describes very little about this. The docs describe the only thing you can do currently: mixing `upcie` and `upcie-cuda`

> In #686 also does sharing... how do these PRs relate? Are they, or could they, fundamentally solve the same problem?

They are completely different. What I'm introducing here is simply a way to mix `hostmem` and `cudamem` with uPCIe in a generalized way.

safl:
Well then... how about just dropping ``feat(be): allow compatible backends to share a device URI`` and replacing it with :

```c
// Controller state is keyed by backend name, except that upcie variants
// share controller state and accept each other's crefs despite the
// inexact name match.
if (strcmp(cref->be_name, cfg->attr.name) != 0 &&
  !(strstr(cref->be_name, "upcie") && strstr(cfg->attr.name, "upcie"))) {
  xnvme_be_cref_put(cref->ctrlr);
  XNVME_DEBUG("FAILED: uri '%s' claimed by another backend config", dev->ident.uri);
    return -EBUSY;
}
```
?

karlowich:
Okay, I guess there's no reason to make it more complicated than that.

tests/dual_be.c
```diff
@@ -0,0 +1,243 @@
...
+	alt_idfy = xnvme_buf_alloc(alt_dev, sizeof(*alt_idfy));
+	if (!alt_idfy) {
+		err = -errno;
+		xnvme_cli_perr("xnvme_buf_alloc(alt_idfy)", err);
+		goto exit;
+	}
+	xnvme_buf_clear(alt_idfy, sizeof(*alt_idfy));
+
+	ctx = xnvme_cmd_ctx_from_dev(dev);
+	err = xnvme_adm_idfy_ns(&ctx, nsid, idfy);
```
safl, line 157: ``xnvme_adm_idfy_ns`` is invoked yet the commit message and the short/long descriptions say it is doing identify controller.

cijoe/tests/logic/test_xnvme_tests_dual_be.py
```diff
@@ -0,0 +1,43 @@
...
+        return "upcie-cuda"
+    elif be == "upcie-cuda":
+        return "upcie"
+    else:
+        return None
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be"])
+def test_open_dev_twice(cijoe, device, be_opts, cli_args):
+    # We are repurposing this test to verify that opening the same device twice with the same backend works
```
safl: What is ``test_open_dev_twice`` being repurposed from? Is it relevant what it used to be?

safl (commented):
See inline.

cijoe/tests/logic/test_xnvme_tests_dual_be.py
karlowich: We are repurposing `xnvme_tests_dual_be idfy-cmp`. It was intended for `--be A` and `--alt-be B`, however, here we are doing `--be == --alt-be`. This is, to my knowledge, the first direct test of our cref infrastructure.

tests/dual_be.c
karlowich: You're right. Let me update it to say identify namespace everywhere.

safl (approved):
lgtm

naddinadja (approved):
Looks good to me :)

Issue #624, opened 2026-01-27
Controller ref-count issues


safl:
# Background

The current controller reference counting logic has multiple structural and correctness issues.

At a basic level, there are nearly identical implementations of the refcount helpers in both the SPDK and libvfn backends. This duplication is error-prone and highlights the need for generalization. The first step is therefore to eliminate the duplicate implementations. That refactoring is the motivation for this PR:
https://github.com/xnvme/xnvme/pull/623

The logic issues described below are intentionally not addressed in that PR, to avoid mixing refactoring with behavioral fixes. However, they exist today in both backends and should be addressed next.

# Identified issues

1. insert does not check for duplicate URIs  
It is possible to insert the same URI multiple times, creating multiple controller entries for a single URI. This leads to inconsistent and incorrect reference counting.

2. ref can resurrect zero-refcount entries  
The code checks for refcount < 0 but not refcount == 0. As a result, an entry that has been decremented to zero (without DESTROY_IMMEDIATE) can be re-referenced before deferred cleanup runs, effectively reviving a controller that is supposed to be destroyed.

3. insert can overwrite zero-refcount entries  
Free slots are identified by refcount == 0. However, when an entry is dereferenced to zero without DESTROY_IMMEDIATE, the controller pointer is still non-NULL and destruction has not yet occurred. insert may then reuse that slot, leaking the old controller.

Issues 2 and 3 are closely related. Both stem from an ambiguous state where entries with refcount == 0 may still own a valid controller when deferred destruction is used.

4. No cross-backend conflict detection  
lookup and ref return a controller solely based on URI, without considering which backend inserted it. If SPDK holds a controller for a given URI and libvfn performs a lookup for the same URI, it may silently receive an SPDK controller handle, which is incorrect.

5. strncmp vs strcmp for URI comparison  
URIs are null-terminated strings stored in fixed-size buffers. Using strncmp(..., XNVME_IDENT_URI_LEN) can treat two distinct URIs as equal if they share a common prefix of that length. Using strcmp would be simpler and semantically correct

karlowich:
All of these have been addressed.

PR #667, opened 2026-05-07
refactor(be/cref): simplify cref API and remove per-backend enumerate

karlowich:
Simplifies the controller reference-counting API and consolidates the
  enumerate path so backends no longer need to implement it themselves.

  ## cref API changes

  - `cref_lookup(uri, be_name)` → `cref_get(uri, &be_name)`: be_name is
    now an out-parameter rather than a filter, removing the need for a
    separate conflict-detection lookup.
  - `cref_deref(ctrlr, flags)` → `cref_put(ctrlr)`: flags removed;
    destroy-vs-defer is now controlled by explicit `cref_defer(ctrlr)`.
  - `cref_ref()` (combined lookup-or-insert) removed; callers use
    `cref_get` + `cref_insert` explicitly.
  - `cref_cleanup(be_name)` → `cref_cleanup(void)`: the be_name filter
    is no longer needed since deferred state is tracked per-entry.

  ## enumerate consolidation

  Removes the `enumerate` function pointer from `xnvme_be_dev` and all
  backend implementations. Local device enumeration was already handled
  by the platform scan path; fabrics enumeration now goes through
  `enumerate_controller` directly. This eliminates ~200 lines of
  duplicated SPDK-specific enumerate logic.

  ## SPDK backend refactoring

  Splits the old `state_init` / `state_term` into `ctrlr_init` /
  `ctrlr_term`, aligning SPDK with the libvfn and upcie pattern.
  `ctrlr_init` probes and returns the controller handle; namespace
  validation, timeout registration, and qpair setup move to `dev_open`.

  ## Bug fixes

  - `dev_open` was silently skipped on a same-name cref hit.
  - `ctrlr_dev` was not closed on `xnvme_buf_clear` failure in
    `enumerate_controller`.
  - RTE teardown could occur mid-enumeration when the ctrlr_dev was
    closed before all namespace devices were opened; fixed by deferring
    teardown via `cref_defer` and calling `cref_cleanup` once enumeration
    is complete.
  - `cref_insert` scanned for a free slot using `refcount == 0`, which
    is no longer equivalent to an empty slot once deferred entries
    (refcount=0, ctrlr non-NULL) are possible; fixed to check
    `ctrlr == NULL`.
  - `attach_cb` discarded a duplicate controller handle from
    `spdk_nvme_probe` without detaching it.

### DIFF
--- a/lib/xnvme_be_spdk_dev.c
+++ b/lib/xnvme_be_spdk_dev.c
@@ -24,19 +24,6 @@
 #define XNVME_BE_SPDK_MAX_PROBE_ATTEMPTS 1
 #define XNVME_BE_SPDK_AVLB_TRANSPORTS 3
 
-static int
-_spdk_ctrlr_destructor(void *ctrlr)
-{
-	int err;
-
-	err = spdk_nvme_detach((struct spdk_nvme_ctrlr *)ctrlr);
-	if (err) {
-		XNVME_DEBUG("FAILED: spdk_nvme_detach()");
-	}
-
-	return err;
-}
-
 static int g_xnvme_be_spdk_transport[] = {
 #ifdef XNVME_BE_SPDK_TRANSPORT_PCIE_ENABLED
 	SPDK_NVME_TRANSPORT_PCIE,
@@ -361,8 +348,12 @@ timeout_cb_func(void *XNVME_UNUSED(cb_arg), struct spdk_nvme_ctrlr *ctrlr,
 }
 
 /**
- * Sets up the state{ns, ctrlr, attached} given via the cb_ctx
- * detached if dev->nsid is not a match
+ * Records the attached controller in state so ctrlr_init can return it.
+ *
+ * When nsid is set, rejects controllers that do not have the requested namespace active.
+ * This is needed for NVMe-oF auto-discovery, where SPDK may call attach_cb for both
+ * the discovery controller and the actual NVM subsystem controllers; we must skip the
+ * discovery controller (which has no data namespaces) and wait for the right one.
  */
 static void
 attach_cb(void *cb_ctx, const struct spdk_nvme_transport_id *trid, struct spdk_nvme_ctrlr *ctrlr,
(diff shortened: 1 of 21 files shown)


karlowich:
This is fixing some of the concerns in #624 and enabling the progress of #665

lib/xnvme_be_spdk_dev.c
```diff
@@ -371,78 +358,50 @@ attach_cb(void *cb_ctx, const struct spdk_nvme_transport_id *trid, struct spdk_n
...
+		XNVME_DEBUG("FAILED: spdk_nvme_detach()");
 	}
+	return err;
 }
 
 void
 xnvme_be_spdk_dev_close(struct xnvme_dev *dev)
 {
-	if (!dev) {
+	struct xnvme_be_spdk_state *state = (void *)dev->be.state;
```
naddinadja, line 399: You remove the check for `dev` being null, and might therefore dereference a null-pointer. Idk if it's "OK", but it would probably be good practice to still fail on `(!dev)`

lib/xnvme_be_spdk_dev.c
```diff
@@ -663,99 +622,36 @@ verify_ctrlr_ok(struct spdk_nvme_ctrlr *ctrlr)
...
-
 	if (dev->opts.core_mask) {
 		XNVME_DEBUG("INFO: multi-process setup");
 		env_opts.shm_id = dev->opts.shm_id;
 		env_opts.core_mask = dev->opts.core_mask;
 		env_opts.main_core = dev->opts.main_core;
 	}
 
-	err = _spdk_env_init(&env_opts);
-	if (err) {
```
naddinadja, line 689: I think i liked it better when the function call was outside the if statement.

lib/xnvme_be_spdk_dev.c
```diff
@@ -663,99 +622,36 @@ verify_ctrlr_ok(struct spdk_nvme_ctrlr *ctrlr)
...
- * - Parse options from dev->ident
- * - Initialize SPDK environment
- * - Attach to controller matching dev->ident
- * - create sync-io-qpair
- * - create lock protecting sync-io-qpair
+ * Initialize a new SPDK controller for the device URI.
+ *
+ * Inits the SPDK environment and probes for the matching device. The returned
+ * ctrlr handle is stored in cref and written to dev->be.state[0] by the
+ * platform. Namespace validation and qpair setup are deferred to dev_open.
```
naddinadja: Do we need to specify the last sentence here? Sounds like an explanation of the changes in this PR rather than an explanation of the function

lib/xnvme_be_spdk_dev.c
```diff
@@ -770,53 +666,84 @@ xnvme_be_spdk_state_init(struct xnvme_dev *dev)
...
+int
+xnvme_be_spdk_dev_open(struct xnvme_dev *dev)
+{
+	struct xnvme_be_spdk_state *state = (void *)dev->be.state;
+	int err;
+
+	// ctrlr is at state->ctrlr, placed there by platform (cref reuse or ctrlr_init).
+	// For reused controllers, verify health before setting up a new qpair.
+	if (!state->attached) {
+		int check = verify_ctrlr_ok(state->ctrlr);
```
naddinadja: why not use `err` here?

lib/xnvme_be_spdk_dev.c
```diff
@@ -770,53 +666,84 @@ xnvme_be_spdk_state_init(struct xnvme_dev *dev)
...
+		if (check < 0) {
+			XNVME_DEBUG("FAILED: verify_ctrlr_ok, err: %d", check);
+		}
+		if (check < 0 || spdk_nvme_ctrlr_is_failed(state->ctrlr)) {
+			if (reconnect_ctrlr(state->ctrlr) < 0) {
+				return -EBUSY;
+			}
```
naddinadja: Do we want a debug log here?

naddinadja (changes requested):
A few comments here

lib/xnvme_be_spdk_dev.c
karlowich: Well, this function can only be called by `xnvme_dev_close()` which has the guard. But, I don't mind adding the guard here too.

lib/xnvme_be_spdk_dev.c
karlowich: Sure, will fix that

lib/xnvme_be_spdk_dev.c
karlowich: I think it should just be removed

lib/xnvme_be_spdk_dev.c
karlowich: We should

lib/xnvme_be_spdk_dev.c
karlowich: Sure

karlowich: I decided to remove this path. Since our code allows for reusing controllers, we should be resetting the controller even if it is in a failed state. Doing so, would affect all handles to the controller without notifying the user - instead we should simply fail and let the user figure out what they want to do.

lib/xnvme_be_spdk_dev.c
karlowich: fixed

lib/xnvme_be_spdk_dev.c
karlowich: fixed

lib/xnvme_be_spdk_dev.c
karlowich: fixed

lib/xnvme_be_spdk_dev.c
karlowich: fixed

safl (commented):
Thanks for picking up the cref feedback from #665. I want to land this with
you, but the shape of the PR is making it hard to give it the review it
deserves, and I would rather pause than fall back on a shallow AI-assisted
pass.

The description names four topics. A few things I want us to talk through:

**1. cref direction.** `_lookup` to `_get` (`f5af087`) looks like a clean
simplification. `xnvme_be_cref_defer` (`0290b8e`) goes the other way: new
API, new per-entry `deferred` state, a new ordering constraint on
`cleanup()`, and exactly one caller (`lib/xnvme_platform.c:338`). Given #665
went to draft to reduce cref complexity, doing the opposite here is
surprising.

**2. Bug fixes bundled with the refactor.** Four `fix(...)` commits ride
along (`980d1aa`, `a063150`, `8c84538`, `ec814b4`). They deserve their own
review. The RTE-teardown fix depends on the new `_defer` API, which suggests
the API was shaped to fit the fix. Can we explore solving it without
`_defer`?

**3. SPDK delegation deserves its own thread.** `ea788f0` drops around 460
lines from `lib/xnvme_be_spdk_dev.c`, likely the most valuable change in the
series. It should not share oxygen with cref API discussion.

**4. Minor: ASCII commit subjects.** `a7e5093` uses a unicode arrow (U+2192)
for both arrows in the subject. Please don't.

**Could we split this?** Rough cut:

- A: enumeration consolidation + dev_open simplification.
- B: SPDK lifecycle delegation.
- C: standalone fixes.
- D: cref API + `_defer` + RTE-teardown fix. This is the one that needs the most scrutiny.

I just want us to land each piece on its own merits, especially given the
cref and SPDK lifecycle are complicated and error-prone areas.

karlowich:
@safl Thanks for the input, I will split this PR into 4 as you described. The order matters, so I won't open all PRs at once.
Let's start from this: https://github.com/xnvme/xnvme/pull/672

karlowich:
The final PR (#683 ) from the split is up, thus closing this.

PR #671, opened 2026-05-11
ci(tests): experiment — move pytest.skip to collection-time to reduce CI wall time

karlowich:
Each `pytest.skip()` call that fires inside a test body still pays the full
  fixture setup cost: the `device` fixture runs, which calls
  `XnvmeDriver.attach()` and optionally spins up a fabrics target over SSH
  (~0.22 s per invocation). Across hundreds of parametrised combinations this
  adds up to non-trivial overhead on every CI run.

  This PR is an experiment to measure whether moving known-incompatible
  combinations out of the test body and into collection time makes a meaningful
  difference to wall-clock CI duration.

  ## What changed

  Two new parameters are added to `@xnvme_parametrize` (and the underlying
  `xnvme_setup`):

  - `exclude={}` — drop combinations at collection time based on backend option
    values (`sync`, `async`, `admin`, `mem`, `be`). A combination is dropped if
    *any* key/value matches (OR semantics).
  - `os_exclude={}` — same, but only applied when the current OS (resolved from
    the CIJOE config at collection time) matches the key.

  These replace all `pytest.skip()` calls in test bodies that were conditioned
  purely on `be_opts` values or on `(OS, be_opts)` pairs. Three FreeBSD patterns
  are covered by `os_exclude`:

  | `os_exclude` key | Values excluded | Reason |
  |---|---|---|
  | `be` | `kqueue`, `thrpool`, `emu` | FreeBSD kernel doesn't support ZNS |
  | `sync` | `nvme` | FreeBSD nvme sync does not implement iovec |
  | `admin` | `nvme` | FreeBSD nvme admin does not support metadata/PI |

  Runtime skips that depend on device labels or other runtime state (e.g.
  `"cdev" not in device["labels"]`, spdk fabrics format check) are left as-is.

  A comment above each `@xnvme_parametrize` decorator documents the reason for
  each exclusion so the information is not lost.

  ## What is not changed

  - Test correctness — the same combinations that were previously skipped at
    runtime are now simply never scheduled.
  - OS-detection based behaviour that is not a skip (e.g. `test_scan`,
    `test_show_regs`) is untouched.
  - The two `!= libvfn` skips (`test_xnvme_single_async_efd`,
    `test_xnvme_tests_map`) require AND/include semantics not supported by the
    current `exclude` design and are left for a follow-up if the experiment
    proves worthwhile.

### DIFF
--- a/cijoe/tests/apps/test_fio.py
+++ b/cijoe/tests/apps/test_fio.py
@@ -10,16 +10,16 @@ FIO_OUTPUT_FPATH = (
 )
 
 
-@xnvme_parametrize(labels=["dev"], opts=["be", "admin", "sync", "async", "mem"])
+# mem=[upcie-cuda]: fio engine does not support CUDA memory
+@xnvme_parametrize(
+    labels=["dev"],
+    opts=["be", "admin", "sync", "async", "mem"],
+    exclude={"mem": ["upcie-cuda"]},
+)
 def test_fio_engine(cijoe, device, be_opts, cli_args):
     """
     The construction of the fio-invocation is done in 'fio_fancy'
     """
-    if be_opts["mem"] == "upcie-cuda":
-        pytest.skip(
-            reason="[mem=upcie-cuda] fio xNVMe ioengine does not support CUDA memory"
-        )
-
     size = "64M"
     # size = "1G"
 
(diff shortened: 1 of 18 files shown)


karlowich:
This did very little (if anything) for the running time of the CI. However, it did remove ~400 skipped tests from the report 

From next:
Screenshot 2026-05-11 at 17 14 51

With the changes:
Screenshot 2026-05-11 at 17 16 03

karlowich:
Closing this, as it is not relevant for now.

PR #672, opened 2026-05-12
be/spdk: delegate controller lifecycle to platform cref machinery

karlowich:
The SPDK backend previously managed the shared controller reference
  itself, calling xnvme_be_cref_* directly from within state_init/state_term
  and its enumerate path. This bypasses the platform's cref machinery and
  makes the lifecycle hard to reason about when multiple devices share a
  controller.

  Move controller probe and detach into ctrlr_init / ctrlr_term hooks so
  the platform owns the controller reference from creation to destruction:

  - Split state_init into ctrlr_init (env init + SPDK probe) and dev_open
    (namespace validation + qpair setup)
  - Strip attach_cb down to recording the ctrlr handle only; ns validation
    and timeout registration move to dev_open; duplicate ctrlr from a
    second attach_cb invocation is detached immediately to avoid a leak
  - Remove state_term; dev_close now only tears down the qpair
  - Replace _spdk_ctrlr_destructor with ctrlr_term
  - Reorder xnvme_be_spdk_state so ctrlr is at offset 0 (platform
    convention)

  Also tighten verify_ctrlr_ok: check spdk_nvme_ctrlr_is_failed early,
  check the return value of spdk_nvme_ctrlr_cmd_admin_raw, and fix err to
  reflect the NVMe completion status on the success path. Remove
  reconnect_ctrlr: resetting a shared ctrlr invalidates qpairs of other
  devices holding the same cref handle — fail fast and let cref evict the
  stale entry instead.

### DIFF
--- a/include/xnvme_be_spdk.h
+++ b/include/xnvme_be_spdk.h
@@ -39,20 +39,21 @@ XNVME_STATIC_ASSERT(sizeof(struct spdk_nvme_ns_data) == sizeof(struct xnvme_spec
 		    "Incorrect size")
 
 struct xnvme_be_spdk_state {
-	union {
-		pthread_mutex_t qpair_lock; ///< LOCK for SYNC IO commands
-		uint8_t _fill[64];
-	};
-	struct spdk_nvme_qpair *qpair; ///< QPAIR for SYNC IO commands
-
-	struct spdk_nvme_ctrlr *ctrlr; ///< Pointer to attached controller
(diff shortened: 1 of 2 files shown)


karlowich:
This depends on #673

safl (approved):
lgtm. Note to self, we need to re-visit hot-plug-management once all backends / user-space drivers follow the same pattern.

naddinadja (approved):
LGTM :)

PR #673, opened 2026-05-13
platform: fix three bugs in the dev-open and enumerate paths

karlowich:
Three independent correctness fixes to _dev_open_init_cref and
  enumerate_controller:

  1. Same-name cref hit skipped dev_open. When _dev_open_init_cref found
     an existing cref entry for the same URI and backend, it stored the
     ctrlr pointer and returned 0 without calling dev_open. Per-handle
     resources initialised in dev_open (e.g. atomic counters) were left
     uninitialised for every open after the first.

  2. _dev_open_init_cref duplicated the ctrlr_init check. The early
     return for backends without ctrlr_init was also present in the
     caller, making _dev_open_init_cref dead for those backends. Removing
     the duplicate caller check makes the function the single owner of
     ctrlr_init logic.

  3. ctrlr_dev leaked on xnvme_buf_clear failure in enumerate_controller.
     The error path freed the identify buffer but returned without closing
     ctrlr_dev, leaving the controller reference open.
     
This should fix the failure we're seeing in #672

### DIFF
--- a/lib/xnvme_platform.c
+++ b/lib/xnvme_platform.c
@@ -78,7 +78,6 @@ static int
 _dev_open_init_cref(struct xnvme_dev *dev, struct xnvme_be *be, const struct xnvme_be_config *cfg)
 {
 	void *ctrlr = NULL;
-	int cref_inserted = 0;
 	int err;
 
 	if (!be->dev.ctrlr_init) {
@@ -87,8 +86,7 @@ _dev_open_init_cref(struct xnvme_dev *dev, struct xnvme_be *be, const struct xnv
 
 	ctrlr = xnvme_be_cref_lookup(dev->ident.uri, cfg->attr.name);
 	if (ctrlr) {
-		((void **)be->state)[0] = ctrlr;
-		return 0;
+		goto ctrlr_ready;
 	}
 
 	{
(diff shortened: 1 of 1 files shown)


naddinadja (dismissed):
lgtm! :)

lib/xnvme_platform.c
safl, line 115:
```suggestion
ctrlr_ready:
```

lib/xnvme_platform.c
```diff
@@ -82,13 +82,17 @@ _dev_open_init_cref(struct xnvme_dev *dev, struct xnvme_be *be, const struct xnv
 	int err;
 
 	if (!be->dev.ctrlr_init) {
-		return 0;
+		return be->dev.dev_open(dev);
 	}
 
 	ctrlr = xnvme_be_cref_lookup(dev->ident.uri, cfg->attr.name);
 	if (ctrlr) {
```
safl, line 88:
```suggestion
	if (ctrlr) {
	  goto ctrlr_ready;
```

safl (changes requested):
Looks great, awesome! One small suggestions to combine the initiatlization-path with a label.

lib/xnvme_platform.c
karlowich: Done

lib/xnvme_platform.c
karlowich: Done

naddinadja (approved):
lgtm still! :)

safl (approved):
Great!

PR #679, opened 2026-05-26
fix(be): respect config-selected backends in dev_open across platforms

### LINKED ISSUE #678: `be=io_uring` doesn't select the correct async interface
When running `xnvme info /dev/nvme0n1 --be io_uring` I'm seeing the following:
```
xnvme_dev:
...
  xnvme_be:
    dev: {id: 'linux'}
    admin: {id: 'nvme'}
    sync: {id: 'nvme'}
    async: {id: 'emu'}
    mem: {id: 'posix'}
    attr: {name: 'io_uring', descr: 'io_uring with NVMe ioctl', caps: 0x3}
...
```
Notice: 
attr.name = 'io_uring'   <- right config selected
async.id  = 'emu'        <- but async impl is emu, not liburing

It's unclear how this happens, since the backend config is correctly setup in the `lib/xnvme_be_linux.c` file.

karlowich:
Closes #678.

Every platform's `dev_open` implementation wrote `dev->be.{async,sync,admin}`
inside `if (!opts->{async,sync,admin}) { ... }` guards, defaulting to specific
backend implementations. These defaults predate the flat-config refactor:
`_dev_open_try_config()` in `lib/xnvme_platform.c` now populates
`dev->be.{async,sync,admin}` directly from the matched `xnvme_be_config`
before invoking `dev_open`. The overrides therefore silently clobber the
config's choice whenever the caller did not pass `opts->{async,sync,admin}`
explicitly — the common case for the CLI tools and any consumer that selects a
backend via `opts->be`.

The visible bug (Linux): `xnvme_dev_open(uri, { .be = "io_uring" })` selected
the `io_uring` config (async impl `g_xnvme_be_linux_async_liburing`), then
`xnvme_be_linux_dev_open` overwrote `dev->be.async` with
`g_xnvme_be_cbi_async_emu`. Every "async" submission ran through emu's
threadpool into the sync NVMe ioctl backend — producing hundreds of thousands
of ioctls/sec where io_uring SQEs were expected.

The core change removes these stale overrides on Linux, FreeBSD, macOS,
ramdisk, and Windows, so the matched config is the single source of truth for
backend selection.

Dropping the overrides removed a layer that had been silently papering over
several latent bugs: backends were being swapped out from under callers, which
masked misconfigured configs, an unsafe assumption in the async copy, and
Windows path-classification gaps. With the override gone, the test suite goes
red until each of these is fixed, so they are bundled here to keep CI green. 
They are split into self-contained commits, grouped below.

### Core fix — remove the overrides

- `fix(be/linux)`: stop overriding config-selected backends in `dev_open`
  (the bug above).
- `fix(be/fbsd)`, `fix(be/macos)`, `fix(be/ramdisk)`, `fix(be/windows)`:
  the same removal on the remaining platforms.
- `style(be/linux)`: unify the per-file-type debug prints in
  `xnvme_be_linux_dev_open` (cosmetic, touched while in the area).

On Windows, one deliberate exception remains: the S_IFCHR admin override for
the `BusTypeScsi` / `BusTypeSata` case is kept, because no Windows config
exposes `g_xnvme_be_windows_admin_block` and the configs all set `admin_nvme`,
so a SCSI/SATA disk opened by URI would otherwise land on NVMe admin commands.

### Latent bugs exposed by the removal

- `fix(be): classify non-stattable URIs as files` — a not-yet-created URI
  (e.g. a copy destination) matched no ramdisk/PCI/fabrics pattern and fell
  through to the first (NVMe) config, whose sync backend cannot service file
  I/O. Default such URIs to `XNVME_BE_CAP_FILE`, mirroring Windows `classify()`.
- `fix(be): use CBI shim admin for FreeBSD file backend` — `kqueue_file` was
  wired to the NVMe-ioctl admin, which cannot identify a regular file, so
  geometry derivation failed with `EIO`. Route it to the CBI shim admin and
  make it `CAP_FILE`-only, mirroring the Linux `*_file` configs.
- `fix(be): default FreeBSD file I/O to emulated async` — `kqueue_file`
  (kqueue/POSIX-aio) was the only `CAP_FILE` backend on FreeBSD; POSIX aio
  completes out of order, which the xdd async copy did not tolerate. Add an
  `emu_file` config and list it ahead of `kqueue_file` so file URIs select it
  by default (`kqueue_file` stays available by explicit name).
- `fix(xdd): don't assume in-order completion in async copy` — `copy_async`
  picked a buffer slot by `block % qdepth`, only safe under in-order
  completion. Track a per-slot `in_use` flag and pick a free slot per
  submission, making the copy correct regardless of completion ordering. This
  is an independent correctness bug; the FreeBSD POSIX-aio backend just made it
  observable.

### Windows device-path handling

- `fix(platform): classify forward-slash device paths on Windows as bdev` and
  `fix(be/windows): match forward-slash PhysicalDrive URIs in fstat fallback` —
  `//./PhysicalDriveN` resolves identically to `\\.\PhysicalDriveN` in
  `CreateFile()`, but `classify()` and the fstat fallback only matched the
  backslash form. The backslash form does not survive shell evaluation over
  SSH, so fio received `.PhysicalDrive1` and silently benchmarked a regular
  file in its cwd. Recognise the forward-slash form, which needs no escaping.
- `fix(cijoe/configs): use forward-slash device path for the Windows DUT` —
  switch the DUT URI to `//./PhysicalDrive1` for the same reason.

### Test and benchmark config plumbing

- `test(cijoe): skip NVMe-only commands on [admin=block]` — the block-layer
  sync/admin backends only service read/write/flush and zone-management;
  compare, write-zeroes, dsm and zone-append return `-ENOSYS`. These previously
  passed only because `dev_open` swapped block backends for NVMe-ioctl ones.
- `fix(cijoe/configs): select fio xnvme backend by config name` — the fio
  engines selected `be = "<platform>"`, which is not a config name and no
  longer resolves. Point `be` at the real config name and drop the redundant
  `async` field.
- `refactor(cijoe/scripts): drop redundant async from fio latency engines` —
  config name now fully determines the async backend, so remove the `async_`
  field / `--xnvme_async` flag and derive `name_id` from `be` (renames the
  Windows engine label `xnvme_io_ring` → `xnvme_ioring`).

### DIFF
--- a/cijoe/auxiliary/plot-latency-legends.yaml
+++ b/cijoe/auxiliary/plot-latency-legends.yaml
@@ -42,7 +42,7 @@ xnvme_null:
   legend: 'null'
 
 # Windows
-xnvme_io_ring:
+xnvme_ioring:
   legend: 'xnvme(ioring)'
 
 xnvme_iocp:
--- a/cijoe/configs/ax42-freebsd.toml
+++ b/cijoe/configs/ax42-freebsd.toml
@@ -114,10 +114,10 @@ null = {type = "builtin", group = "null", device = "bdev"}
 spdk_nvme = {type = "external_preload", group = "spdk", path = "/tmp/spdk/build/fio/spdk_nvme", device = "pcie"} # Userspace
 spdk_bdev = {type = "external_preload", group = "spdk", path = "/tmp/spdk/build/fio/spdk_bdev", device = "pcie"} # Userspace
 
-xnvme_kqueue = {type = "builtin", group = "posixaio", be = "freebsd", async = "kqueue", device = "bdev"} # Kernel
-xnvme_posixaio = {type = "builtin", group = "posixaio", be = "freebsd", async = "posix", device = "bdev"} # Kernel
-xnvme_spdk = {type = "builtin", group = "spdk", be = "spdk", async = "spdk", device = "pcie"} # Userspace
-xnvme_null = {type = "builtin", group = "null", be = "freebsd", async = "nil", device = "bdev"} # Kernel
+xnvme_kqueue = {type = "builtin", group = "posixaio", be = "kqueue", device = "bdev"} # Kernel
+xnvme_posixaio = {type = "builtin", group = "posixaio", be = "posix", device = "bdev"} # Kernel
+xnvme_spdk = {type = "builtin", group = "spdk", be = "spdk", device = "pcie"} # Userspace
+xnvme_null = {type = "builtin", group = "null", be = "nil", device = "bdev"} # Kernel
 
 [[devices]] # OS Managed
 key = "bdev"
--- a/cijoe/tests/logic/test_xnvme_tests_znd_append.py
+++ b/cijoe/tests/logic/test_xnvme_tests_znd_append.py
@@ -11,5 +11,7 @@ def test_verify(cijoe, device, be_opts, cli_args):
         "ramdisk_thrpool",
     ]:
         pytest.skip(reason="Freebsd kernel doesn't support zns")
+    if be_opts["admin"] == "block":
+        pytest.skip(reason="[admin=block] does not implement zone-append")
     err, _ = cijoe.run(f"xnvme_tests_znd_append verify {cli_args}")
     assert not err
--- a/lib/xnvme_be_fbsd.c
+++ b/lib/xnvme_be_fbsd.c
@@ -24,6 +24,20 @@ const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_nvme = {
 		},
 };
 
+const struct xnvme_be_config g_xnvme_be_fbsd_emu_file = {
+	.async = &g_xnvme_be_cbi_async_emu,
+	.sync = &g_xnvme_be_cbi_sync_psync,
+	.admin = &g_xnvme_be_cbi_admin_shim,
+	.dev = &g_xnvme_be_fbsd_dev,
+	.mem = &g_xnvme_be_cbi_mem_posix,
+	.attr =
+		{
+			.name = "emu_file",
+			.descr = "Emulated async with file I/O",
+			.caps = XNVME_BE_CAP_FILE,
+		},
+};
+
 const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_psync = {
 	.async = &g_xnvme_be_fbsd_async,
 	.sync = &g_xnvme_be_cbi_sync_psync,
--- a/lib/xnvme_be_fbsd.c
+++ b/lib/xnvme_be_fbsd.c
@@ -27,14 +27,14 @@ const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_nvme = {
 const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_psync = {
 	.async = &g_xnvme_be_fbsd_async,
 	.sync = &g_xnvme_be_cbi_sync_psync,
-	.admin = &g_xnvme_be_fbsd_admin_nvme,
+	.admin = &g_xnvme_be_cbi_admin_shim,
 	.dev = &g_xnvme_be_fbsd_dev,
 	.mem = &g_xnvme_be_cbi_mem_posix,
 	.attr =
 		{
 			.name = "kqueue_file",
 			.descr = "kqueue with POSIX sync",
-			.caps = XNVME_BE_CAP_NVME_CDEV | XNVME_BE_CAP_FILE,
+			.caps = XNVME_BE_CAP_FILE,
 		},
 };
 
--- a/lib/xnvme_be_fbsd_dev.c
+++ b/lib/xnvme_be_fbsd_dev.c
@@ -88,15 +88,6 @@ xnvme_be_fbsd_dev_open(struct xnvme_dev *dev)
 		dev->ident.dtype = XNVME_DEV_TYPE_FS_FILE;
 		dev->ident.csi = XNVME_SPEC_CSI_FS;
 		dev->ident.nsid = 1;
-		if (!opts->admin) {
-			dev->be.admin = g_xnvme_be_cbi_admin_shim;
-		}
-		if (!opts->sync) {
-			dev->be.sync = g_xnvme_be_cbi_sync_psync;
-		}
-		if (!opts->async) {
-			dev->be.async = g_xnvme_be_cbi_async_emu;
-		}
 		state->fd.ctrlr = state->fd.ns;
 		break;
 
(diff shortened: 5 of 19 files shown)


safl:
Awaiting resolution of https://www.githubstatus.com/incidents/gnftqj9htp0g and https://github.com/xnvme/xnvme/pull/677
Then I will take a look this.

naddinadja (dismissed):
lgtm! :)

safl (changes requested):
Really nice find, looks sound, can you rebase it?

safl:
The freebsd failure looks like a CI regression, however the windows part is not.

karlowich:
> the windows part is not

I'm fully aware, been trying to figure it out since yesterday. We had some very tricky / unintuitive stuff going on with how backends are chosen which I'm trying to unravel. Somehow this meant breaking things left and right.

safl:
This should fix the freebsd regression: https://github.com/xnvme/xnvme/pull/687

safl:
I should get my morning coffee before interacting with github...

naddinadja (dismissed):
lgtm :)

lib/xnvme_be_linux_dev.c
```diff
@@ -81,26 +81,24 @@ xnvme_be_linux_dev_open(struct xnvme_dev *dev)
...
+		XNVME_DEBUG("INFO: open() : retrieving nsid, got: %x", err);
 		if (err < 1) {
-			XNVME_DEBUG("INFO: open() : retrieving nsid, got: {nsid: %x, err: %d}",
-				    err, err);
 			break;
 		}
 		dev->ident.dtype = XNVME_DEV_TYPE_NVME_NAMESPACE;
 		dev->ident.csi = XNVME_SPEC_CSI_NVM;
 		dev->ident.nsid = err;
-		XNVME_DEBUG("INFO: open() : block-device file (is a NVMe namespace)");
```
safl, line 125: By removing this, then it is no longer communicated that the block device is an NVMe namespace. That seems like a regression.

lib/xnvme_be_fbsd_dev.c
```diff
@@ -88,15 +88,6 @@ xnvme_be_fbsd_dev_open(struct xnvme_dev *dev)
 		dev->ident.dtype = XNVME_DEV_TYPE_FS_FILE;
 		dev->ident.csi = XNVME_SPEC_CSI_FS;
 		dev->ident.nsid = 1;
-		if (!opts->admin) {
```
safl, line 91: Great that you caught this.

lib/xnvme_be_linux_dev.c
```diff
@@ -71,37 +71,19 @@ xnvme_be_linux_dev_open(struct xnvme_dev *dev)
...
 
-	// Change {async,sync,admin} based on file unless one was explicitly requested
+	// Refine identifier fields based on the underlying file type
 	switch (dev_stat.st_mode & S_IFMT) {
 	case S_IFREG:
 		XNVME_DEBUG("INFO: open() : regular file");
 		dev->ident.dtype = XNVME_DEV_TYPE_FS_FILE;
 		dev->ident.csi = XNVME_SPEC_CSI_FS;
 		dev->ident.nsid = 1;
-		if (!opts->admin) {
```
safl, line 81: Great that you caught this.

lib/xnvme_be_macos_dev.c
```diff
@@ -200,16 +200,6 @@ xnvme_be_macos_dev_open(struct xnvme_dev *dev)
 		dev->ident.nsid = _get_nvme_ns(ioservice_device);
 		dev->ident.dtype = XNVME_DEV_TYPE_BLOCK_DEVICE;
 		dev->ident.csi = XNVME_SPEC_CSI_NVM;
-
-		if (!opts->admin) {
```
safl, line 204: Great that you caught this.

lib/xnvme_be_ramdisk_dev.c
```diff
@@ -66,16 +65,6 @@ xnvme_be_ramdisk_dev_open(struct xnvme_dev *dev)
 		return -errno;
 	}
 
-	if (!opts->admin) {
```
safl, line 69: Great that you caught this.

lib/xnvme_be_windows_dev.c
```diff
@@ -154,15 +153,6 @@ xnvme_be_windows_dev_open(struct xnvme_dev *dev)
 		dev->ident.dtype = XNVME_DEV_TYPE_FS_FILE;
 		dev->ident.csi = XNVME_SPEC_CSI_FS;
 		dev->ident.nsid = 1;
-		if (!opts->admin) {
```
safl, line 157: Great that you caught this.

lib/xnvme_platform_freebsd.c
```diff
@@ -63,7 +63,9 @@ xnvme_platform_fbsd_classify(const char *uri)
 		return XNVME_BE_CAP_NVME_TCP;
 	}
 
-	return 0;
+	/* A path we could not stat() and which matched nothing else is assumed
```
safl, line 66: File-create would trigger here as well?

cijoe/tests/logic/test_xnvme_tests_znd_append.py
```diff
@@ -11,5 +11,7 @@ def test_verify(cijoe, device, be_opts, cli_args):
         "ramdisk_thrpool",
     ]:
         pytest.skip(reason="Freebsd kernel doesn't support zns")
+    if be_opts["admin"] == "block":
```
safl, line 14: Great that you caught this.

cijoe/configs/ax42-freebsd.toml
```diff
@@ -114,10 +114,10 @@ null = {type = "builtin", group = "null", device = "bdev"}
 spdk_nvme = {type = "external_preload", group = "spdk", path = "/tmp/spdk/build/fio/spdk_nvme", device = "pcie"} # Userspace
 spdk_bdev = {type = "external_preload", group = "spdk", path = "/tmp/spdk/build/fio/spdk_bdev", device = "pcie"} # Userspace
 
-xnvme_kqueue = {type = "builtin", group = "posixaio", be = "freebsd", async = "kqueue", device = "bdev"} # Kernel
```
safl, line 117: Great that you caught this.

cijoe/auxiliary/plot-latency-legends.yaml
```diff
@@ -42,7 +42,7 @@ xnvme_null:
   legend: 'null'
 
 # Windows
-xnvme_io_ring:
```
safl, line 45: Great that you caught this.

lib/xnvme_be_fbsd.c
```diff
@@ -27,14 +27,14 @@ const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_nvme = {
 const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_psync = {
 	.async = &g_xnvme_be_fbsd_async,
 	.sync = &g_xnvme_be_cbi_sync_psync,
-	.admin = &g_xnvme_be_fbsd_admin_nvme,
+	.admin = &g_xnvme_be_cbi_admin_shim,
```
safl, line 44: Great that you caught this.

lib/xnvme_be_fbsd.c
```diff
@@ -24,6 +24,20 @@ const struct xnvme_be_config g_xnvme_be_fbsd_kqueue_nvme = {
 		},
 };
 
+const struct xnvme_be_config g_xnvme_be_fbsd_emu_file = {
```
safl, line 27: excellent.

safl (dismissed):
Amazing. This is a huge amount of technical debt semingly purged. Awesome. I have one small thing inline, but it is not really worth a respin.

lib/xnvme_platform_freebsd.c
karlowich: This (on FreeBSD) makes sure that we open with `xnvme_be_fbsd_dev_open()`. This calls `open()` which creates the file if it doesn't exist.

lib/xnvme_be_linux_dev.c
karlowich: Yes, let me fix that

karlowich: Actually, I will just drop that commit entirely.

karlowich:
Just dropped the irrelevant style commit and rebased.

safl (approved):
Excellent!

naddinadja (approved):
great :)

PR #680, opened 2026-05-29
be: remove dead enumerate field and fix stale docstrings

karlowich:
xnvme_enumerate() has routed through the platform's scan +
  enumerate_controller path for some time. The per-backend enumerate
  function pointer in struct xnvme_be_dev is never called through the
  public API, but every backend still carried an assignment and SPDK/uPCIe
  had non-trivial implementations that were dead code.

  Remove the enumerate field from struct xnvme_be_dev, drop
  xnvme_be_nosys_enumerate, and remove the custom enumerate
  implementations from the SPDK and uPCIe backends.

  Also fix two docstring issues that became visible during this cleanup:
  the xnvme_enumerate_cb return value description incorrectly documented
  generic errno convention instead of the KEEP_OPEN/CLOSE values, and
  xnvme_platform_enumerate still referenced per-backend enumerate
  delegation that no longer exists.

### DIFF
--- a/include/libxnvme_dev.h
+++ b/include/libxnvme_dev.h
@@ -44,11 +44,11 @@ int
 xnvme_dev_pr(const struct xnvme_dev *dev, int opts);
 
 /**
- * Signature of callback function used with 'xnvme_enumerate' invoked for each discoved device
+ * Signature of callback function used with 'xnvme_enumerate', invoked for each discovered device
  *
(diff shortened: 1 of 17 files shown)


naddinadja (approved):
This looks good to me - maybe it needs a rebase, because it's showing the first two commits that have already been merged in PR #672

safl (approved):
Awesome.

PR #681, opened 2026-05-29
KV via vfu_kvssd over vfio-user and switch to nosi images

safl:
Use vfu_kvssd (vfio-user) for KV when QEMU lacks native kv=on; add [kvssd] config, odus transport and root_unlock for nosi guests. Inert on the fork


safl:
See #682

safl:
See #682

PR #682, opened 2026-05-30
ci: retire the SamsungDS QEMU fork

safl:
This PR touches the foundation of xNVMe's CI: the QEMU binary, the
container that bootstraps QEMU, the verify jobs, the MAAS benchmark
runs, and the documentation generation. Migrating off the SamsungDS
fork means changing all of those at once, since they were chained
together by the fork's native NVMe KV support.

The changeset is therefore relatively large, but mostly mechanical:
container swaps, two new helper scripts that bridge the upstream
image's defaults to the existing CIJOE workflows (guest-disk staging
and root-unlock), and config table additions that activate them.

## What changes

Retire the SamsungDS/qemu `for-xnvme` fork from xNVMe's CI. The fork
was unmaintained and only existed for its native NVMe KV support;
upstream QEMU 10.x covers every other feature the suite exercises
(ZNS, FDP, PI, copy, write_zeroes, fabrics, large_mdts, ...) and the
KV namespace is now supplied by a small out-of-process [vfu_kvssd][1]
device attached over vfio-user.

After this lands, no CI job references the fork container; the fork
docker image (`ghcr.io/xnvme/xnvme-qemu`) can be archived.

  [1]: https://github.com/safl/vfio-user-kvssd

## Commits

Reading order is bottom up (oldest first):

**Preparation** — adds the helpers the migration commit wires in. Each
is inert on its own: the new scripts only run when the configs grow the
matching tables.
- `toolbox: add nosi-related helpers`
- `cijoe(scripts): support external vfu_kvssd in xnvme_guest_start_nvme`
- `cijoe(scripts): add root_unlock helper`
- `cijoe(scripts): add stage_nosi_guest helper`

**The migration** — the actual swap. Host containers move to nosi,
guest disks pivot to the nosi qcow2s, the prep scripts get activated.
- `ci: migrate docgen + verify to upstream QEMU on nosi`
- `ci: migrate verify-maas + benchmark-{bdev,latency} to nosi`

**Test infrastructure adaptations** — surfaced or forced by the
migration.
- `test(xnvme): skip subsystem-reset when the controller cannot perform it`
  (upstream QEMU has `CAP.NSSRS=0`)
- `test(conftest): order tests by backend` (mitigate driver-swap churn
  the new path stresses harder)
- `test(py): correct parametrize matrix and assertion messages` (test
  infra bugs that masked migration failures during the investigation)

**Docs**
- `docs(ci): describe the nosi container + guest-image inventory`

**Cleanup** — orphans created or revealed by the migration.
- `ci: drop bootimg-* helpers superseded by nosi`
- `ci: remove orphaned freebsd-13 + ramdisk-freebsd configs`
- `ci: remove orphaned configs and workflows`
- `ci(macos): reach verify-macos by short hostname` (unrelated drive-by,
  but harmless and small)

## Follow-ups (not in this PR)

- **CAP.NSSRS=0**: upstream QEMU doesn't expose `NSSR` on its emulated
  NVMe; would need an upstream patch (same shape as the `mdts=0` work).
- **`mdts=0`**: patches sent upstream restoring `mdts=0` as "no limit";
  once they land, the `large_mdts = 0 if ENABLE_KV else 9` gate in
  `xnvme_guest_start_nvme.py` drops.
- **FreeBSD ZRWA cdev**: FreeBSD's `nvme(4)` driver does not expose ZNS
  namespaces as character devices under upstream QEMU. The pcie BDF
  stays in `freebsd-14.toml` for `test_scan` parity; the cdev re-adds
  once FreeBSD changes.
- **libvfio-user range-spanning DMA_UNMAP**: vfu_kvssd ships a local
  patch against the embedded libvfio-user submodule
  (`patches/0001-libvfio-user-dma-range-unmap.patch`); should be
  upstreamed to `nutanix/libvfio-user` and the local patch dropped.
- **Windows benchmark-latency**: stays on the legacy Hetzner storagebox
  guest until a nosi Windows image is published.


cijoe/scripts/root_unlock.py
```diff
@@ -0,0 +1,163 @@
...
+def main(args, cijoe):
+    """Open root SSH over the operator account, then wait for root to be up."""
+
+    # No-op when root SSH is already available (the fork-based guest or any
+    # image that ships with root enabled). Probing root first means this
+    # step is inert wherever it is not needed, independent of whether an
+    # operator transport happens to be configured.
+    if wait_for_transport(cijoe, None, args.root_probe_timeout):
+        log.info("root SSH already available; skipping unlock")
+        return 0
```
naddinadja: By giving `transport_name=None` here, you are relying on the config file to always define the `root` transport as the first one. This is also what you've done, but it's a somewhat "fragile" assumption.

cijoe/scripts/root_unlock.py
```diff
@@ -0,0 +1,163 @@
...
+
+    transports = cijoe.config.options.get("cijoe", {}).get("transport", {})
```
naddinadja: use `cijoe.getconf("cijoe.transport", {})`

cijoe/scripts/root_unlock.py
```diff
@@ -0,0 +1,163 @@
...
+    if "linux" in name:
+        return "linux"
+    return ""
+
+
+def unlock_commands(os_name):
+    """Return the shell commands to unlock root for the given uname -s family.
+
+    Each list is executed in order under the operator transport. A failure
+    in any single command aborts the unlock and returns the error.
```
naddinadja: Since the commands are not run in this function, I think it would be better to describe it as "A failure in any single command *should* abort the unlock".

cijoe/scripts/stage_nosi_guest.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def add_args(parser: ArgumentParser):
+    pass
+
+
+def expand_home(value):
+    """Expand the cijoe '{{ local.env.HOME }}' template in a config string."""
+
+    return value.replace("{{ local.env.HOME }}", os.environ["HOME"])
```
naddinadja: I do not think this is necessary - testing on my local machine, cijoe ensures to expand this when loading the config file.

cijoe/scripts/stage_nosi_guest.py
```diff
@@ -0,0 +1,163 @@
...
+                else:
+                    out.write(buf)
+            out.truncate()
+        blob.unlink(missing_ok=True)
+
+        err, _ = cijoe.run_local(f"qemu-img convert -f raw {raw} -O qcow2 {dst}")
+        if err:
+            log.error(f"qemu-img convert failed: err({err})")
+            return err
+        raw.unlink(missing_ok=True)
```
naddinadja: Do you also want to unlink on error paths?

cijoe/scripts/xnvme_guest_start_nvme.py
```diff
@@ -255,6 +266,26 @@ def namespace(controller_id, nsid, aux={}):
...
+                f"pcie-root-port,id={root_port},chassis={slot},slot={slot}",
+            ]
+            vfio_dev = (
+                '{"driver":"vfio-user-pci","socket":{"type":"unix","path":"'
+                + serial
+                + '"},"bus":"'
+                + root_port
+                + '"}'
+            )
+            nvme += ["-device", "'" + vfio_dev + "'"]
```
naddinadja: maybe do 
```python
nvme += ["-device", f"'{vfio_dev}'"]
```

cijoe/scripts/xnvme_guest_start_nvme.py
```diff
@@ -276,6 +307,66 @@ def namespace(controller_id, nsid, aux={}):
...
+    server is launched detached so it outlives this invocation; the guest
+    connects to its socket after boot.
+    """
+    binary = cijoe.getconf("kvssd.binary", None)
+    url = cijoe.getconf("kvssd.url", None)
+    if not (binary or url):
+        return None
+
+    socket = cijoe.getconf("kvssd.socket", "/tmp/vfu_kvssd.sock")
+    log_path = str(Path(socket).with_suffix(".log"))
```
naddinadja: I don't think you need the string-cast here, it should cleanly insert the path into the format-strings without it too.

cijoe/scripts/xnvme_guest_start_nvme.py
```diff
@@ -276,6 +307,66 @@ def namespace(controller_id, nsid, aux={}):
...
+
+    need_fetch = True
+    if binary:
+        err, _ = cijoe.run_local(f"[ -x {binary} ]")
+        need_fetch = bool(err)
+    if need_fetch:
+        if not binary:
+            binary = str(Path(socket).with_name("vfu_kvssd"))
```
naddinadja: same here with unnecessary string-cast

cijoe/tests/conftest.py
```diff
@@ -441,26 +435,15 @@ def inner(*args, **kwargs):
...
-    Non-parametrized tests (e.g. test_enum) sort before the parametrized
-    batches; their relative order among themselves is preserved.
-
-    Linux only. On FreeBSD the same reordering causes fio + kqueue async
-    write/verify to fail (CRC mismatches between written and read-back data
-    at backend-grouping boundaries -- a separate test-isolation or async
-    completion-flush issue exposed by the new order). Leave freebsd at its
-    original interleaved order until the kqueue write/verify regression is
-    diagnosed; the trixie wedge motivating this hook only manifests on
-    Linux driver-swap paths anyway.
```
naddinadja: Should these changes have been in commit [`93f1b48`](https://github.com/xnvme/xnvme/commit/93f1b48aef4baa056160c234f076cd22c1d51dbcl), where you introduce the function ? :)

python/tests/conftest.py
```diff
@@ -427,26 +434,15 @@ def inner(*args, **kwargs):
...
-    Non-parametrized tests sort before the parametrized batches; their
-    relative order among themselves is preserved.
-
-    Linux only. On FreeBSD the same reordering causes fio + kqueue async
-    write/verify to fail (CRC mismatches between written and read-back data
-    at backend-grouping boundaries -- a separate test-isolation or async
-    completion-flush issue exposed by the new order). Leave freebsd at its
-    original interleaved order until the kqueue write/verify regression is
-    diagnosed; the trixie wedge motivating this hook only manifests on
-    Linux driver-swap paths anyway.
```
naddinadja: Should these changes have been in commit [`93f1b48`](https://github.com/xnvme/xnvme/commit/93f1b48aef4baa056160c234f076cd22c1d51dbcl), where you introduce the function ? :)

python/tests/conftest.py
```diff
@@ -250,15 +262,10 @@ def fabrics_setup(cijoe):
...
 def fabrics_teardown(cijoe):
-    """Teardown fabrics.
-
-    Probe first with pgrep and only signal when a nvmf_tgt is actually
-    running. pkill on its own exits non-zero when nothing matches and the
-    cijoe transport captures that as the command's error, which the
-    XnvmeDriver fixture was logging as "ERROR fabrics teardown failed with
-    error: -15" on every cdev/bdev test setup. A "pkill ... || true" would
-    still leak pkill's own stderr; the if-form is silent when there is
-    nothing to tear down.
```
naddinadja: Should these changes have been in commit [`5e3495f`](https://github.com/xnvme/xnvme/commit/5e3495ffe8af2021f7a6ef41a2006ce1fabe14a1), where you create this docstring ? :)

cijoe/tests/conftest.py
```diff
@@ -258,15 +257,10 @@ def fabrics_setup(cijoe):
...
 def fabrics_teardown(cijoe):
-    """Teardown fabrics.
-
-    Probe first with pgrep and only signal when a nvmf_tgt is actually
-    running. pkill on its own exits non-zero when nothing matches and the
-    cijoe transport captures that as the command's error, which the
-    XnvmeDriver fixture was logging as "ERROR fabrics teardown failed with
-    error: -15" on every cdev/bdev test setup. A "pkill ... || true" would
-    still leak pkill's own stderr; the if-form is silent when there is
-    nothing to tear down.
```
naddinadja: Should these changes have been in commit [`5e3495f`](https://github.com/xnvme/xnvme/commit/5e3495ffe8af2021f7a6ef41a2006ce1fabe14a1), where you create this docstring ? :)

naddinadja (changes requested):
Okay, I have a couple of comments - the last ones are just about the changes to docstrings probably being in the wrong commits :)

cijoe/scripts/root_unlock.py
safl: How else should it be done?

naddinadja: We could name it "cijoe.transport.root" in the config and then use "root" as transport name

python/tests/conftest.py
safl: Comment seems to be on an old state?

cijoe/tests/conftest.py
safl: Comment seems to be on an old state?

python/tests/conftest.py
safl: Comment seems to be on an old state?

cijoe/tests/conftest.py
safl: Comment seems to be on an old state?

cijoe/scripts/root_unlock.py
safl: Good idea, thanks!

cijoe/tests/conftest.py
naddinadja: oh woops, i thought it would take both the old and new state if I marked like this - the idea here is just that you're changing the docstrings in the very last commits, but these docstrings were added in previous commits, so maybe they should just have been the new version from the first commit they were introduced

safl: Ah, now i understand. You are right. Fixed.

python/tests/conftest.py
safl: Ah, now i understand. You are right. Fixed.

python/tests/conftest.py
safl: Ah, now i understand. You are right. Fixed.

cijoe/tests/conftest.py
safl: Ah, now i understand. You are right. Fixed.

cijoe/scripts/xnvme_guest_start_nvme.py
safl: Fixed.

cijoe/scripts/xnvme_guest_start_nvme.py
safl: ack

cijoe/scripts/xnvme_guest_start_nvme.py
safl: ack

cijoe/scripts/stage_nosi_guest.py
safl: ack

cijoe/scripts/stage_nosi_guest.py
safl: ack

cijoe/scripts/root_unlock.py
safl: ack

cijoe/scripts/root_unlock.py
safl: ack

safl:
> Okay, I have a couple of comments - the last ones are just about the changes to docstrings probably being in the wrong commits :)

Review addressed. Additionally, i realized a bunch of now unused cruft could be removed (the makefile helpers generating boot-images since the CI now just consumes the nosi images.

cijoe/scripts/root_unlock.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env python3
+"""
+Unlock root SSH on a nosi guest
+===============================
+
+nosi images ship safe-by-default: the root account is LOCKED; you log in as the
```
karlowich: Why is this the case? Seems like we can save quite a bit of code, if the root wasn't locked.

cijoe/workflows/test-debian-trixie.yaml
```diff
@@ -17,7 +17,17 @@ steps:
...
-    args: '{{ config.xnvme.repository.sync.remote_path }}/python/tests -k "not fabrics and not libvfn and not test_list"'
+    # Keep the hang-aware diagnostics from the test_key_value investigation
+    # so any future wedge surfaces with a Python traceback rather than just a
+    # silent timeout. pytest-timeout's thread method dumps every Python
+    # thread's stack right before killing the case; faulthandler_timeout=55
+    # fires 5s earlier for a second snapshot, --durations=20 surfaces real
+    # outliers so we can size the cap on data, and -v -s gives live progress
+    # plus unbuffered stderr from the C layer. The wedge that motivated
+    # these flags was resolved by the conftest backend-ordering hook and the
+    # libvfio-user range-unmap fix.
```
karlowich: Is this comment (and many others in the commit/PR) necessary? It seems kind of overkill, when all you added was a timeout.
If you look through the PR there are many very verbose/convoluted comments - I think it would be nice to trim them down or remove them.

docs/contributing/ci/index.rst
```diff
@@ -90,6 +90,75 @@ In addition to cathing issues, then the CI is also utilized for:
...
+
+* **Host containers** carry the toolchain the **GHA** job itself uses: **QEMU**
+  binaries, **CIJOE**, devbind / hugepages helpers.
+
+* **Guest disk images** are the qcow2 boot disk the in-job **QEMU** instance
+  runs. They are pulled in-job by the ``stage_nosi_guest`` **CIJOE** script
+  (invoked from ``provision-using-tgz.yaml`` and ``provision-maas-using-tgz
+  .yaml`` right before ``guest_initialize``), which talks directly to the
+  **OCI** Distribution API via stdlib ``urllib`` (no ``oras`` CLI dependency);
+  the blob digest is pinned in each **CIJOE** config's ``[guest_image].url``. The nosi guests ship with root locked and an operator
```
karlowich: This should be reflowed to fit the width of the rest of the text

cijoe/tests/conftest.py
```diff
@@ -423,3 +423,39 @@ def inner(*args, **kwargs):
...
+
+    Sort items by backend config first and test name second so the timeline
+    runs as test_foo[kernel] -> test_bar[kernel] -> ... -> test_foo[spdk]
+    -> test_bar[spdk] -> ..., minimising the number of driver-swap
+    boundaries in the run and clustering failures by backend. Tests without
+    parametrization sort before the parametrized batches with their
+    relative order preserved.
+
+    Linux only. FreeBSD's kqueue async write/verify fails when reordered
+    this way, so leave it at the original collection order.
```
karlowich, line 443: Let's make an issue for this, I think that would be good to fix.

karlowich (changes requested):
Looks interesting, can't quite grasp what exactly is going on here tbh.
I don't like this commit: "ci: migrate docgen + verify to upstream QEMU + vfu_kvssd on nosi" It is huge, hard to review, and by the looks of the title it could've been 3 commits instead.

cijoe/scripts/root_unlock.py
safl: Locked root is the default for OS images. Shipping images with root able to login with plaintext as root:root over SSH is considered a hazard. xNVMe's previous in house image build broke that convention to make CI simpler (root:root everywhere), at the cost of owning image build infrastructure that ended up unmaintained and undocumented (the bootimg workflows removed in this PR plus the hand built qcow2s on Hetzner storagebox the old `system_imaging.toml` pointed at).

Moving to general purpose nosi images restores the convention. The unlock script is the small adapter that keeps the existing CI workflows running unchanged on top of it. A 200 line adapter is about as minimal as it can get.

The nicer long term solution would be cijoe supporting privilege elevation, or xNVMe's CI using sudo everywhere instead of relying on root. That is well out of scope here, and I believe you already find this PR too large.

cijoe/workflows/test-debian-trixie.yaml
safl: This PR took a lot of iterations to weed out failures because the cascading errors was brutal. Whhen something broke it was very hard to tell whether the culprit was the new qemu, changes in qemu's built-in NVMe emulation, changes on the image side, the container, or the vfio user emulation. So much churn. The comments made a lot of sense while I was in the middle of that, as I was capturing what I had just figured out so it would not have to be re figured out in the next ci run.

Now that you point at it, though, I agree it reads as very verbose and seemingly out of context. Will take a pass trimming and removing.

cijoe/tests/conftest.py
safl: ack https://github.com/xnvme/xnvme/issues/685

docs/contributing/ci/index.rst
safl: ack

safl:
> Looks interesting, can't quite grasp what exactly is going on here tbh. I don't like this commit: "ci: migrate docgen + verify to upstream QEMU + vfu_kvssd on nosi" It is huge, hard to review, and by the looks of the title it could've been 3 commits instead.

Regarding the size of the commit, i can split it into a series of preparation-commits, adding scripts etc. which is not being used, just so they are introduced with a commit-message motivating them. That might help with the digestion...

edit: That actually did clean out several changes that were there during the troubleshooting phase which was no longer needed.

naddinadja (approved):
I don't have anymore comments, and this looks good. :) I see that there are still quite a few long in-code comments, but as this is all CI-related, and not something we look at *that* often, I think the remaining comments may be good to have, so that we do not forget what is going on when we look at these parts of the code again. :)

karlowich (approved):
Great!
Splitting the large commits into smaller ones made the reviewing process much nicer, thanks!

cijoe/scripts/root_unlock.py
karlowich: All good

cijoe/workflows/test-debian-trixie.yaml
karlowich: Good

PR #683, opened 2026-06-01
be/cref: API cleanup and deferred teardown to fix RTE churn during enumeration

karlowich:
**Single-lookup conflict detection** (`refactor(be/cref): change lookup
  be_name from filter to out-parameter`)
  `xnvme_be_cref_lookup()` previously took `be_name` as a filter, requiring
  callers to call it twice to distinguish a same-name hit from a
  conflict with a different backend. Change `be_name` to a `const char**`
  out-parameter so a single call covers both cases.

  **Remove `xnvme_be_cref_ref()`** (`cleanup(be/cref)`)
  `xnvme_be_cref_ref()` was a combined lookup-or-insert helper used only by
  the old SPDK enumerate path (removed in #680). Dead code.

  **Deferred teardown** (`refactor + fix`)
  During enumeration, each `ctrlr_dev` close triggers the `ctrlr` destructor
  immediately when refcount hits 0. For SPDK and uPCIe this tears down the RTE
  between controllers, forcing re-initialisation for every subsequent
  controller in the scan.

  Introduce `xnvme_be_cref_defer()`: callers mark an entry as deferred
  before releasing it. `xnvme_be_cref_put()` then leaves `refcount==0` entries
  alive rather than destroying them. `xnvme_be_cref_cleanup()` destroys all
  pending entries in one pass and is called after the full scan completes.

  For devices that stay open across an enumeration boundary, `cleanup()`
  clears the deferred flag so they are destroyed normally on the caller's
  eventual close.

  Fix `xnvme_be_cref_insert()` to scan for a free slot using `ctrlr == NULL` rather than
  `refcount == 0`, with deferred entries the two are no longer equivalent.

  **Rename** (`rename(be/cref)`)
  `lookup()` -> `get()`, `deref()` -> `put()`. get/put is a natural acquire/release pair
  that makes the refcounting semantics clear.

### DIFF
--- a/include/xnvme_be_cref.h
+++ b/include/xnvme_be_cref.h
@@ -17,15 +17,28 @@ enum xnvme_be_cref_flags {
 };
 
 /**
- * Lookup a controller by @uri and @be_name, and increment its refcount
+ * A controller reference entry
+ *
+ * @ctrlr holds the controller handle and @be_name holds the backend name the
+ * entry was inserted with.
+ */
+struct xnvme_be_cref_entry {
+	void *ctrlr;
+	xnvme_be_cref_destructor_fn destructor;
+	const char *be_name;
+	int refcount;
+	char uri[XNVME_IDENT_URI_LEN + 1];
+};
+
+/**
+ * Lookup a controller by @uri and increment its refcount
  *
  * @param uri Device URI to match against
- * @param be_name Backend name to match (pointer equality), or NULL for any
  *
- * @return On success, the controller handle. On error, NULL.
+ * @return On a hit, a const pointer to the matched entry. On a miss, NULL.
  */
-void *
-xnvme_be_cref_lookup(const char *uri, const char *be_name);
+const struct xnvme_be_cref_entry *
+xnvme_be_cref_lookup(const char *uri);
 
 /**
  * Insert a new controller entry with refcount=1
--- a/include/xnvme_be_cref.h
+++ b/include/xnvme_be_cref.h
@@ -11,11 +11,6 @@
 
 typedef int (*xnvme_be_cref_destructor_fn)(void *ctrlr);
 
-enum xnvme_be_cref_flags {
-	XNVME_BE_CREF_NONE              = 0x0,
-	XNVME_BE_CREF_DESTROY_IMMEDIATE = 0x1,
-};
-
 /**
  * A controller reference entry
  *
@@ -47,7 +42,7 @@ xnvme_be_cref_lookup(const char *uri);
  * already present, e.g. by calling xnvme_be_cref_lookup() first.
  *
  * @param uri Device URI to associate with the controller
- * @param be_name Backend name pointer, used for filtering in xnvme_be_cref_cleanup()
+ * @param be_name Backend name pointer, stored for later matching
  * @param ctrlr Controller handle to store, must be non-NULL
  * @param destructor Callback invoked to destroy the controller when refcount reaches zero
  *
@@ -58,49 +53,16 @@ xnvme_be_cref_insert(const char *uri, const char *be_name, void *ctrlr,
 		     xnvme_be_cref_destructor_fn destructor);
 
 /**
- * Combined lookup-or-insert
- *
- * If an entry with matching @uri exists, increment its refcount and return the
- * stored handle. Otherwise, if @ctrlr is non-NULL, insert it with refcount=1
- * and associate it with @be_name and @destructor.
- *
- * @param uri Device URI to match or insert
- * @param be_name Backend name pointer, stored for filtering in xnvme_be_cref_cleanup()
- * @param ctrlr Controller handle to insert if no match is found, may be NULL for lookup-only
- * @param destructor Callback invoked to destroy the controller when refcount reaches zero
- *
- * @return On success, the controller handle. On error, NULL.
- */
-void *
-xnvme_be_cref_ref(const char *uri, const char *be_name, void *ctrlr,
-		  xnvme_be_cref_destructor_fn destructor);
-
-/**
  * Decrement the refcount for the given controller
  *
- * If the refcount reaches zero and XNVME_BE_CREF_DESTROY_IMMEDIATE is set,
- * call the stored destructor and remove the entry. Otherwise the entry remains
- * with refcount=0 until xnvme_be_cref_cleanup() is called.
+ * When the refcount reaches zero, call the stored destructor and remove the
+ * entry.
  *
  * @param ctrlr Controller handle to dereference, must be non-NULL
- * @param flags Bitmask of enum xnvme_be_cref_flags
  *
  * @return 0 on success, negative errno on error.
  */
 int
-xnvme_be_cref_deref(void *ctrlr, enum xnvme_be_cref_flags flags);
-
-/**
- * Destroy all entries with zero refcount matching @be_name
- *
- * Iterates all entries, and for those matching @be_name with refcount=0,
- * calls the stored destructor and removes the entry.
- *
- * @param be_name Backend name pointer to match (compared by pointer equality)
- *
- * @return 0 on success, negative errno of the last failed destructor on error.
- */
-int
-xnvme_be_cref_cleanup(const char *be_name);
+xnvme_be_cref_deref(void *ctrlr);
 
 #endif /* __INTERNAL_XNVME_BE_CREF_H */
--- a/include/xnvme_be_cref.h
+++ b/include/xnvme_be_cref.h
@@ -33,13 +33,13 @@ struct xnvme_be_cref_entry {
  * @return On a hit, a const pointer to the matched entry. On a miss, NULL.
  */
 const struct xnvme_be_cref_entry *
-xnvme_be_cref_lookup(const char *uri);
+xnvme_be_cref_get(const char *uri);
 
 /**
  * Insert a new controller entry with refcount=1
  *
  * Does not check for duplicate URIs. The caller must ensure the URI is not
- * already present, e.g. by calling xnvme_be_cref_lookup() first.
+ * already present, e.g. by calling xnvme_be_cref_get() first.
  *
  * @param uri Device URI to associate with the controller
  * @param be_name Backend name pointer, stored for later matching
@@ -63,6 +63,6 @@ xnvme_be_cref_insert(const char *uri, const char *be_name, void *ctrlr,
  * @return 0 on success, negative errno on error.
  */
 int
-xnvme_be_cref_deref(void *ctrlr);
+xnvme_be_cref_put(void *ctrlr);
 
 #endif /* __INTERNAL_XNVME_BE_CREF_H */
--- a/lib/xnvme_be_cref.c
+++ b/lib/xnvme_be_cref.c
@@ -6,14 +6,6 @@
 #include <errno.h>
 #include <xnvme_be_cref.h>
 
-struct xnvme_be_cref_entry {
-	void *ctrlr;
-	xnvme_be_cref_destructor_fn destructor;
-	const char *be_name;
-	int refcount;
-	char uri[XNVME_IDENT_URI_LEN + 1];
-};
-
 static struct xnvme_be_cref_entry g_cref_table[XNVME_BE_CREF_MAX_ENTRIES];
 
 static int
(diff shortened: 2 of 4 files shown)


include/xnvme_be_cref.h
```diff
@@ -64,7 +64,7 @@ xnvme_be_cref_defer(void *ctrlr);
  * @return 0 on success, negative errno on error.
  */
 int
-xnvme_be_cref_deref(void *ctrlr);
+xnvme_be_cref_put(void *ctrlr);
```
naddinadja, line 66: I don't like the name "put". It's good that it's changed away from deref, as it's too close to defer, but usually "put" means updating something ("putting" a value in a place), not releasing it ("putting" it down). I know "release" is longer, but it might be a better alternative.

naddinadja (commented):
One comment for this

include/xnvme_be_cref.h
```diff
@@ -17,15 +17,15 @@ enum xnvme_be_cref_flags {
...
  *
  * @param uri Device URI to match against
- * @param be_name Backend name to match (pointer equality), or NULL for any
+ * @param be_name If non-NULL, receives the backend name of the matched entry
  *
  * @return On success, the controller handle. On error, NULL.
  */
 void *
-xnvme_be_cref_lookup(const char *uri, const char *be_name);
+xnvme_be_cref_lookup(const char *uri, const char **be_name);
```
safl: The ``resolve-conflict-by-name`` strategy had issues. To simplify ref-counting let's step away from overloading what ``cref_lookup()`` does, e.g. let's have it pure and simple lookup of the controller-reference, irrespective of the name of the backend. When we need to determine compatibility, something other than the name should be used... possibly information carried in the controller-struct. This would also assist with the current hidden assumptions, e.g. recent commits/changes re-arranged ordering of members, this could be explicitly modeled by some common ``xnvme_cref_ctrlr`` / ``xnvme_ctrlr`` // ``xnvme_be_cref_controller`` or some other fitting name. The lookup can then be typed, enabling the caller to check various properties on controller-state matching a given uri.

Signature should probably go more towards a direction of:

```c
struct xnvme_be_cref_controller {
// fields assumed to be common among backends with explicit controller-management
};

struct xnvme_be_cref_controller *xnvme_be_cref_lookup(const char *uri);
```

While touching on this. We might even be able to drop ``xnvme_be_cref_ref()``?

safl (commented):
Great, this is the juicy part. Comment and questions inline.

include/xnvme_be_cref.h
safl: I see that ``xnvme_be_cref_ref`` is removed in this PR, that is great. I just didn't make it any further than the first commit before i found something worth discussing.

karlowich: Having a struct is definitely nicer than the current convention of `((void **)be->state)[0] = ctrlr;`

`xnvme_be_cref_ref()` is dropped as part of this PR.

include/xnvme_be_cref.h
karlowich: So we go with `get()/release()`?

safl: Consider conventional naming pairs here. E.g. ``get()/put()``, ``ref()/unref()``, ``increment()/decrement()``. Considering that then ``get()/release()`` is hard to follow.

naddinadja: I personally don't like get/put, but I see that the kernel does the same for reference counting, so maybe that's the best option.

include/xnvme_be_cref.h
karlowich: Okay, I tried this out. It is a huge refactor and honestly the gain is so tiny that I don't think it is worth it.
Instead, I will create a struct local to the cref logic that has only what is needed there.

include/xnvme_be_cref.h
karlowich: We will keep get/put

include/xnvme_be_cref.h
```diff
@@ -17,15 +17,27 @@ enum xnvme_be_cref_flags {
...
  *
- * @return On success, the controller handle. On error, NULL.
+ * @return On a hit, a reference whose @ctrlr is the controller handle and whose
+ * @be_name is the backend name the entry was inserted with. On a miss, a zeroed
+ * reference whose @ctrlr is NULL.
  */
-void *
-xnvme_be_cref_lookup(const char *uri, const char *be_name);
+struct xnvme_be_cref
+xnvme_be_cref_lookup(const char *uri);
```
safl: Agreed on dropping ``be_name`` from the signature, great.

As an alternative to the new tuple type, the lookup could return the entry pointer directly:

```c
struct xnvme_be_cref_entry *xnvme_be_cref_lookup(const char *uri);
```

The signature remains stable as fields are added to xnvme_be_cref_entry, and callers can read whichever fields they need.

include/xnvme_be_cref.h
```diff
@@ -56,24 +56,6 @@ int
 xnvme_be_cref_insert(const char *uri, const char *be_name, void *ctrlr,
 		     xnvme_be_cref_destructor_fn destructor);
 
-/**
```
safl, line 47: **AWESOME**

safl (commented):
See inline.

include/xnvme_be_cref.h
karlowich: Yes, I almost made this implementation. However, I was worried that someone at some point might decide to manually change the refcount or any other sensitive information. This can lead to some nasty bugs, which could be avoided if the `cref_entry` is only modifiable through the API.

safl: Well, then:

```c
const struct xnvme_be_cref_entry *xnvme_be_cref_lookup(const char *uri);
```

No accidental writes to refcount, and the pointer-or-NULL contract stays.

karlowich: Done

safl (commented):
Regarding:

- refactor(be/cref): introduce xnvme_be_cref_defer for deferred teardown
- fix(platform): use cref defer to prevent RTE teardown during enumeration

Is the latter truly a fix, or more like an optimization? And is doing the ``cref_refer``/``cref_defer`` (which btw. is too easy to overlook as a typo...), i guess this is refactored later as get()/put(), so nevermind that... but is this really the right solution? As i recall then the SPDK backend guards the teardown of the DPDK/env. so this is more about doing 1 RTE init/term when enumerating a system with N controllers, instead of doing N RTE init/term when enumerating a system with N controllers, right? Or have you observed errors due to this? And currently, this addresses behavior of uPCIe, not SPDK an libvfn, right?

If my understanding is correct, then i am considering that perhaps backend/controller RTEs should self-guard? Or provide knobs for "pinning" the RTE, such that we don't model the cref-api around some other underlying behavior. I guess doing so would allow us to completely wipe this ``defer()/cleanup()`` life-time managed done here.

karlowich:
> Is the latter truly a fix

Maybe "fix" isn't the right word. I don't see any errors. It is an optimization.
SPDK doesn't guard teardown, since it is now managed by the general purpose cref infra.
As such, the patch applies to all userspace backends.

> this is more about doing 1 RTE init/term when enumerating a system with N controllers, instead of doing N RTE init/term when enumerating a system with N controllers, right?

Exactly.

> I am considering that perhaps backend/controller RTEs should self-guard?

I don't think that is possible. The backends doesn't have the information necessary for this.

safl:
> > Is the latter truly a fix
> 
> Maybe "fix" isn't the right word. I don't see any errors. It is an optimization. SPDK doesn't guard teardown, since it is now managed by the general purpose cref infra. As such, the patch applies to all userspace backends.



> 
> > this is more about doing 1 RTE init/term when enumerating a system with N controllers, instead of doing N RTE init/term when enumerating a system with N controllers, right?
> 
> Exactly.
> 
> > I am considering that perhaps backend/controller RTEs should self-guard?
> 
> I don't think that is possible. The backends doesn't have the information necessary for this.



> > Is the latter truly a fix
> 
> Maybe "fix" isn't the right word. I don't see any errors. It is an optimization. SPDK doesn't guard teardown, since it is now managed by the general purpose cref infra. As such, the patch applies to all userspace backends.
> 
> > this is more about doing 1 RTE init/term when enumerating a system with N controllers, instead of doing N RTE init/term when enumerating a system with N controllers, right?
> 
> Exactly.
> 
> > I am considering that perhaps backend/controller RTEs should self-guard?
> 
> I don't think that is possible. The backends doesn't have the information necessary for this.

Looking at the userspace backends that define `ctrlr_init` / `ctrlr_term`:

- `lib/xnvme_be_spdk_dev.c`: SPDK never tears down its env, only the controller. `spdk_env_init` is one-shot, guarded by a static flag, with no symmetric `_fini` anywhere.
- `lib/xnvme_be_vfio_dev.c`: libvfn has no env-level init/term at all; pure per-controller lifecycle.
- `lib/xnvme_be_macos_driverkit_dev.c`: same shape as libvfn; `ctrlr_term` just closes the IOService connections and frees the controller, no global state, no env teardown.
- `lib/upcie/xnvme_be_upcie_dev.c`: the only base backend with env teardown the deferred infra is protecting, via its own `g_ctrlr_count` and `_rte_term()`.
- `lib/upcie/xnvme_be_upcie_cuda_dev.c`: builds on upcie and adds a *second* env layer of its own, an `_Atomic int g_cuda_ctrlr_count` driving `_cuda_rte_term()`, which tears down both the CUDA-DMA heap (`cudamem_heap_term`) and the CUDA context (`cuCtxDestroy`). Its `ctrlr_term` calls the base upcie `ctrlr_term` first (decrementing `g_ctrlr_count`) and then decrements `g_cuda_ctrlr_count`, so a single controller close can transitively tear down two layers of env state when both counters hit zero.

So the deferred-teardown lifecycle on cref entries is solving a backend-runtime lifetime problem at the wrong layer. Controller refcounting belongs in cref; environment refcounting is a separate concern that sits below the controller, inside each backend's own runtime. I'd drop `xnvme_be_cref_defer` / `xnvme_be_cref_cleanup` so cref stays a pure refcount module with the conventional contract, and handle env lifetime in one of the shapes below.

One approach would be to add explicit env lifecycle hooks at the backend layer. Backends declare `env_init` / `env_term` (or `env_pin` / `env_unpin`) alongside `ctrlr_init` / `ctrlr_term`. The platform layer takes one env reference around `xnvme_platform_enumerate` and drops it at the end. The backend's `g_ctrlr_count` (or equivalent) still decides whether that was the last user. Same kref idiom that already governs controllers, applied one level up. Clean and explicit, at the cost of new ops in the backend interface. Having said this, if my memory serves me correctly, the SPDK/DPDK env does not like, or possibly does not support, repeated bring-up/teardown, which leaves us back at providing an entire backend-interface extension for a single backend... so probably not ideal. The upcie-cuda case also makes this fiddlier: a single env hook would need to compose across the two stacked counters, or both layers need their own hooks.

Another approach would be to drop upcie's `_rte_term` and let the env stay up like SPDK does. Make `_rte_init` one-shot (set the flag, never clear it), retire `g_ctrlr_count`. The hugepage heap is held for the process lifetime, same as SPDK's env. For upcie-cuda the same has to be done a second time (neutralise `_cuda_rte_term` and retire `g_cuda_ctrlr_count`), but it's the same pattern repeated. No new API surface, the cref deferred lifecycle goes away entirely, and the userspace backends end up with consistent env semantics. The cost is that a process that briefly enumerates then idles still holds the heap(s), which is what every other RTE-using tool in this space does anyway and which DPDK is famously bad at giving back to the OS in the first place.

A third approach would be to take an extra cref reference per controller for the duration of enumeration. When the platform layer opens a controller during `xnvme_platform_enumerate`, it bumps the refcount once via the function named something like `_get()` and records the pointer in a local list. Enumeration's own `dev_open`/`dev_close` pairs balance as usual. After enumeration completes, the platform walks its list and drops the extra references; each `_put` runs the destructor normally as it hits zero. For upcie, that means `g_ctrlr_count` only reaches zero (and `_rte_term()` only runs) once enumeration is finished. For upcie-cuda this falls out for free: the pin sits at the cref level above both env counters, so delaying the destructor naturally keeps both layers alive without the platform code needing to know they exist. No new backend API, no env lifecycle on cref entries, no perpetual heap hold; the only cost is that the platform layer carries a small list of pinned controllers for the duration of one enumeration. This is the kref idiom for "extend an object's lifetime across a scope".

I lean mostly towards the latter.

karlowich:
@safl I think the new shape of the PR matches what you were going for

safl (approved):
Great!

naddinadja (approved):
lgtm :)

PR #684, opened 2026-06-02
fix(tools/xnvmeperf): advance slba cursor only on a queued IO

fmma:
submit_io() advanced the sequential LBA even when the submit failed. Since backends return -EBUSY transiently at queue saturation (e.g. upcie) and the run loop retries, the position skipped ahead on every failed attempt, silently turning --iopattern read into a random read.

Split slba selection into a side-effect-free peek_slba() and an advance_slba() that runs only on a successful submit.

### DIFF
--- a/tools/xnvmeperf/xnvmeperf.c
+++ b/tools/xnvmeperf/xnvmeperf.c
@@ -23,7 +23,8 @@ struct xnvmeperf_job {
 	unsigned int seed;
 	uint64_t io_completed;
 	uint64_t io_failed;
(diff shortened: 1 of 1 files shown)


naddinadja (approved):
This is great! Thank you :)

safl (approved):
Excellent!

There is currently a hicup in the CI, work ongoing here: https://github.com/xnvme/xnvme/pull/682
Once the dust settles on that one, then we can start the maas-testing on this one.

PR #686, opened 2026-06-08
Multi-process support for xNVMe

naddinadja:
Here, I am adding multi-process support, which lets multiple processes share a single userspace NVMe device. The primary process initializes the hardware and preallocates a pool of queue pairs, and secondary processes attach and claim pairs from that pool via shared memory, without issuing any admin commands themselves.
  
The upcie backend gets full support for this through POSIX shm and hugepage import. SPDK gains the ability to enter multi-process mode through proc_role alone, without needing core_mask. Kernel backends supported this already, and the vfio backend does not support multi-process mode (by not allowing secondary processes to connect to xNVMe devices).

### DIFF
--- a/docs/tutorial/multi_process/index.rst
+++ b/docs/tutorial/multi_process/index.rst
@@ -1,0 +1,177 @@
+.. _sec-tutorials-multi-process:
+
+================
+ Multi-Process
+================
+
+**xNVMe** supports multiple processes accessing the same userspace NVMe
+device simultaneously. The primary process allocates all hardware queue pairs
+upfront; each process then claims one or more pairs exclusively for I/O.
+No queue pair is used by more than one process at a time.
+
+Roles
+=====
+
+Each process that opens a device declares one of three roles via
+``xnvme_opts.proc_role``:
+
+``XNVME_PROC_SINGLE``
+  Default. Single-process mode with no multi-process coordination. Queue pairs
+  are allocated on demand; no shared memory is created. Use this when only one
+  process will ever access the device.
+
+``XNVME_PROC_PRIMARY``
+  Initializes the hardware, pre-allocates the shared queue pool, and publishes
+  shared state. The primary must open the device before any secondary process
+  attempts to attach.
+
+``XNVME_PROC_SECONDARY``
+  Attaches to an already-initialized controller without touching the hardware.
+  Returns an error if no primary is running for the given device.
+
+Backends
+========
+
+Multi-process works across all backends. The difference is in how coordination
+happens:
+
+**Kernel-based backends** (e.g., ``linux``, ``io_uring``)
+  The OS handles concurrent device access natively. Multiple processes can
+  open the same device independently and ``proc_role`` has no effect.
+
+**vfio** (``libvfn``)
+  Single-process only. The VFIO group-based API allows a group to be in only
+  one IOMMU container at a time, so a second process cannot share the
+  primary's IOMMU domain. ``proc_role`` has no effect for this backend.
+
+**upcie** (``uio_pci_generic`` driver only)
+  Full multi-process support. The uPCIe backend programs the NVMe controller
+  directly and coordinates through POSIX shared memory. Setting ``proc_role``
+  tells the library which process should initialize the hardware and which
+  should attach to an already-running controller. Without this coordination,
+  a secondary process would reset the device and destroy in-flight I/O.
+
+  Devices bound to ``vfio-pci`` do not support secondary attachment;
+  ``XNVME_PROC_SECONDARY`` returns ``ENOTSUP`` for those devices.
+
+Usage
+=====
+
+The queue API is identical regardless of process role. Each process calls
+``xnvme_queue_init()`` to claim a queue pair and ``xnvme_queue_term()`` to
+release it. Queue pairs are drawn from a pool pre-allocated by the primary at
+device-open time.
+
+Primary process
+---------------
+
+.. code-block:: c
+
+   struct xnvme_opts opts = xnvme_opts_default();
+   opts.be = "upcie";
+   opts.proc_role = XNVME_PROC_PRIMARY;
+
+   struct xnvme_dev *dev = xnvme_dev_open("0000:01:00.0", &opts);
+   if (!dev) {
+       perror("xnvme_dev_open");
+       return 1;
+   }
+
+   struct xnvme_queue *queue;
+   if (xnvme_queue_init(dev, 256, 0, &queue)) {
+       perror("xnvme_queue_init");
+       return 1;
+   }
+
+   /* submit and poll I/O as normal */
+
+   xnvme_queue_term(queue);
+   xnvme_dev_close(dev);
+
+Secondary process
+-----------------
+
+.. code-block:: c
+
+   struct xnvme_opts opts = xnvme_opts_default();
+   opts.be = "upcie";
+   opts.proc_role = XNVME_PROC_SECONDARY;
+
+   struct xnvme_dev *dev = xnvme_dev_open("0000:01:00.0", &opts);
+   if (!dev) {
+       perror("xnvme_dev_open");
+       return 1;
+   }
+
+   struct xnvme_queue *queue;
+   if (xnvme_queue_init(dev, 256, 0, &queue)) {
+       perror("xnvme_queue_init");
+       return 1;
+   }
+
+   /* submit and poll I/O as normal */
+
+   xnvme_queue_term(queue);
+   xnvme_dev_close(dev);
+
+Note that the only difference between the two snippets is ``proc_role``. Once
+the device is open, all queue operations work identically.
+
+Each process may open the same device multiple times (for example, once per
+namespace) and init as many queues as needed, subject to the pool capacity
+described below.
+
+Queue pool
+==========
+
+The primary pre-allocates a fixed pool of NVMe I/O queue pairs at device-open
+time. Secondary processes claim pairs from this pool atomically; no admin
+commands are issued after primary initialization. The pool is released when
+the primary calls ``xnvme_dev_close()``.
+
+The pool holds at most 63 queue pairs per device. One of those pairs is
+reserved for the synchronous command path used internally by each process, so
+in practice each process that calls ``xnvme_dev_open()`` consumes one slot
+before any ``xnvme_queue_init()`` calls. This leaves at most 62 slots
+distributed across all attached processes.
+
+Device geometry
+===============
+
+Secondary processes can call ``xnvme_dev_get_geo()``, ``xnvme_dev_get_ctrlr()``,
+and related functions. The primary publishes its Identify data into shared
+memory the first time any of those functions is called, and secondaries read
+from shared memory instead of issuing admin commands.
+
+This means the primary must trigger at least one identify-dependent call (e.g.
+``xnvme_dev_get_geo()``) **before** any secondary calls those same functions.
+In practice this happens naturally: the primary warms up the device and starts
+I/O before launching secondary processes.
+
+Only the namespace that the primary opened first is published. Secondaries must
+open the same namespace ID as the primary. Attempting to derive geometry for a
+different namespace from a secondary returns an error.
+
+FS-interface I/O (byte-addressed reads and writes via ``XNVME_SPEC_FS_OPC_*``
+opcodes) works correctly on secondaries once geometry is available.
+
+Constraints
+===========
+
+* The primary must open the device before any secondary attempts to attach.
+  A secondary that finds no primary returns ``ENOENT``.
+
+* Secondaries must open the same namespace ID as the primary used for its
+  first identify call.
+
+* Each process is responsible for calling ``xnvme_queue_term()`` for every
+  queue it has initialized and ``xnvme_dev_close()`` for every device it has
+  opened, in order to return queue pairs to the pool.
+
+* Recovery from a crashed primary is the caller's responsibility. If the
+  primary exits without calling ``xnvme_dev_close()``, the hardware queue
+  state is lost and attached secondaries are left in an undefined condition.
+  The safest recovery path is to restart all processes.
+
+* Buffers allocated with ``xnvme_buf_alloc()`` are per-process and must not
+  be passed between processes.
--- a/lib/upcie/xnvme_be_upcie_dev.c
+++ b/lib/upcie/xnvme_be_upcie_dev.c
@@ -273,79 +273,25 @@ _upcie_release_qpair(struct xnvme_be_upcie_ctrlr *ctrlr, struct nvme_qpair *qp)
 	atomic_fetch_or_explicit(&ctrlr->mproc->qidmap, 1ULL << qp->qid, memory_order_release);
 }
 
-static int
-_upcie_prealloc_queues(struct xnvme_be_upcie_ctrlr *ctrlr, int nqueues, int depth)
-{
-	struct xnvme_be_upcie_mproc *mp = ctrlr->mproc;
-	struct hostmem_heap *heap = &g_upcie_rte.heap;
-	int err;
-
-	ctrlr->preallocated = calloc(nqueues + 1, sizeof(*ctrlr->preallocated));
-	if (!ctrlr->preallocated) {
-		return -ENOMEM;
-	}
-
-	for (int qid = 1; qid <= nqueues; qid++) {
-		struct nvme_qpair *qp = &ctrlr->preallocated[qid];
-
-		err = nvme_controller_create_io_qpair(ctrlr->ctrl, qp, (uint16_t)depth);
-		if (err) {
-			XNVME_DEBUG("FAILED: nvme_controller_create_io_qpair(qid=%d); err(%d)",
-				    qid, err);
-			for (int j = 1; j < qid; j++) {
-				struct nvme_qpair *pqp = &ctrlr->preallocated[j];
-
-				_upcie_send_delete_iosq(ctrlr->ctrl, (uint16_t)j);
-				_upcie_send_delete_iocq(ctrlr->ctrl, (uint16_t)j);
-				hostmem_dma_free(heap, pqp->sq);
-				hostmem_dma_free(heap, pqp->cq);
-			}
-			free(ctrlr->preallocated);
-			ctrlr->preallocated = NULL;
-			return err;
-		}
-
-		/* Pre-allocated queues share ring memory via shm but don't need PRP pools —
-		 * each process allocates its own when claiming via _upcie_claim_qpair.
-		 * Without this, 63 queues × 4 MB/pool = 252 MB exhausts the 256 MB DMA heap. */
-		nvme_request_pool_term_prps(qp->rpool, heap);
-		free(qp->rpool);
-		qp->rpool = NULL;
-
-		mp->queues[qid].qid = (uint32_t)qid;
-		mp->queues[qid].depth = (uint16_t)depth;
-		mp->queues[qid].sq_offset =
-			(uint64_t)((uint8_t *)qp->sq - (uint8_t *)heap->memory.virt);
-		mp->queues[qid].cq_offset =
-			(uint64_t)((uint8_t *)qp->cq - (uint8_t *)heap->memory.virt);
-		mp->queues[qid].sq_tail = 0;
-		mp->queues[qid].cq_head = 0;
-		mp->queues[qid].cq_phase = 1;
-	}
-
-	return 0;
-}
-
 static void
 _upcie_free_preallocated_queues(struct xnvme_be_upcie_ctrlr *ctrlr)
 {
-	struct xnvme_be_upcie_mproc *mproc = ctrlr->mproc;
-	int nqueues = mproc->nqueues;
-
-	for (int qid = 1; qid <= nqueues; qid++) {
+	for (int qid = 1; qid < XNVME_BE_UPCIE_NQUEUES_MAX; qid++) {
 		struct nvme_qpair *qp = &ctrlr->preallocated[qid];
+
+		if (!qp->sq) {
+			continue;
+		}
 
 		_upcie_send_delete_iosq(ctrlr->ctrl, (uint16_t)qid);
 		_upcie_send_delete_iocq(ctrlr->ctrl, (uint16_t)qid);
 
-		if (qp->sq) {
-			if (qp->rpool) {
-				nvme_request_pool_term_prps(qp->rpool, qp->heap);
-				free(qp->rpool);
-			}
-			hostmem_dma_free(qp->heap, qp->sq);
-			hostmem_dma_free(qp->heap, qp->cq);
+		if (qp->rpool) {
+			nvme_request_pool_term_prps(qp->rpool, qp->heap);
+			free(qp->rpool);
 		}
+		hostmem_dma_free(qp->heap, qp->sq);
+		hostmem_dma_free(qp->heap, qp->cq);
 	}
 	free(ctrlr->preallocated);
 	ctrlr->preallocated = NULL;
@@ -358,10 +304,7 @@ _upcie_mproc_primary_init(struct xnvme_dev *dev, struct xnvme_be_upcie_ctrlr *ct
 	struct hostmem_heap *heap = &g_upcie_rte.heap;
 	char shm_name[64];
 	size_t shm_size = sizeof(*mproc);
-	int shm_fd, nqueues, depth, qpid, err;
-
-	nqueues = XNVME_BE_UPCIE_NQUEUES_MAX - 1;
-	depth = XNVME_BE_UPCIE_QUEUE_DEPTH;
+	int shm_fd, err;
 
 	_upcie_shm_name(dev->ident.uri, shm_name, sizeof(shm_name));
 	snprintf(ctrlr->shm_name, sizeof(ctrlr->shm_name), "%s", shm_name);
@@ -389,7 +332,7 @@ _upcie_mproc_primary_init(struct xnvme_dev *dev, struct xnvme_be_upcie_ctrlr *ct
 	}
 
 	memset(mproc, 0, shm_size);
-	mproc->nqueues = nqueues;
+	mproc->nqueues = 0;
 
 	snprintf(mproc->hugepage_path, sizeof(mproc->hugepage_path), "%s", heap->memory.path);
 	mproc->hugepage_size = heap->memory.size;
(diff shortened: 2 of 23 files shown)


lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -38,12 +41,6 @@ xnvme_be_upcie_get_driver_name(const char *bdf, char *driver_name, size_t driver
 	return 0;
 }
 
-/**
- * Terminate the uPCIe runtime-environment
- *
- * The state is globally accessible via g_upcie_rte, this function terminates it, unless it is
- * not initialized, then it exits early.
- */
```
karlowich, line 46: No need to remove these docstrings (`_rte_term()`, `_rte_init()`, `_pci_enable_bus_master()`).

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -271,6 +780,63 @@ xnvme_be_upcie_dev_open(struct xnvme_dev *dev)
...
+	ectx.cb_args = cb_args;
+
+	err = pci_scan(_instantiate, &ectx);
+	if (err) {
+		perror("pci_scan()");
+	}
+
+	return 0;
+}
+
```
karlowich: This code was deleted on `main`. It shouldn't go back in, maybe you're missing a rebase?

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Queue pool
+==========
+
+The primary pre-allocates a fixed pool of NVMe I/O queue pairs at device-open
+time. Secondary processes claim pairs from this pool atomically; no admin
+commands are issued after primary initialization. The pool is released when
+the primary calls ``xnvme_dev_close()``.
+
+The pool holds at most 63 queue pairs per device. One of those pairs is
```
karlowich, line 132: Are these numbers correct? The code reads: `queues[XNVME_BE_UPCIE_NQUEUES_MAX + 1];`

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,181 @@
...
+opcodes) works correctly on secondaries once geometry is available.
+
+Constraints
+===========
+
+* The primary must open the device before any secondary attempts to attach.
+  A secondary that finds no primary returns ``ENOENT``.
+
+* The primary must call an identify-dependent function (e.g.
+  ``xnvme_dev_get_geo()``) before secondaries call such functions, so that
```
karlowich: Since this is a hard requirement, perhaps it should always be called from xNVMe's side when doing multiprocess ?

safl (commented):
A couple of quick ones:

- **Why pre-allocate qpairs?** It's a behavioral change I'd prefer not to introduce. We already allocate one qpair per controller to support the sync-interface on the user-space backends, and even that single allocation has been a source of grief on its own. Users do not expect the library to consume/allocate scarce resources like qpairs unsolicited.

- **Tests:** the existing tests in this area should be sensitive to this change. Are they still passing, or have they been masked out?

- On the multi-process mode: is the model one primary system-wide with N secondaries, or can multiple primaries coexist? The use case I want supported is several processes each managing a disjoint set of controllers.

tests/mproc.c
```diff
@@ -0,0 +1,222 @@
...
+#include <libxnvme.h>
+
+#define MPROC_FILL_PATTERN "!"
+#define MPROC_SLBA 0
+#define MPROC_NRUNS 2
+
+static char g_exe[PATH_MAX];
+
+static int
+run_secondary(struct xnvme_cli *cli, uint32_t nsid)
```
karlowich: Can we look into whether it makes sense to instrument the second process from the python test, instead of having a self invoking C program?

karlowich (changes requested):
Looks promising, but I have a couple of comments.

I'm unsure of this commit: "feat(be): publish identify data in mproc so secondaries can derive geometry"
It is a lot of code to add to our generic path for handling a uPCIe specific thing. Perhaps we can find another way? Maybe this can be handled when starting the primary, perhaps we don't need the "manual" derivation of geometry when using the multiprocess path?

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: fixed

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: Hmmm, no but git's automatic rebase must have thought that I added these back. I'll remove it

docs/tutorial/multi_process/index.rst
naddinadja: Technically yes. As we talked about in person, we use a uint64_t as bitmask to indicate which queues are available for secondary processes, so we can only have 64 queues. NVMe reserves the first queue (so idx 0) for the admin queue, but we don't expose this one via shared memory, since it's owned by the primary process, and therefore idx 0 in the shared queue-array is just unused (so keep the mapping of ids and position in array). So we can only have 63 queues, and I'll change the mproc struct to not have `NQUEUES_MAX + 1` as the array size, but just `NQUEUES_MAX`.

docs/tutorial/multi_process/index.rst
naddinadja: Good idea 😎

tests/mproc.c
naddinadja: I've completely rewritten the test-commit, so that this is instrumented through python

naddinadja:
> **Why pre-allocate qpairs?** It's a behavioral change I'd prefer not to introduce. We already allocate one qpair per controller to support the sync-interface on the user-space backends, and even that single allocation has been a source of grief on its own. Users do not expect the library to consume/allocate scarce resources like qpairs unsolicited.

This behaviour is borrowed from SPDK, which pre-allocates a given amount of queuepars. It allows users to use the multi-process mode without establishing communication between the primary and secondary processes. However, we could change it to *not* pre-allocate them, and if a user wants to preallocate, they can just do so in their own implementation. The main use case for this, for now, is HOMI which already has IPC, so we can just allocate on-demand. What do you think of this?

> **Tests:** the existing tests in this area should be sensitive to this change. Are they still passing, or have they been masked out?

I haven't removed any existing tests, but since the last time you looked, I've implemented the discussed changes, such that all existing pcie tests for SPDK and uPCIe are run with multi-process mode, and the instrumentation of setting up the primary process has been moved to the python test setup.

> On the multi-process mode: is the model one primary system-wide with N secondaries, or can multiple primaries coexist? The use case I want supported is several processes each managing a disjoint set of controllers.

The model is one primary per device. A single process should in theory be both the primary for one device and secondary for another device at the same time. Whether a process is primary or secondary is just defined by the xnvme_opts given when calling xnvme_dev_open()

naddinadja:
> I'm unsure of this commit: "feat(be): publish identify data in mproc so secondaries can derive geometry" It is a lot of code to add to our generic path for handling a uPCIe specific thing. Perhaps we can find another way?

In my eyes, multi-process mode is not an uPCIe specific thing, but just a mode, which is currently only supported by the SPDK and uPCIe backends. I would prefer to keep it on the "general" level.

> Maybe this can be handled when starting the primary, perhaps we don't need the "manual" derivation of geometry when using the multiprocess path?

Do you refer to only the old requirement of having to call derive_geo() after opening? If so, yes I agree, and I added it as suggested. If you refer to something else, please explain more :)

karlowich:
> I would prefer to keep it on the "general" level

What I was referring to is that you are introducing hook now that helps uPCIe manage it's internal  `xnvme_be_upcie_mproc`. Perhaps, since this `mproc` struct is specific to uPCIe it doesn't need a general hook.
Instead, you could populate it in `xnvme_be_upcie_ctrlr_init()` when it is created.

naddinadja:
> What I was referring to is that you are introducing hook now that helps uPCIe manage it's internal `xnvme_be_upcie_mproc`. Perhaps, since this `mproc` struct is specific to uPCIe it doesn't need a general hook. Instead, you could populate it in `xnvme_be_upcie_ctrlr_init()` when it is created.

Valid point - I have changed the way that it's setup such that publishing the identify data (and deriving it) is integrated into the setup of the primary and secondary processes, instead of being its own hooks. These changes ended up making that specific commit very small, and since sharing of idfy data is a required part of the multi-process mode, I figured it just belonged to the previous commit regarding mproc support in upcie.

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -215,36 +432,341 @@ xnvme_be_upcie_ctrlr_init(struct xnvme_dev *dev)
...
 
-	err = nvme_controller_create_io_qpair(ctrlr->ctrl, &ctrlr->sync, 16);
+	if (dev->opts.proc_role == XNVME_PROC_PRIMARY) {
+		struct xnvme_be_upcie_mproc *mproc;
+		struct hostmem_heap *heap = &g_upcie_rte.heap;
+		char shm_name[64];
+		int shm_fd;
+		int nqueues = XNVME_BE_UPCIE_NQUEUES_MAX - 1;
+		int depth = XNVME_BE_UPCIE_QUEUE_DEPTH;
+		size_t shm_size = sizeof(*mproc);
```
karlowich: Could we make a helper function out of this huge if block? The ctrlr_init function is already very large

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -215,36 +432,341 @@ xnvme_be_upcie_ctrlr_init(struct xnvme_dev *dev)
...
+						  &ctrlr->sync);
+			if (qpid < 0) {
+				XNVME_DEBUG("FAILED: claim sync qpair: %d", qpid);
+				_upcie_free_preallocated_queues(ctrlr);
+
+				munmap(mproc, shm_size);
+				close(shm_fd);
+				shm_unlink(shm_name);
+				ctrlr->shm_fd = -2;
+				goto fail_close;
```
karlowich: Looks like there's a similar section of code in ctrlr_attach, perhaps we can make a helper out of it?

karlowich (changes requested):
This is starting to look really nice.
This commit: feat(be/upcie): add multi-process support with shared queue pool
Made heaps of changes to `xnvme_be_upcie_dev.c`, I think there's potential to clean it up and make some helpers etc.

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: I assume you mean that these to section are similar:

```C
// in xnvme_be_upcie_ctrlr_init()
{
	uint8_t *bar0 = ctrlr->ctrl->func.bars[0].region;
	uint64_t cap = nvme_mmio_cap_read(bar0);
	int dstrd = nvme_reg_cap_get_dstrd(cap);
	int qpid;

	qpid = _upcie_claim_qpair(ctrlr, heap->memory.virt, bar0, dstrd,
						&ctrlr->sync);
	if (qpid < 0) {
		XNVME_DEBUG("FAILED: claim sync qpair: %d", qpid);
		_upcie_free_preallocated_queues(ctrlr);

		munmap(mproc, shm_size);
		close(shm_fd);
		shm_unlink(shm_name);
		ctrlr->shm_fd = -2;
		goto fail_close;
	}
}


// in xnvme_be_upcie_ctrlr_attach()
{
	uint8_t *bar0 = ctrlr->ctrl->func.bars[0].region;
	uint64_t cap = nvme_mmio_cap_read(bar0);
	int dstrd = nvme_reg_cap_get_dstrd(cap);
	int qpid;

	ctrlr->ctrl->timeout_ms = nvme_reg_cap_get_to(cap) * 500;

	qpid = _upcie_claim_qpair(ctrlr, ctrlr->imported_hugepage.virt, bar0, dstrd,
					&ctrlr->sync);
	if (qpid < 0) {
		XNVME_DEBUG("FAILED: claim sync qpair: %d", qpid);
		atomic_fetch_sub_explicit(&mproc->refcount, 1, memory_order_release);
		pci_bar_unmap(&ctrlr->ctrl->func.bars[0]);
		pci_func_close(&ctrlr->ctrl->func);
		goto fail_hugepage;
	}

	snprintf(dev->ident.kernel_driver, sizeof(dev->ident.kernel_driver), "%s",
		driver_name);
}
```

I agree that they look similar, since they define the same variables and both call `_upcie_claim_qpair()`, but the error path is completely different, and in the attach function I also need the variables to set the timout. So the only "real" thing I can simplify here is the call to `_upcie_claim_qpair()` - which already is a helper function 😅

naddinadja: The "cleanest" refactor I can do is moving the variable derivation to the helper function such that we have the setup below, but idk if that's the improvement that you'd hoped for 

```C
int
_upcie_claim_qpair(struct xnvme_be_upcie_ctrlr *ctrlr, void *ring_base)
{ /* ... now derives bar0, dstrd and dest from given ctrlr */  }

// in xnvme_be_upcie_ctrlr_init()
{
	int qpid = _upcie_claim_qpair(ctrlr, heap->memory.virt);
	if (qpid < 0) {
		XNVME_DEBUG("FAILED: claim sync qpair: %d", qpid);
		_upcie_free_preallocated_queues(ctrlr);

		munmap(mproc, shm_size);
		close(shm_fd);
		shm_unlink(shm_name);
		ctrlr->shm_fd = -2;
		goto fail_close;
	}
}


// in xnvme_be_upcie_ctrlr_attach()
{
	uint8_t *bar0 = ctrlr->ctrl->func.bars[0].region;
	uint64_t cap = nvme_mmio_cap_read(bar0);
	int qpid;

	ctrlr->ctrl->timeout_ms = nvme_reg_cap_get_to(cap) * 500;

	qpid = _upcie_claim_qpair(ctrlr, ctrlr->imported_hugepage.virt);
	if (qpid < 0) {
		XNVME_DEBUG("FAILED: claim sync qpair: %d", qpid);
		atomic_fetch_sub_explicit(&mproc->refcount, 1, memory_order_release);
		pci_bar_unmap(&ctrlr->ctrl->func.bars[0]);
		pci_func_close(&ctrlr->ctrl->func);
		goto fail_hugepage;
	}

	snprintf(dev->ident.kernel_driver, sizeof(dev->ident.kernel_driver), "%s",
		 driver_name);
}
```

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: i agree! fixing

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: Oh wait, I realise now, that the part of the code that looked duplicated in ctrlr_init might also go away, if I go towards not pre-allocating all queue pairs as Simon suggested, so this might have not mattered, but I guess it was a good exercise. I'll add a new commit with this change (removing pre-allocation of all queue pairs), just so we can compare the two.

naddinadja:
@safl I have added a new commit, which removes the preallocation of all queue pairs. I added as a new commit, rather than squashing it into the existing commits, such that we can easily switch back to the old implementation, if that's what we end up going with. But if this is how we want it to be, I will of course merge it into the previous commits :)

safl:
> > On the multi-process mode: is the model one primary system-wide with N secondaries, or can multiple primaries coexist? The use case I want supported is several processes each managing a disjoint set of controllers.
> 
> The model is one primary per device. A single process should in theory be both the primary for one device and secondary for another device at the same time. Whether a process is primary or secondary is just defined by the xnvme_opts given when calling xnvme_dev_open()

Say i have devices **A**, and **B**. Now, process ``X`` starts and is the primary for **A**, process ``Y`` starts and is the primary for **B**. Now process ``Z`` starts as a secondary that needs device **B**. How does ``Z`` know, or how it is parametrized to know, that ``Y`` is the primary for **B**?

naddinadja:
> Say i have devices **A**, and **B**. Now, process `X` starts and is the primary for **A**, process `Y` starts and is the primary for **B**. Now process `Z` starts as a secondary that needs device **B**. How does `Z` know, or how it is parametrized to know, that `Y` is the primary for **B**?

By opening device **B** with xnvme_opts.proc_role = secondary (there's a proc_role enum) in process `Z`, upcie will try to fetch the shared device configs at shm name "/xnvme-upcie-<device bdf>". So process `Z` will have to use xnvme_dev_open() as usual, but just set the proc_role in the opts too. If you try to open with a different backend than the primary process, it will fail as there the shared memory does not exist.

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,177 @@
...
+
+**Kernel-based backends** (e.g., ``linux``, ``io_uring``)
+  The OS handles concurrent device access natively. Multiple processes can
+  open the same device independently and ``proc_role`` has no effect.
+
+**vfio** (``libvfn``)
+  Single-process only. The VFIO group-based API allows a group to be in only
+  one IOMMU container at a time, so a second process cannot share the
+  primary's IOMMU domain. ``proc_role`` has no effect for this backend.
+
```
safl, line 46: What about SPDK?

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,177 @@
...
+``XNVME_PROC_PRIMARY``
+  Initializes the hardware, pre-allocates the shared queue pool, and publishes
+  shared state. The primary must open the device before any secondary process
+  attempts to attach.
+
+``XNVME_PROC_SECONDARY``
+  Attaches to an already-initialized controller without touching the hardware.
+  Returns an error if no primary is running for the given device.
+
```
safl, line 31: What about SPDK? I seems like the SPDK backend uses a different interface that what is described here.

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,177 @@
...
+Note that the only difference between the two snippets is ``proc_role``. Once
+the device is open, all queue operations work identically.
+
+Each process may open the same device multiple times (for example, once per
+namespace) and init as many queues as needed, subject to the pool capacity
+described below.
+
+Queue pool
+==========
+
```
safl, line 126: Is this still the behavior? Perhaps pre-allocation etc. should be parameterized, would that make sense?

docs/tutorial/multi_process/index.rst
```diff
@@ -0,0 +1,177 @@
...
+================
+
+**xNVMe** supports multiple processes accessing the same userspace NVMe
+device simultaneously. The primary process allocates all hardware queue pairs
+upfront; each process then claims one or more pairs exclusively for I/O.
+No queue pair is used by more than one process at a time.
+
+Roles
+=====
+
```
safl, line 14: Conceptually how is this managed? How is shared memory used? How are the roles setup or negotiated?

safl (commented):
I have a lot of questions some inline some follow below. Also, remember to write commit-messages and sign off the commits.

In general, since this is such a fundamental and core change, it needs some description of how it is designed and how to use it, the latter does have some docs. which is great. But to the former a description of the design that you are implementing to provide "Multi-process support for xNVMe" would make it a lot clearer. From the top the immediate questions are:

* Is this a general interface/infrastructure in xNVMe that enables all backends to provide multi-process support? Is it specific to user-space drivers? Is it a an intefrace for backends to implement? What is the scope here? Is it just a front-loaded knob wired down to one backend implementation?
* How is admin-queue-sharing handled? E.g. how does a secondary process submit admin commands?
* How are the roles divided / setup? E.g. how to become a PRIMARY vs SECONDARY? Is it through parameterization, some sort of negotiation? What is the protocol for sharing between processes and what is shared?
* Is it aligned to the SPDK multi-process mode, does it replace it, work with it, what is the state of alignment here?

Nitpick: "single userspace NVMe device", physical NVMe devices are not in "userspace" drivers can be implemented for and run in user space, assuming that is the goal described here, sharing NVMe driver-state between multiple processes?

naddinadja:
> I have a lot of questions some inline some follow below. Also, remember to write commit-messages and sign off the commits.

Yes of course, my editor usually automatically signs off, so I must have messed something up, when I've been rebasing and adding things to prior commits. Will fix that of course.

> In general, since this is such a fundamental and core change, it needs some description of how it is designed and how to use it, the latter does have some docs. which is great. But to the former a description of the design that you are implementing to provide "Multi-process support for xNVMe" would make it a lot clearer.

I will try to write a much more description. Do you want this in the docs, and if so, is the current location of the existing docs right? Right now I put it under "Tutorials", but I found it a bit difficult to place.

> * Is this a general interface/infrastructure in xNVMe that enables all backends to provide multi-process support? Is it specific to user-space drivers? Is it a an intefrace for backends to implement? What is the scope here? Is it just a front-loaded knob wired down to one backend implementation?

In my eyes, "multi-process support" is a general thing. However, kernel backends already support multiple processes to open and communicate with the device, so no change is essentially needed here. It's only for userspace backends that we need to implement something. Therefore, I've made it such that it's for the different backends to implement - also because it's very different what the different backends need for this to work. For now, I've added the backend implementations for SPDK (since it was fairly simple) and for uPCIe (since that's what I need for the work on HOMI), but I don't see any inherent problem in it being implemented for other userspace backends, other than it being tricky with VFIO and IOMMU groups.

> * How is admin-queue-sharing handled? E.g. how does a secondary process submit admin commands?

Admin queues are not shared, only the primary process can submit admin commands. This is something we have talked about in person, ~~and also how SPDK is implemented~~ no, I see that secondaries *can* submit admin commands to e.g. allocate new queues.

> * How are the roles divided / setup? E.g. how to become a PRIMARY vs SECONDARY? Is it through parameterization, some sort of negotiation? What is the protocol for sharing between processes and what is shared?
>
>     * Is it aligned to the SPDK multi-process mode, does it replace it, work with it, what is the state of alignment here?

As I wrote in my last comment: You set it through parameterization with xnvme_opts.proc_role. There's an enum `xnvme_proc_role`, that you use to define whether you want to start as either the primary or secondary - or just "single" for normal non-multi-process-mode, which is the default. 
It _kinda_ mimics the SPDK interface, and I gotta be honest that I thought my implementation was closer than I now realise. I have changed a bit along the way and strayed away from the SPDK setup. 
* With SPDK, a user has to define an `shm_id`, and optionally the process role too. If you do not define the process role (which is the common use case), the processes will coordinate automatically (become primary if first process, else secondary). The primary owns the shm region and puts all controllers for all devices in this region, and is the "owner" of all opened devices.
* In the xNVMe implementation, I derive the shm_id automatically from the device bpf, which means that there's a shared memory region for each device. I don't know how useful (or damaging?) it is, but with the xNVMe a single process **A** can act as primary for one device, and secondary for another device, given a second process **B** has opened that device as primary process.

> Nitpick: "single userspace NVMe device", physical NVMe devices are not in "userspace" drivers can be implemented for and run in user space, assuming that is the goal described here, sharing NVMe driver-state between multiple processes?

I will fix this ^^

docs/tutorial/multi_process/index.rst
naddinadja: I am unsure what you mean. With SPDK, you use the same proc_role option to open as a primary/secondary, but the implementation is different (talk directly to SPDK's multi-process mode instead of implementing anything ourselves). But from the user's perspective, the interface is the same.

safl:
> > I have a lot of questions some inline some follow below. Also, remember to write commit-messages and sign off the commits.
> 
> Yes of course, my editor usually automatically signs off, so I must have messed something up, when I've been rebasing and adding things to prior commits. Will fix that of course.
> 
> > In general, since this is such a fundamental and core change, it needs some description of how it is designed and how to use it, the latter does have some docs. which is great. But to the former a description of the design that you are implementing to provide "Multi-process support for xNVMe" would make it a lot clearer.
> 
> I will try to write a much more description. Do you want this in the docs, and if so, is the current location of the existing docs right? Right now I put it under "Tutorials", but I found it a bit difficult to place.

Great. I guess a usage guide would fit best in the tutorial section, however, when it also has implementation details and "capabilities" tied to it, then the background section might be a better place to put it.

> 
> > * Is this a general interface/infrastructure in xNVMe that enables all backends to provide multi-process support? Is it specific to user-space drivers? Is it a an intefrace for backends to implement? What is the scope here? Is it just a front-loaded knob wired down to one backend implementation?
> 
> In my eyes, "multi-process support" is a general thing. However, kernel backends already support multiple processes to open and communicate with the device, so no change is essentially needed here. It's only for userspace backends that we need to implement something. Therefore, I've made it such that it's for the different backends to implement - also because it's very different what the different backends need for this to work. For now, I've added the backend implementations for SPDK (since it was fairly simple) and for uPCIe (since that's what I need for the work on HOMI), but I don't see any inherent problem in it being implemented for other userspace backends, other than it being tricky with VFIO and IOMMU groups.
> 

By a "general thing" i mean, is it an entirely new or an extension to existing backend or platform interfaces, what are the mechanics like. There are mentions of ``proc_role`` but without description of what that entails, by the first commit it just looks like some cli-flags... then ``mproc`` pops up as a backend-capability. This stuff needs to be described in a conceptually sound manner and concepts introduced. There is a mention that about SPDK achieving something without ``core_mask`` why is that significant?

> > * How is admin-queue-sharing handled? E.g. how does a secondary process submit admin commands?
> 
> Admin queues are not shared, only the primary process can submit admin commands. This is something we have talked about in person, ~and also how SPDK is implemented~ no, I see that secondaries _can_ submit admin commands to e.g. allocate new queues.

In the usage guide this should be emphasized that by design then a secondary process cannot do admin-commands. And a description why it is not needed and what data is available in the device-handle that one would otherwise send admin-commands for e.g. device geometry. 

> 
> > * How are the roles divided / setup? E.g. how to become a PRIMARY vs SECONDARY? Is it through parameterization, some sort of negotiation? What is the protocol for sharing between processes and what is shared?
> >   
> >   * Is it aligned to the SPDK multi-process mode, does it replace it, work with it, what is the state of alignment here?
> 
> As I wrote in my last comment: You set it through parameterization with xnvme_opts.proc_role. There's an enum `xnvme_proc_role`, that you use to define whether you want to start as either the primary or secondary - or just "single" for normal non-multi-process-mode, which is the default. It _kinda_ mimics the SPDK interface, and I gotta be honest that I thought my implementation was closer than I now realise. I have changed a bit along the way and strayed away from the SPDK setup.

Why drift from that? It this is aligned to SPDK then the user has fewer knobs to tweak. Things like lazily allocating qeueus makes sense to me that we enable instead of pre-allocating as much as possible, but in general, why not just allign with the SPDK parameterization?

> 
>     * With SPDK, a user has to define an `shm_id`, and optionally the process role too. If you do not define the process role (which is the common use case), the processes will coordinate automatically (become primary if first process, else secondary). The primary owns the shm region and puts all controllers for all devices in this region, and is the "owner" of all opened devices.
> 
>     * In the xNVMe implementation, I derive the shm_id automatically from the device bpf, which means that there's a shared memory region for each device. I don't know how useful (or damaging?) it is, but with the xNVMe a single process **A** can act as primary for one device, and secondary for another device, given a second process **B** has opened that device as primary process.

Users of these things would today do things the SPDK way, if we follow the same mechanics then there are fewer surprises and it should ideally enable the xNVMe user to swap the SPDK NVMe drivers with the uPCIe ones, without changing any pareters. E.g. we can do comparative studies of HOMI usage without having to change anything in the setup except for the backend name.

> 
> 
> > Nitpick: "single userspace NVMe device", physical NVMe devices are not in "userspace" drivers can be implemented for and run in user space, assuming that is the goal described here, sharing NVMe driver-state between multiple processes?
> 
> I will fix this ^^

naddinadja:
> By a "general thing" i mean, is it an entirely new or an extension to existing backend or platform interfaces, what are the mechanics like. There are mentions of `proc_role` but without description of what that entails, by the first commit it just looks like some cli-flags... then `mproc` pops up as a backend-capability. This stuff needs to be described in a conceptually sound manner and concepts introduced. There is a mention that about SPDK achieving something without `core_mask` why is that significant?

I will make sure to describe this much better. Do you mean only in commit-messages and PR description or in the docs (or both)?

> In the usage guide this should be emphasized that by design then a secondary process cannot do admin-commands. And a description why it is not needed and what data is available in the device-handle that one would otherwise send admin-commands for e.g. device geometry.

I am working on chancing it such that secondaries can submit admin commands too. After I realised this is how SPDK does it, I find it cleaner to do it this way.

> Why drift from that? It this is aligned to SPDK then the user has fewer knobs to tweak. Things like lazily allocating qeueus makes sense to me that we enable instead of pre-allocating as much as possible, but in general, why not just allign with the SPDK parameterization?
> 
> ...
> 
> Users of these things would today do things the SPDK way, if we follow the same mechanics then there are fewer surprises and it should ideally enable the xNVMe user to swap the SPDK NVMe drivers with the uPCIe ones, without changing any pareters. E.g. we can do comparative studies of HOMI usage without having to change anything in the setup except for the backend name.

My approach was to have 1 memory section per device, since devices are opened individually with `xnvme_dev_open()`. In SPDK it is environment-scoped instead, meaning the controllers of multiple devices can be in the same shared memory region. With my current approach (device-scoped memory regions), the "SPDK way" of just giving an shm_id would require extra logic on the auto-detection of who's the primary/secondary. If I recall correctly, Karl also mentioned that he and Ankit had tried implementing something like this, where it was tricky with the auto-detection. I think it's possible though, and could look into it!
I would prefer not to change to an environment-scoped setup like SPDK -  I think it's possible too, but it would probably require a new "level" of environment initialisation, which kinda goes against our current setup of xNVMe.

naddinadja:
I will close this, as I have used your feedback to replace it with #696

PR #687, opened 2026-06-11
ci(freebsd): use patched nosi image and collect serial console

safl:
Bump [guest_image] sha to the freebsd-14-headless build cut from safl/nosi#15: firstboot_freebsd_update is disabled in rc.conf, so the consumer first boot no longer pulls thousands of patches before LOGIN (125s+ -> 38s locally).

Extend 'Collect guest serial console' to freebsd and template the serial.output path on ${os}-${ver}-amd64. No behavioural change for debian; freebsd failures now ship a serial console artifact.


karlowich (approved):
LGTM

PR #689, opened 2026-06-16
feat(be): allow upcie and upcie-cuda to share an NVMe controller simultaneously

karlowich:
This is #665 once more, since it was accidentally merged without running the maas tests.

-----

This series allows the `upcie` backend variants (`upcie` and `upcie-cuda`) to
share the same NVMe controller at the same time, enabling dual-backend use
cases such as opening a single device with both `upcie` and `upcie-cuda`
concurrently.

### Mechanism

Controller state in the `cref` table is keyed by backend name, so a second
backend config opening an already-open URI is normally rejected with `-EBUSY`.
This is relaxed for the `upcie` variants: because they share the same
underlying NVMe controller infrastructure, they accept each other's crefs
despite the inexact name match.

### Changes

- **`feat(be)`** — Relax the cref name-match check so `upcie` and `upcie-cuda`
  reuse the same controller for a given URI instead of conflicting.
- **`feat(upcie)`** — Move the CUDA runtime lifecycle (`_cuda_rte_init` /
  `_cuda_rte_term`) out of `ctrlr_init` / `ctrlr_term` and into `dev_open` /
  `dev_close`, so CUDA is initialised for every `upcie-cuda` handle regardless
  of which backend performed the primary controller open. The driver check
  (`uio_pci_generic` required, `vfio-pci` rejected) is kept in both
  `ctrlr_init` (primary path) and `dev_open` (secondary path, where
  `ctrlr_init` is skipped).
- **`feat(cli)`** — Add `XNVME_CLI_OPT_ALT_BE` / `--alt-be` (a second backend
  name independent of `--be`). This is purely a CLI/test concern, used so tests
  can open the same device with two different backends; there is no
  primary/secondary hierarchy and the order is irrelevant.
- **`test(dual-be)`** — New `dual-be` test tool with two subcommands:
  - `io-cmp`: write a known pattern via `--be`, read it back via `--alt-be`,
    and compare.
  - `idfy-cmp`: issue Identify Controller via both backends and byte-compare
    the responses.
- **`test(cijoe)`** — pytest-based integration tests:
  - `test_open_dev_twice`: opens the same device twice with identical backends
    to verify same-name cref reuse.
  - `test_io_cmp` / `test_idfy_cmp`: exercise the dual-backend path for
    `upcie` / `upcie-cuda` pairs (skipped for backends without an alternate).
- **`docs(upcie)`** — Document dual-backend operation for `upcie` and
  `upcie-cuda`.

### User-facing impact

Minimal and intentionally so. Today the only supported combination is mixing
`upcie` and `upcie-cuda`. Just as a device URI like `0000:01:00.0` can already
be opened multiple times with the same backend, it can now also be opened with
these two different backends at once. No new concepts are exposed to ordinary
users.

### DIFF
--- a/cijoe/tests/logic/test_xnvme_tests_dual_be.py
+++ b/cijoe/tests/logic/test_xnvme_tests_dual_be.py
@@ -1,0 +1,44 @@
+import pytest
+
+from ..conftest import xnvme_parametrize
+
+
+def get_alt_be(be: str):
+    if be == "upcie":
+        return "upcie-cuda"
+    elif be == "upcie-cuda":
+        return "upcie"
+    else:
+        return None
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be"])
+def test_open_dev_twice(cijoe, device, be_opts, cli_args):
+    # We are repurposing "xnvme_tests_dual_be idfy-cmp" to verify that opening the same
+    # device twice with the same backend works.
+    err, _ = cijoe.run(
+        f"xnvme_tests_dual_be idfy-cmp {device['uri']} --be {be_opts['be']} --alt-be {be_opts['be']} --dev-nsid {device['nsid']}"
+    )
+    assert not err
+
+
+@xnvme_parametrize(labels=["dev", "cuda"], opts=["be"])
+def test_io_cmp(cijoe, device, be_opts, cli_args):
+    alt_be = get_alt_be(be_opts["be"])
+    if not alt_be:
+        pytest.skip(reason=f"[be={be_opts['be']}] does not have an alternate backend")
+    err, _ = cijoe.run(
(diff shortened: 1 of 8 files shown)


karlowich:
@safl @naddinadja I tracked down what caused the CI to fail. My test was doing idfy-ns, which reports some numbers that are changing throughout the lifetime of the device (NUSE - Namespace utilization). The test is sending the identify command twice and comparing the results byte-wise. With real NVMe devices there is the possibility of a mismatch since the NUSE numbers are changing slower than an emulated QEMU NVMe device.
The fix is to send idfy-ctrlr instead.

safl (approved):
lgtm

naddinadja (approved):
great! :)

PR #690, opened 2026-06-17
fix(be/spdk): use shm_id as trigger for multi-process mode

naddinadja:
In SPDK, the shm_id option is the signal for using multi-process mode, and therefore it should be the same in our code. Before, we were using core_mask which was misleading, as this option is a signal of which CPU cores to use - not necessarily that the user wants multi-process mode.

Also, move the core_mask and main_core out of the if-statement, as these options are not necessarily tied to multi-process mode and can be given without the user needing multi-process mode. The doc-strings have been updated accordingly.


karlowich (dismissed):
LGTM

karlowich (approved):
LGTM

safl (approved):
Nice one.

PR #691, opened 2026-06-19
feat(upcie): configurable heap sizes with workload-aware sizing in xnvmeperf

### LINKED ISSUE #642: Hugepage memory exhausted quickly for uPCIe backend
The uPCIe backend uses a single 256 MB hostmem_heap (g_upcie_rte.heap) for both internal driver allocations (queue pairs, admin buffers, PRP pools) and user-facing buffer allocations (xnvme_buf_alloc). Each controller needs ~4.2 MB and each IO qpair ~4.1 MB (dominated by 4 MB PRP pool). When running xnvmeperf across 8+ controllers, e.g.:

```bash
 xnvmeperf run --iopattern randread --qdepth 128 --iosize 512 --runtime 10 --cpumask 0xff /dev/nvme0n1 [dev1, dev2, ..., devN]
```

the heap is exhausted because IO queue allocations, PRP pools, and user data buffers all compete for the same 256 MB.

The target is to support up to 64 IO qpairs, spread across a single or multiple controllers. For xnvmeperf this means exercising IO with up to 64 locally attached NVMe devices.

 Split into:
 - RTE heap (rte_heap, 768 MB): qpairs, admin, PRP pools for up to 64 controllers + 64 async qpairs
 - Buffer heap (buf_heap, 256 MB): user data buffers
 - Total: 1 GB globally for the process. Multi-process improvements deferred to future work.

The memory subsystem has several areas that could be improved: auto-growing heaps, dynamic PRP pool sizing (currently static at 1024 entries per qpair), smarter hugepage reservation, etc. Rather than tackling all of that at once, this is a small incremental step: separate the runtime heap (rte_heap) from the buffer heap (buf_heap). This unblocks multi-controller workloads now, and creates the right structure for future optimizations on each heap independently.

karlowich:
Closes #642

------

 Make the uPCIe host/device heap sizes configurable, raise the host default,
  and have xnvmeperf size the heaps to its workload so large runs no longer
  run out of memory.

  ## Changes

  1. **Configurable heap sizes** (`feat(upcie): add host_heap_size and
     device_heap_size opts`). Adds `host_heap_size` (upcie host DMA heap) and
     `device_heap_size` (upcie-cuda GPU heap) to `struct xnvme_opts`, letting
     callers override the heap sizes. Both keep their existing behavior when left
     unset (`0`). The values are surfaced in `xnvme_opts_yaml()` for debugging.
     For upcie-cuda opened after upcie in the same process, host_heap_size has no 
     effect since the call to open upcie already initialized the host heap.

  2. **Raise the default host heap 256 MiB → 1 GiB** (`fix(upcie): increase
     default host heap from 256 MiB to 1 GiB`). The host heap backs both NVMe
     control structures and user IO buffers: each controller alone consumes
     ~8 MiB (admin and sync qpairs each carry a 4 MiB PRP pool plus two 64 KiB
     SQ/CQ buffers), and `xnvme_buf_alloc` draws from the same heap, so multiple
     devices with large buffers exhaust 256 MiB quickly. Introduces
     `XNVME_BE_UPCIE_DEFAULT_HEAP_SIZE` as the shared default for the host and
     device heaps.

  3. **Workload-aware sizing in xnvmeperf** (`feat(xnvmeperf): size uPCIe heaps
     to the workload`). Running many devices, many queues, or deep queues
     exhausts even the 1 GiB default. xnvmeperf now derives `host_heap_size` /
     `device_heap_size` from `ndevs`, `nqueues`, `qdepth`, and `iosize` before
     opening the devices. Each queue gets a generous fixed control slab; data
     buffers are sized onto the host heap (host backends) or the device heap
     (CUDA backend, scaled by qdepth). The estimate deliberately overshoots to
     keep the math simple.

  ## Notes

  - The new opts are programmatic-only (set via `struct xnvme_opts`); no CLI flag
    is added, since xnvmeperf derives the sizes itself.

### DIFF
--- a/include/libxnvme_opts.h
+++ b/include/libxnvme_opts.h
@@ -50,7 +50,7 @@ struct xnvme_opts {
 	uint32_t command_timeout;  ///< SPDK fabrics: enable io command timeout
 	uint32_t spdk_fabrics;     ///< Is assigned a value by backend if SPDK uses fabrics
 	uint32_t keep_alive_timeout_ms; ///< SPDK fabrics: set keep alive timeout
-	size_t host_heap_size;   ///< upcie: host DMA heap size in bytes (0 = default 256 MiB)
+	size_t host_heap_size;          ///< upcie: host DMA heap size in bytes (0 = default 1 GiB)
 	size_t device_heap_size; ///< upcie-cuda: GPU device heap size in bytes (0 = default 1 GiB)
 };
 
--- a/include/libxnvme_opts.h
+++ b/include/libxnvme_opts.h
@@ -50,6 +50,8 @@ struct xnvme_opts {
 	uint32_t command_timeout;  ///< SPDK fabrics: enable io command timeout
 	uint32_t spdk_fabrics;     ///< Is assigned a value by backend if SPDK uses fabrics
 	uint32_t keep_alive_timeout_ms; ///< SPDK fabrics: set keep alive timeout
+	size_t host_heap_size;   ///< upcie: host DMA heap size in bytes (0 = default 256 MiB)
+	size_t device_heap_size; ///< upcie-cuda: GPU device heap size in bytes (0 = default 1 GiB)
 };
 
 /**
--- a/lib/upcie/xnvme_be_upcie.h
+++ b/lib/upcie/xnvme_be_upcie.h
@@ -12,6 +12,12 @@
 #define _UPCIE_WITH_NVME
 #include <upcie/upcie.h>
 
+/**
+ * Default heap size used for the host DMA heap (upcie) and the GPU device heap
+ * (upcie-cuda) when xnvme_opts leaves the size unset (0)
+ */
+#define XNVME_BE_UPCIE_DEFAULT_HEAP_SIZE (1024ULL * 1024 * 1024)
+
 struct xnvme_queue_upcie {
 	struct xnvme_queue_base base;
 	struct nvme_qpair qpair;
--- a/lib/upcie/xnvme_be_upcie_cuda_dev.c
+++ b/lib/upcie/xnvme_be_upcie_cuda_dev.c
@@ -36,7 +36,7 @@ _cuda_rte_init(size_t heap_size)
 	}
 
 	if (!heap_size) {
-		heap_size = 1024 * 1024 * 1024;
+		heap_size = XNVME_BE_UPCIE_DEFAULT_HEAP_SIZE;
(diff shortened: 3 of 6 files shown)


safl (changes requested):
For clarity, ``feat(upcie): add host_heap_size and device_heap_size opts`` should be split.

The change of default heap sizes is the primary point of contention and should stand on its own, clearly explaining what changes, from what to what, and why.

The addition of configuration options to override the defaults is a separate feature and should be reviewed independently.

Regarding the second commit, installing uPCIe headers system-wide is not an acceptable solution to a build issue.

One of the main advantages of a vendored header-only library is that it does not require installation. Installing the headers into global include paths largely defeats that benefit and introduces packaging concerns. We should address the underlying build issue rather than work around it by placing third-party headers in default system paths.

karlowich:
> For clarity, feat(upcie): add host_heap_size and device_heap_size opts should be split.

Funny, I just squashed them right before pushing. It feels weird to add a default value, and changing it immediately after.

> We should address the underlying build issue rather than work around it by placing third-party headers in default system paths.

Well the issue is that we need to import the uPCIe headers in `libxnvme_cuda.h`. Since, this is a public xNVMe header, the consumer needs to have the uPCIe headers available on their system. A user is not supposed to install the standalone uPCIe, since that can create unexpected problems when compiling xNVMe. Installing uPCIe separately also defeats the purpose of the vendored header-only library. I can't spot any real solutions without completely changing the relationship between uPCIe and xNVMe.

----

By the way, I'm working on using the new heap size opts in xnvmeperf. I will make this into a draft until that is done.

karlowich:
@safl I dropped the commit about the headers, since it is somewhat unrelated and clearly requires some more thinking.

safl (commented):
Great, this is a good step towards getting this to function better, we probably need more eventually, like a protocol for each backend instance saying how much it **needs/wants** such that it can be checked whether the pre-allocated heaps have enough to satisfy it, instead of it crashing later as it and its peers start to consume the memory.
Anyway, that can all be addressed later, thus just one small thing, the commit message mention an undefined concept "compat-family secondary", please remove that and instead just describe what the state is.

safl (approved):
Awesome.

karlowich:
> "compat-family secondary"

Right that is an abandoned term. I removed it

naddinadja (approved):
lgtm :)

PR #692, opened 2026-06-19
fix(cref): use strcmp for URI comparison in cref lookup

### LINKED ISSUE #624: Controller ref-count issues

karlowich:
strncmp(..., XNVME_IDENT_URI_LEN) can treat two distinct URIs as equal when they share a common prefix of that length. URIs are null-terminated strings stored in fixed-size buffers, so strcmp is both simpler and semantically correct.

This is the last piece required for closing #624.

### DIFF
--- a/lib/xnvme_be_cref.c
+++ b/lib/xnvme_be_cref.c
@@ -15,7 +15,7 @@ xnvme_be_cref_get(const char *uri)
 		if (!g_cref_table[i].ctrlr) {
 			continue;
 		}
-		if (strncmp(g_cref_table[i].uri, uri, XNVME_IDENT_URI_LEN)) {
+		if (strcmp(g_cref_table[i].uri, uri)) {
 			continue;
 		}
 		if (g_cref_table[i].refcount < 1) {


safl (commented):
I rarely find ``strcmp`` defendable in this kind of code, thus i am struggling to understand what it actually is that you fixed here e.g. what was the error that you were seeing and how going from ``strncmp`` to ``strcmp`` fixes it.

You write: "can treat two distinct URIs as equal when they share a common prefix of that length." Two URIs sharing a prefix up to and including `XNVME_IDENT_URI_LEN` can in fact occur, but that is exactly the point of the bound. Beyond `XNVME_IDENT_URI_LEN` we are either reading out of bounds or comparing irrelevant bytes. `strncmp` guards against that; `strcmp` does not.

The scenario the commit message describes looks like this, where `D` marks identical bytes and both strings span the full `XNVME_IDENT_URI_LEN` (currently 384):

```
g_cref_table[i].uri = 'DDDDDDDDDDDDDDDDDDDDDDDDDDDD...DDDDDDDDDDDDDDDDDDDDDDDDDDDDA'  /* 384 bytes, differs only at the final byte */
uri                 = 'DDDDDDDDDDDDDDDDDDDDDDDDDDDD...DDDDDDDDDDDDDDDDDDDDDDDDDDDDB'  /* 384 bytes */
```

So, to me this is where things start to smell... since usually uris are relatively short like `/dev/nvme0n1` or `0000:01:00.0!nsid=1`. A URI long enough to even approach `XNVME_IDENT_URI_LEN` is itself a strong indicator that something else has gone wrong. So the "false positive equal" the commit message worries about is reachable only in a regime where the input is already in a bad state.

With `strncmp(..., XNVME_IDENT_URI_LEN)` the two are treated as equal, which is the correct outcome given the buffers cannot represent more than `XNVME_IDENT_URI_LEN` bytes of meaningful URI. With `strcmp` we keep reading past the bound, which is the actual hazard, not the protection.

So i suspect there is some garbled data or some other underlying bug behind this, where switching to `strcmp` happens to mask the symptom. I do not see how `strcmp` is either simpler or "more semantically correct" than `strncmp` here. Could you elaborate on what you are seeing?

If anything in this area is worth a fix, the better target is the size discrepancy between `ident.uri` (`XNVME_IDENT_URI_LEN`) and `cref.uri` (`XNVME_IDENT_URI_LEN + 1`). That asymmetry is the actual smell.

karlowich:
@safl You opened #624 and wrote:

“ strncmp vs strcmp for URI comparison
URIs are null-terminated strings stored in fixed-size buffers. Using strncmp(..., XNVME_IDENT_URI_LEN) can treat two distinct URIs as equal if they share a common prefix of that length. Using strcmp would be simpler and semantically correct.”

As such, I wanted to address this in order to close the issue. 

I see you have changed your mind. I will close this PR and the linked issue.

safl:
Gosh... you are right. Sorry about that.

PR #693, opened 2026-06-24
Re-arrange the NVMe TCP Transport tests and tutorial

safl:
The goal of this PR is to provide a path to conveniently test and expand NVMe transport support for TCP and RDMA.

CI and the `docs/tutorial/fabrics/` tutorial used to set up NVMe targets two different ways, with no shared source to keep them in sync. This PR collapses both into cijoe scripts that drive ad-hoc developer use, CI, and the tutorial. The transport is explicit in workflow names (`test-<guest>-<transport>.yaml`). CI exercises TCP via SPDK, the only target tested in the past. The scripts also provide Linux nvmet setup, with the CI matrix entry in place but parked for follow-up. RDMA and renaming the `fabrics` device label are deferred.


cijoe/scripts/nvme_target_start.py
```diff
@@ -0,0 +1,191 @@
...
+
+    xnvme_path = Path(xnvme_repos)
+    rpc = xnvme_path / "subprojects" / "spdk" / "scripts" / "rpc.py"
+    nvmf_tgt = xnvme_path / "subprojects" / "spdk" / "build" / "bin" / "nvmf_tgt"
+    nvmf_tgt_src = xnvme_path / "subprojects" / "spdk" / "app" / "nvmf_tgt"
+
+    drivers = [
+        "modprobe nvme",
+        "modprobe nvmet",
+        "modprobe nvmet_tcp",
```
karlowich: It's not a big deal, but are these used on the SPDK path? If no, I think they should be removed so the difference between the two paths is more obvious.

karlowich (commented):
Great!
Just one very minor thing.

cijoe/scripts/nvme_target_start.py
safl: Good point... they should not be needed, on the initiator side ``nvme_tp`` is needed and i agree it is confusing that they are there. Let me clean it up.

karlowich (dismissed):
LGTM

cijoe/scripts/nvme_target_start.py
```diff
@@ -0,0 +1,191 @@
...
+        log.error("FAILED: no device labelled 'fabrics' in CIJOE config")
+        return errno.ENOENT
+
+    pcie_id = device["pcie_id"]
+    subnqn = device["subnqn"]
+
+    xnvme_path = Path(xnvme_repos)
+    rpc = xnvme_path / "subprojects" / "spdk" / "scripts" / "rpc.py"
+    nvmf_tgt = xnvme_path / "subprojects" / "spdk" / "build" / "bin" / "nvmf_tgt"
+    nvmf_tgt_src = xnvme_path / "subprojects" / "spdk" / "app" / "nvmf_tgt"
```
naddinadja: since all three use the same path to subprojects->spdk, is there a reason we do not just define an `spdk_path = Path(xnvme_repos) / "subprojects" / "spdk"` instead of `xnvme_path` ? :)

naddinadja (changes requested):
One question here :)

cijoe/scripts/nvme_target_start.py
safl: Right, that would have done the trick as well.

safl: Respun with the var-rename.

karlowich (approved):
LGTM

naddinadja (approved):
LGTM :)

PR #695, opened 2026-06-25
Probe/revert lto auto

safl:


safl:
LTO=auto ruled out as suspect; the actual trixie regression was the emitter regen silently dropping automake from debian-trixie.sh (fixed on add-ubuntu-26.04 in 9ee0f019).

PR #696, opened 2026-06-30
Multi-process mode in uPCIe backend + HOMI

naddinadja:
This is a replacement for PR #686. I have reworked my initial implementation such that the admin queue is now shared and guarded with a mutex. Also, "multi-process mode" is no longer a generic thing for xNVMe but a specific functionality within the uPCIe (and SPDK) backend.

Let me explain how it works, and sorry for this being so long:

Overall we use both flocks and shared memory regions on two different levels: for the runtime environment, which lives as long as the process (or as long as *some* devices are open), and for each device, which lives as long as the specific device is open.

To enable multi-process mode, a user must provice a `shm_id > 0` via `xnvme_opts.shm_id` or the `--shm_id` cli arguments. Then, a process will attempt to claim an flock derived from the given id: if successful, it becomes the **primary** and owner of the NVMe controllers and shared memory, and if not successful (`EWOULDBLOCK`), it becomes a **secondary**, which just attaches to shared memory. An flock is used, as the kernel manages these and releases it if the primary crashes. There is not any recovery for a crashed primary, but this ensures that a new primary can be started and clear the shared memory from any previously crashed primaries.

The primary process is responsible for initializing the uPCIe heap and hugepages, and this process then opens a shared memory region to publish the hugepage path and base pointer, which will later be used to map the admin queues in secondary processes. A secondary (that failed to claim the flock), simply maps this shared memory region and imports the hugepage from the primary. I've added a ~1 sek timeout here, since if primary and secondary are started at the same time, there might be a race on this shared memory region, so this just lets the primary finish before the secondary decides to fail :)

When the runtime environment is initialised, devices can be opened (rte_init is called at the first device startup, so the following will happen at least once, maybe more times). When an NVMe controller needs to be initialised, another flock is used to claim ownership of this if the process is the primary. This is to prevent two different primary processes (shm_id=1 and shm_id=2) from opening the same device. A shared memory region is then opened for this device, where a mutex for the admin queue is stored, along with the nvme_controller and some other necessary informations. 
Secondary processes map this shared memory region but stores a local version of the nvme_controller. Here, it updates the pointer to the admin queue using the hugepage base read from the rte shared memory. A device cannot be opened by a secondary before the primary, as it depends on the shared memory region from the primary - again a ~1 sek timeout has been added to prevent a speedy secondary from failing early because a primary was not 100% ready.
The admin queue is, as mentioned, guarded by a mutex, and a helper function for this ensures that the local and shared version of the admin queue's head/tail/phase are synchronized.

When terminating a controller, a secondary merely decreases a shared refcount and removes its local setup, but does not teardown any shared memory or the admin queue. The primary will warn when terminating a controller if the refcount is more than 1 (the last one being itself), but will not block - this is to ensure that a crashed secondary does not prevent a primary to exit, but this of course means that the state of any running secondaries will be corrupted. This is documented in the docs.
The exact same heuristic is used for the runtime environment: a secondary munmaps shared memory without unlinking shared memory and decreases the refcount; a primary will teardown the hugepages and unlink shared memory with a warning if refcount > 1.

### DIFF
--- a/docs/background/backends/upcie/mproc.md
+++ b/docs/background/backends/upcie/mproc.md
@@ -1,0 +1,204 @@
+(sec-backends-upcie-mproc)=
+# uPCIe (multi-process)
+
+The **upcie** backend supports a **multi-process** mode in which several
+independent processes share NVMe controllers. In a multi-process setup, there
+will always be one **primary** process, and any number of **secondary**
+processes. The **primary** owns the DMA hugepages and the controller hardware
+state. The **secondary** processes attach to that state and drive their own
+I/O without re-initializing the device.
+
+Multi-process mode is opt-in and enabled per device by supplying a non-zero
+**shared-memory id** via the `--shm_id` option (or the `shm_id` field of
+`struct xnvme_opts`). All processes that should share a controller must be
+launched with the **same** `shm_id`.
+
+```bash
+# Primary (first process to claim the shm_id) and secondaries use the same id
+xnvme info 0000:03:00.0 --be upcie --shm_id 1
+```
+
+(sec-backends-upcie-mproc-model)=
+## Process model
+
+### Role election
+
+Roles are decided with an advisory file lock (`flock`) on a lock file keyed by
+`shm_id` (`/tmp/xnvme-upcie-flock-<shm_id>`):
+
+- The first process to acquire the exclusive, non-blocking lock becomes the
+  **primary**. It holds the lock for its entire lifetime.
+- Any process that finds the lock already held becomes a **secondary**.
+
+### Shared memory segments
+
+Two kinds of POSIX shared-memory segments are used:
+
+- **Runtime segment** (`/xnvme-upcie-shm-<shm_id>`) — one per `shm_id`, created
+  by the primary. It carries the hugepage backing-file path and the primary's
+  hugepage virtual base address so that secondaries can import the same DMA
+  memory. A shared `refcount` keeps track of how many processes are attached.
+  Terminating a primary process while `refcount > 1` will not block, but instead
+  only warn that the state of secondary processes will be corrupted.
+- **Controller segment** (`/xnvme-upcie-<bdf>`) — one per physical controller,
+  created by the primary. It embeds the full `struct nvme_controller`, a robust
+  process-shared mutex guarding the admin queue, the bound driver name, an
+  attach `refcount`, and an `is_initialized` publication flag. Again, primary
+  processes will not be blocked from terminating even if secondary processes are
+  still attached.
+
+### Device ownership
+
+The controller segment is keyed by device BDF, independent of `shm_id`. To stop
+two primaries — for example two processes launched with different `shm_id`s —
+from both driving the same physical controller, the creating primary holds a
+second advisory lock keyed by the BDF (`/tmp/xnvme-upcie-<bdf>-lock`) for as long
+as it owns the controller. A process that cannot acquire this lock fails rather
+than creating a conflicting segment.
+
+Because the kernel releases the lock when its holder exits, the controller
+segment is self-healing across a primary crash: the next primary acquires the
+released lock, unlinks any segment the dead primary left behind, and creates a
+fresh one.
+
+### Hugepage sharing
+
+Every process allocates it own private DMA heap from its own hugepages. This
+means that I/O queue pairs and buffers are allocated from a private heap. The
+admin PRP pools are also allocated on the private heap, but the admin queue itself
+is allocated only once on the heap of the primary process. The primary
+publishes its hugepage backing-file path and virtual base into the runtime shared
+memory segment; a secondary imports that hugepage purely to reach the admin
+submission/completion rings and does not allocate from the imported mapping. The
+import lands at a kernel-chosen address, so the admin-queue pointers stored in
+the controller segment (which are primary-relative) are relocated into the
+secondary's mapping by the constant offset between the secondary's import base
+and the primary's published base. PRP entries for a secondary's own commands are
+translated against its private heap, whose physical pages it resolves itself.
+
+(sec-backends-upcie-mproc-startup)=
+## Startup handshake
+
+Secondaries may be launched concurrently with the primary, so attachment is
+guarded by two ordering barriers:
+
+1. **Runtime segment** — a secondary will wait up to ~1 second for the primary
+   to publish the hugepage information before importing hugepages.
+2. **Controller segment** — a secondary will wait up to ~1 second for the
+   per-controller segment to (a) exist, (b) be fully sized, and (c) have its
+   `is_initialized` flag published before reading any shared field.
+
+If the primary does not become ready within the timeout, attachment fails with
+`-ENOENT`.
+
+(sec-backends-upcie-mproc-admin)=
+## Admin queue serialization
+
+There is a single admin queue per controller, shared by all processes. Access
+is serialized with a **robust, process-shared** mutex (`PTHREAD_PROCESS_SHARED`
+and `PTHREAD_MUTEX_ROBUST`) held in the controller segment:
+
+- On **lock**, the holder reloads the admin queue head/tail/phase from shared
+  memory into its local controller view.
+- On **unlock**, the holder publishes its head/tail/phase back to shared memory.
+
+Synchronous admin commands and I/O queue-pair create/delete operations all run
+under this mutex. I/O queue identifiers are tracked in a shared bitmap in the
+controller segment and mutated only while the mutex is held, so each process
+creates and tears down its own I/O queue pairs without colliding.
+
+I/O **data** transfers are not serialized: each process owns its own I/O queue
+pairs and submits to them independently.
+
+(sec-backends-upcie-mproc-recovery)=
+### Crash Recovery
+
+Because the admin mutex is robust, if a process dies while holding it, the next
+locker receives `EOWNERDEAD`. In that case this process will:
+
+1. Drain any completions the dead owner may have left in the admin completion
+   queue, advancing the shared head/phase and ringing the completion-queue
+   doorbell. Since the shared head/phase are only published on unlock and
+   therefore never run ahead of the device, draining can only reap real or
+   phantom completions — never skip one.
+2. Mark the mutex consistent (`pthread_mutex_consistent`) and proceed.
+
+If the mutex has been marked unrecoverable, locking fails with `ENOTRECOVERABLE`.
+
+(sec-backends-upcie-mproc-teardown)=
+## Teardown
+
+Each process decrements the relevant `refcount` on exit. When the primary tears
+down a controller it deletes **all** I/O queue pairs still marked allocated in
+the shared bitmap before closing the controller, and then unmaps and unlinks the
+shared segments. Secondaries detach by deleting their own queue pairs, releasing
+their local request pool and BAR mapping, and unmapping the shared segments
+without unlinking or freeing DMA memory they do not own.
+
+(sec-backends-upcie-mproc-vulns)=
+## Vulnerabilities and limitations
+
+Multi-process mode is functional on the cooperative happy path but has known
+robustness and trust gaps. It is intended for trusted, same-user processes and
+should not be treated as a security boundary.
+
+### Partial crash robustness
+
+Owner-death recovery covers **only** the admin mutex and the admin completion
+queue. Other state left behind by a crashed process is **not** reclaimed:
+
+- **Leaked I/O queues and queue ids.** If a secondary crashes, the I/O queue
+  ids it allocated remain marked as in-use in the shared bitmap and the
+  corresponding device-side queues are never deleted. Over many crash/restart
+  cycles this can exhaust the queue-id space until the primary tears the
+  controller down.
+- **Stale attach refcounts.** A crashed process never decrements its
+  `refcount`. The primary's "secondaries still attached" accounting therefore
+  becomes permanently inflated after any crash. The refcount is advisory only,
+  and therefore it is not used for blocking or preventing a primary from
+  exiting.
+
+### Primary lifetime
+
+The primary owns the hugepages and the controller hardware. If the primary
+exits — cleanly or otherwise — while secondaries are still attached, those
+secondaries are left with admin-queue mapping and controller that may have been
+closed and freed. A clean primary teardown only logs a warning in this case; it
+does not prevent it. Operators must ensure the primary outlives all secondaries.
+
+### No isolation between processes
+
+All participating processes map the controller segment read/write, including the
+embedded `struct nvme_controller`. A misbehaving or compromised secondary can
+corrupt shared state and affect every other process sharing the controller.
+There is no privilege separation: any process running as the same user that can
+open the shared-memory segments can submit admin commands to the device.
+
+### Startup ordering and timeouts
+
+Attachment waits at most ~1 second at each handshake barrier. A primary that
+takes longer than that to initialize a controller will cause secondaries to time
+out with `-ENOENT`. The timeout is fixed and not currently configurable.
+
+### Compatibility with SPDK
+
+SPDK's own multi-process mode claims **all** free hugepages on the system at
+startup. Because uPCIe also needs hugepages for its DMA memory, the two cannot
+be expected to share a system, and running uPCIe and SPDK in multi-process
+mode side by side cannot be not guaranteed to work.
+
+### CUDA backend
+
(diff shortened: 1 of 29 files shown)


naddinadja:
The first two commits are also in https://github.com/xnvme/xnvme/pull/697, just to check whether they're at fault for the failing CI.

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -63,9 +68,10 @@ _rte_term(void)
  * already initialized, then it exits early.
  */
 static int
-_rte_init(size_t heap_size)
+_rte_init(struct xnvme_opts *opts)
 {
 	int err;
+	size_t heap_size = opts->host_heap_size;
```
safl, line 73: A style thing. Put ``int err`` after ``size_t heap_size = opts->host_heap_size;``

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+
+	err = flock(lock_fd, LOCK_EX | LOCK_NB);
+	if (!err) {
+		mproc->lock_fd = lock_fd;
+	} else if (errno == EWOULDBLOCK) {
+		XNVME_DEBUG("INFO: Lock file already claimed, setting role to secondary");
+		close(lock_fd);
+		lock_fd = -1;
+		mproc->is_primary = false;
+		err = 0;
```
safl: why clear ``err`` here?

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+		goto failed;
+	}
+
+	err = flock(lock_fd, LOCK_EX | LOCK_NB);
+	if (!err) {
+		mproc->lock_fd = lock_fd;
+	} else if (errno == EWOULDBLOCK) {
+		XNVME_DEBUG("INFO: Lock file already claimed, setting role to secondary");
+		close(lock_fd);
+		lock_fd = -1;
```
safl: Why set ``lock_fd`` here?

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+		mproc->is_primary = false;
+		err = 0;
+		errno = 0;
```
safl: Why clear ``errno`` here?

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	free(mproc);
+}
+
+int
+xnvme_be_upcie_mproc_rte_init(int shm_id)
+{
+	struct xnvme_be_upcie_mproc *mproc = NULL;
+	int lock_fd = -1, err;
```
safl: When you initialize ``lock_fd`` here, then the compiler cannot help you if you use it before actually assigning it. Thus, remove the initialization to ``-1``. Also, consider whether you need this at all, you are already allocating ``mproc``, so why not use it? ``mproc->lock_fd =  open(mproc->lock_name, O_CREAT | O_RDWR, 0600);``

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+	if (!mproc) {
+		err = -ENOMEM;
+		XNVME_DEBUG("FAILED: calloc(xnvme_be_upcie_mproc); err(%d)", err);
+		return err;
+	}
+
+	mproc->is_primary = true;
+
+	snprintf(mproc->lock_name, sizeof(mproc->lock_name), "/tmp/xnvme-upcie-flock-%d", shm_id);
+	mproc->lock_fd = -1;
```
safl: Why assign this ``-1`` instead of just using it and dropping that local ``lock_fd``?

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -0,0 +1,93 @@
...
+		goto failed;
+	}
+
+	g_upcie_rte.mproc = mproc;
+
+	return 0;
+
+failed:
+	if (mproc && mproc->lock_fd >= 0) {
+		flock(mproc->lock_fd, LOCK_UN);
```
safl, line 166: Something is weird about the use of ``lock_fd`` and ``mproc->lock_fd`` here as well.

safl (commented):
Just a couple of minor things to get the ball rolling.

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: Hmm, I see that I don't actually depend on err being 0 after this, so it's unnecessary. The idea was I don't count it as an error for this function that `flock()` fails because of `EWOULDBLOCK`, so therefore I set it back to 0 to not have an err=-1 hanging around for the rest of the function.

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: I think it's good practice to set an `fd` to -1 immediately after closing it, just such that error paths further down in the function does not re-close the fd.

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: This is not an error path, so to ensure that future readers of errno does not think this function failed, I clear the errno.

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: Good point :)
My thought was that I didn't have to set `mproc->lock_fd` unless the flock was successful, but I am also re-writing it back to -1, so this was unnecessary hehe

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: fixed with the your comment above

lib/upcie/xnvme_be_upcie_mproc.c
naddinadja: fixed with the your comment above

lib/upcie/xnvme_be_upcie_dev.c
naddinadja: must have been one of those things where i had squashed the fix into the wrong commit 😅

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -71,18 +98,107 @@ xnvme_be_upcie_mproc_rte_init(int shm_id)
...
+
+	oflag = mproc->is_primary ? O_CREAT | O_EXCL | O_RDWR : O_RDWR;
+	shm_mode = mproc->is_primary ? 0600 : 0;
+	shm_size = sizeof(struct xnvme_be_upcie_mproc_shm);
+
+	if (mproc->is_primary) {
+		shm_unlink(mproc->shm_name);
+	}
+
+	// Wait at most ~1 second to open shared memory segtent. We do not retry in primary
```
karlowich: > segtent

`segment`

docs/background/backends/upcie/mproc.md
```diff
@@ -0,0 +1,204 @@
...
+than creating a conflicting segment.
+
+Because the kernel releases the lock when its holder exits, the controller
+segment is self-healing across a primary crash: the next primary acquires the
+released lock, unlinks any segment the dead primary left behind, and creates a
+fresh one.
+
+### Hugepage sharing
+
+Every process allocates it own private DMA heap from its own huge pages. This
```
karlowich: Should "huge pages" be one word?

karlowich (changes requested):
Tricky code, but it looks okay to me.
I understand why you're "only" adding the CPU part for now, but we will definitely need the upcie-cuda part in a follow up PR :)

Found some typos.

docs/background/backends/upcie/mproc.md
naddinadja: yea it should hehe, woops

naddinadja:
> Tricky code, but it looks okay to me. I understand why you're "only" adding the CPU part for now, but we will definitely need the upcie-cuda part in a follow up PR :)

Yea, I think it would be somewhat OK to just add it in, but I thought I would rather just get this working at first 😅

karlowich (dismissed):
Great!

naddinadja:
When we talked last week at our meeting, it became clear to me that, despite my best efforts in describing above, it was a bit difficult to understand what was happening. Basically, I'm changing "only" the runtime initialisation and the controller initialisation to add multiprocess support (and then the termination functions as well). I didn't change the single-mode path (almost, see comment at the bottom), so here I compare the overall steps of each function to show how they differ (spoiler alert - it's not a lot).

RTE:
| Single mode | Primary | Secondary |
| --- | --- | --- |
| `hostmem_config_init()` | `hostmem_config_init()` | `hostmem_config_init()` |
| | `xnvme_be_upcie_mproc_rte_init()`<br/>which basically just claims the flock and inits/maps the shm | `xnvme_be_upcie_mproc_rte_init()`<br/>which fails to claim the flock and then maps the shm | 
| `hostmem_heap_init()` | `hostmem_heap_init()` | `hostmem_heap_init()` |
| | Share the hugepage_path/base in shm | Read the hugepage_path/base from shm and `hostmem_hugepage_import()` with the info into a separate hugepage struct |

CTRLR:
| Single mode | Primary | Secondary |
| --- | --- | --- |
| `_rte_init()` | `_rte_init()` | `_rte_init()` |
| `_pci_enable_bus_master()` | `_pci_enable_bus_master()` | `_pci_enable_bus_master()` |
| `calloc(ctrlr)` | `calloc(ctrlr)` | `calloc(ctrlr)` |
| `xnvme_be_upcie_get_driver_name()` | `xnvme_be_upcie_get_driver_name()` | |
| `calloc(ctrlr->ctrl)` | `xnvme_be_upcie_mproc_ctrlr_shm_init()`:<br/>claims flock, inits/maps the shm, inits the mutex<br/>set ctrlr->ctrl to point to shm | `calloc(ctrlr->ctrl)`<br/>copy `shm->ctrl` into `ctrlr->ctrl` |
| `nvme_controller_open()` |  `nvme_controller_open()` | Call onlythings from `nvme_controller_open()` that we need, such as mapping bar0 and allocating request pool, since we've already gotten a lot of the ctrlr info by copying in the last step |
| mutex lock (returns early bc no shm) | mutex lock | mutex lock |
| `nvme_controller_create_io_qpair(ctrlr->sync)` | `nvme_controller_create_io_qpair(ctrlr->sync)` | `nvme_controller_create_io_qpair(ctrlr->sync)` |
| mutex unlock (returns early bc no shm) | mutex unlock | mutex unlock |


Beside this, I've added a mutex to the admin queue, which is used when submitting admin commands and also added a helper function for creating queues in multi-process mode, which makes sure to use this mutex and update local versions of the admin queue and claimed queue ids - this is used in the public facing queue API.

_Note on single mode path:_
In the initialisation of the controller, the order of `xnvme_be_upcie_get_driver_name()` and `calloc(ctrlr->ctrl)` is reversed (used to be calloc then driver_name), but since they have no effect on each other on the single mode path, I've swapped them, just like the C compiler might have done. :)

naddinadja:
@safl and @karlowich , I have now added HOMI as a tool in xNVMe for acting as a small primary process for the given devices. I used the CLI setup that xnvmeperf also uses, but realised that with this (regular arguments, and a list of dev URIs), we cannot specify namespace IDs for the given devices. Maybe this is a problem for later, since it also needs to be fixed in xnvmeperf.
Also, please help me evaluate whether the CLI arguments are right as of now for HOMI. Right now it *only* has the backend and shm_id, which might be too few. What else could I need?

Also, I changed the test commit here to use HOMI as a background primary instead of the dummy-primary that I initially had.

docs/background/backends/upcie/mproc.md
```diff
@@ -0,0 +1,204 @@
...
+
+The memory model is not the obstacle: `upcie-cuda` keeps all NVMe control
+structures — the admin queue, the I/O submission/completion rings, PRP lists
+and request pools — on the same host hugepage heap that multi-process mode
+shares, and places only data buffers in per-process GPU memory, which is never
+shared between processes. What is missing is serialization: the `upcie-cuda`
+admin path submits to the shared admin queue without taking the per-controller
+admin mutex, so concurrent processes would race on the admin submission ring.
+Enabling multi-process mode for `upcie-cuda` would require serializing its admin
+path under the same mutex used by the host backend.
```
karlowich, line 208: We should add this in a follow-up PR.

tools/homi.c
```diff
@@ -0,0 +1,168 @@
...
+		"Open the given devices and hold them open",
+		"Open the given devices and hold them open",
+		sub_start,
+		{
+			{XNVME_CLI_OPT_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_URI, XNVME_CLI_POSN},
+			{XNVME_CLI_OPT_NON_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_SHM_ID, XNVME_CLI_LREQ},
+			{XNVME_CLI_OPT_ORCH_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_BE, XNVME_CLI_LOPT},
```
karlowich, line 164: I think we want the opts for sizing the uPCIe heap as well, since that will be controlled by the first open.

tools/homi.c
```diff
@@ -0,0 +1,168 @@
...
+		},
+	},
+};
+
+static struct xnvme_cli g_cli = {
+	.title = "homi - Host-Orchestrated Multi-process I/O",
+	.descr_short = "Hold NVMe devices open for multi-process sharing",
+	.descr_long = "Hold NVMe devices open for multi-process sharin. Secondary "
```
karlowich: missing a `g` in "sharing"

karlowich (changes requested):
Looks good, just very few comments.

karlowich:
Oh, one more thing. We probably want a doc page for the HOMI tool.

tools/homi.c
naddinadja: added with a new commit inserted before adding homi itself

tools/homi.c
naddinadja: fixed

docs/background/backends/upcie/mproc.md
naddinadja: for now, i've created this issue: https://github.com/xnvme/xnvme/issues/728

naddinadja:
> Oh, one more thing. We probably want a doc page for the HOMI tool.

Yes, of course - this has been added as well now. 
I've fought a bit with the tests throughout the day, but it should be good now - and the --host_heap_size help me. :)

safl (commented):
A bit of high-level feedback/questions.

Is it possible to test the homi with the SPDK backend as well? It ought to work using the same parameters (except for the backend name that is).

Also, supporting upcie CUDA/HIP is quint essential, is that doable here? Perhaps there are design decisions that alter what needs to be done for their usecase that might be preferable to get right the first time around? Or is there a natural succession for a follow-up PR?

naddinadja:
> Is it possible to test the homi with the SPDK backend as well? It ought to work using the same parameters (except for the backend name that is).

Yes, this is already happening, but only in the iommu-disabled case - I'll add extra cases for iommu-enabled and freebsd. Since it has to run all tests again in multi-process mode, it will take a lot longer to run the tests, so I think we could benefit from looking into parallellising these in some way. :)
 
> Also, supporting upcie CUDA/HIP is quint essential, is that doable here? Perhaps there are design decisions that alter what needs to be done for their usecase that might be preferable to get right the first time around? Or is there a natural succession for a follow-up PR?

I'll look into how much extra work it is, and try to either add it here or as another PR stacked on this. Didn't want this to become a massive PR (since we've already established that it is a *lot*), so that's why I didn't include it in the first round

karlowich (dismissed):
LGTM

naddinadja:
@safl I've added a final commit with tests for mproc/homi for only spdk (when upcie is not available: freebsd and iommu enabled). I found that important to include here to verify that homi works as intended. As for upcie-cuda and upcie-hip implementations, I have that ready on my machine, but I want to add it in another PR just to ensure that this doesn't blow up. However, it's not possible to create github PR stacks from a forked repository, so I will wait with that until we've merged this one.

safl (approved):
lgtm

PR #702, opened 2026-07-01
build(rdma): add meson build entries to support rdma in spdk

jswaro:
Addresses a gap in the existing meson build infrastructure around RDMA functionality in the downstream spdk build. 

The problem is that the libraries were never linked against the downstream libibverbs/rdmacm libraries, and spdk is built without rdma support by default.


karlowich (dismissed):
Great!

safl (commented):
@jswaro the info below is done by a coding agent, just to provide some input while I am away from keyboard. I think it might be on to something that is worth double-checking (the package-names, workflow-order and the semantics on the meson-feature/option.

Structural review. The design (yielding subproject option, exported `spdk_rdma_enabled` variable, conditional `--with-rdma` configure flag, deps.yaml regeneration) is sound. Three items look blocking.

First, `should_have_rdma = get_option('with-spdk-rdma').enabled() and is_linux` in `subprojects/packagefiles/spdk/meson.build` collapses a tri-state feature to a bool. `.enabled()` is true only for the literal `enabled` value, false for both `auto` and `disabled`. Because `should_have_rdma` only feeds `required:`, `cc.find_library` still probes on `-Dwith-spdk-rdma=disabled` and, if the libs happen to be installed on the host, `rdma_support_enabled` becomes true and SPDK is configured with `--with-rdma`. The same defect also defeats the `is_linux` gate: the subproject builds on FreeBSD, and libs-present there means RDMA on. The idiomatic fix given the project's `meson_version: '>=0.59'` is to pass the feature object straight through, `required: get_option('with-spdk-rdma')`, then compute `rdma_support_enabled = is_linux and ibverbs_dep.found() and rdmacm_dep.found()` so the platform gate is enforced separately.

Second, the new "Container-preparation, install build-requirements" step in `.github/workflows/verify.yml` runs before the "Retrieve and unpack the xNVMe source archive" step (line 265 vs 272). The workspace at that point contains no `toolbox/pkgs/*.sh`, so the `[[ -f \"${BSCRIPT}\" ]]` guard is always false and the step is a silent no-op for every matrix entry. As it stands CI never installs the new packages and never actually links an RDMA-enabled build; the green matrix does not certify what the PR claims. Moving the step below `setup-source` should be enough.

Third, `deps.yaml` uses `libibverbs` / `librdmacm` as both the `default:` and the `emerge:` names, which is wrong for three distros. Alpine and Arch do not have `libibverbs` or `librdmacm` packages: `pkgs.alpinelinux.org/package/edge/community/x86_64/libibverbs` returns 404 and `archlinux.org/packages/search/?name=libibverbs` returns zero hits. The correct package is `rdma-core-dev` on Alpine and `rdma-core` on Arch (extra). Gentoo's `dev-libs/libibverbs` is also stale, current package is `sys-cluster/rdma-core`. Once the workflow-step order is fixed and CI actually runs the installs, alpine/archlinux/gentoo container prep will hard-fail. Adding an `apk:` override in `deps.yaml`, correcting `emerge:` to `sys-cluster/rdma-core`, and switching the `default:` (or listing multiple names) should cover it.

Runtime-path degradation looks clean when RDMA is compiled out: `conf_data.set('XNVME_BE_SPDK_TRANSPORT_RDMA_ENABLED', false)` emits `#undef` in the generated header and the `#ifdef` guard in `lib/xnvme_be_spdk_dev.c` drops the transport from `g_xnvme_be_spdk_transport`, so an RDMA URI ends in ENXIO at probe rather than a link failure.

A few small hygiene items alongside those: `subprojects/packagefiles/spdk/meson.build` is missing its trailing newline; `spdk_configure*.sh` should quote \`\"\$@\"\` rather than \`\$@\`; the workflow's `source \${BSCRIPT}` should be quoted `source \"\${BSCRIPT}\"` to match the guard above it.

subprojects/packagefiles/spdk/meson.build
```diff
@@ -60,15 +60,35 @@ archive_dep = cc.find_library(
   has_headers: ['archive.h'],
 )
 
+should_have_rdma = get_option('with-spdk-rdma').enabled() and is_linux
```
safl: `.enabled()` collapses the tri-state to a bool: true only for the literal `enabled`, false for `auto` (the default) and for `disabled`. Because this only feeds `required:`, `cc.find_library` still probes on `-Dwith-spdk-rdma=disabled`, and if the libs happen to be on the host the resulting `rdma_support_enabled` is true and SPDK is configured with `--with-rdma`. Same defect defeats the `is_linux` gate on the FreeBSD path. Idiomatic fix given `meson_version >= 0.59`: pass the feature straight through as `required: get_option('with-spdk-rdma')`, then compute `rdma_support_enabled = is_linux and ibverbs_dep.found() and rdmacm_dep.found()` separately.

.github/workflows/verify.yml
```diff
@@ -262,6 +262,12 @@ jobs:
       if: contains(matrix.container.os, 'opensuse')
       run: zypper --non-interactive --no-refresh install -y tar gzip
 
+    - name: Container-preparation, install build-requirements
```
safl: This step runs before `Retrieve and unpack the xNVMe source archive` at line 273. At this point `toolbox/pkgs/*.sh` does not yet exist in the workspace, so `[[ -f "${BSCRIPT}" ]]` is always false and the step is a silent no-op for every matrix entry. CI never actually installs the new packages, so the green matrix does not certify what the PR claims. Move the block below `setup-source`.

.github/workflows/verify.yml
```diff
@@ -262,6 +262,12 @@ jobs:
...
+      env:
+        BSCRIPT: toolbox/pkgs/${{ matrix.container.os }}-${{ matrix.container.ver }}.sh
+      run: |
+        if [[ -f "${BSCRIPT}" ]]; then source ${BSCRIPT}; fi
```
safl: Quote to match the guard above: `source "${BSCRIPT}"`.

safl (commented):
Same set as the earlier prose review, anchored inline this time since the branch was rebased without addressing any of it. Three items I would treat as blocking (meson feature semantics, workflow step order, alpine/arch/gentoo package names) and three hygiene nits alongside.

toolbox/pkgs/alpine-latest.sh
```diff
@@ -18,6 +18,8 @@ apk add \
  git \
  libaio-dev \
  libarchive-dev \
+ libibverbs \
```
safl: `libibverbs` and `librdmacm` are not standalone packages in Alpine. The RDMA stack ships as `rdma-core-dev` (edge/community), which provides both `libibverbs` and `librdmacm` with headers. Worth confirming against `pkgs.alpinelinux.org` before merging. Same for the `default:` and `apk:` entries in `toolbox/pkgs/emitter/deps.yaml`.

toolbox/pkgs/archlinux-latest.sh
```diff
@@ -16,6 +16,8 @@ pacman -S --noconfirm \
  git \
  libaio \
  libarchive \
+ libibverbs \
```
safl: Not standalone packages in Arch. `pacman -S rdma-core` (extra) provides both `libibverbs.so` and `librdmacm.so` with headers. `deps.yaml` needs a matching adjustment.

toolbox/pkgs/gentoo-latest.sh
```diff
@@ -17,6 +17,8 @@ emerge \
  dev-lang/nasm \
  dev-libs/isa-l \
  dev-libs/libaio \
+ dev-libs/libibverbs \
```
safl: `dev-libs/libibverbs` and `dev-libs/librdmacm` are stale in the Gentoo tree; current package is `sys-cluster/rdma-core`. `deps.yaml` `emerge:` list needs the same fix.

subprojects/packagefiles/spdk/spdk_configure.sh
```diff
@@ -4,6 +4,7 @@
   --disable-unit-tests \
   --disable-examples \
   --disable-apps \
+  $@ \
```
safl: Use `"$@"` rather than `$@` so a passed flag containing whitespace would survive. Same in `spdk_configure_debug.sh`.

subprojects/packagefiles/spdk/meson.build
```diff
@@ -177,3 +202,5 @@ spdk_dep = declare_dependency(
   link_args: spdk_link_args,
   include_directories: spdk_inc,
 )
+
+spdk_rdma_enabled = rdma_support_enabled
```
safl: Missing trailing newline.

jswaro:
I've applied the feedback. I'm re-running the checks to ensure it works within the github runners.

safl:
I switched this to draft and removed the ``maas`` label as it looks like it is currently in a state of "work-in-progress". By doing so the CI is not pulling in runs on the scarce HW resources.

PR #703, opened 2026-07-01
ci: guest-provisioning and RHEL9-family build fixes

safl:
Three+1 independent CI and guest-provisioning fixes, split out so they stop re-running unrelated matrices.

First, pull python3.11 on the RHEL9-family containers, centos and rocky ship Python 3.9, which cijoe 0.9.60 no longer supports. Second, stage guest images via cijoe's system_imaging.diskimage_from_oras, which skips the redundant 3 GB pull when the pinned digest is already staged. Third, archive the QEMU serial log on guest-start failure so the boot output is available for diagnosis.


karlowich (approved):
LGTM

PR #705, opened 2026-07-03
chore(upcie): vendor uPCIe v0.5.1

karlowich:
This is the only change: https://github.com/safl/upcie/pull/28


safl (approved):
L(VERY)GTM

safl:
Looks like yet-another-spurious-error in the FreeBSD latency-test. Just give it at kick once the others settle.

PR #706, opened 2026-07-03
feat: add gpu_id option to select the GPU for upcie-cuda/upcie-hip

karlowich:
The `upcie-cuda` and `upcie-hip` P2P backends previously hardcoded GPU
ordinal 0 (`cuDeviceGet(&cu_dev, 0)` / `hipSetDevice(0)`), so there was no
way to target a specific GPU on a multi-GPU host. This adds a `gpu_id`
option that flows from `xnvme_opts` through the CLI framework down into
both backends.

The default is `0`, so existing behavior is unchanged.

### Naming: why `gpu_id` and not `device_id`

At first I wanted to name it `device_id`, since it plays nicely with the
existing `device_heap_size` option (both use "device" to mean the GPU).
However, I decided to avoid it because "device" is heavily overloaded in xNVMe. 
It almost always refers to the NVMe device (`struct xnvme_dev`, `--dev`, `dev_nsid`,
etc.), so `device_id` would be easy to misread as an NVMe device selector.

### DIFF
--- a/include/libxnvme_cli.h
+++ b/include/libxnvme_cli.h
@@ -122,6 +122,8 @@ struct xnvme_cli_args {
 	const char *core_mask;
 	const char *iova_mode;
 
+	uint32_t gpu_id; ///< upcie-cuda/upcie-hip: GPU ordinal to use
+
 	struct xnvme_opts_css css; ///< SPDK controller-setup: do command-set-selection
 
 	uint32_t use_cmb_sqs;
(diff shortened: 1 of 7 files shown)


safl (commented):
What happens when there are an NVIDIA and an AMD GPU in the system?

safl:
> What happens when there are an NVIDIA and an AMD GPU in the system?

nevermind... the id is local to the runtime...

PR #708, opened 2026-07-07
feat(libxnvme_cuda): embed device helpers, drop upcie header dependency

### LINKED ISSUE #674: `libxnvme_cuda.h` has untracked dependency on uPCIe headers
`libxnvme_cuda.h` is a public installed header that conditionally includes uPCIe internal
  headers:

  ```c
  #ifdef __CUDACC__
  #include <upcie/nvme/nvme_command.h>
  #include <upcie/nvme/nvme_qpair_cuda.h>
```

  This creates a hard build-time dependency on uPCIe headers for any downstream project that
  compiles CUDA translation units (`.cu` files) against xNVMe. However:

  1. xNVMe does not install the uPCIe headers it depends on.
  2. uPCIe is not declared as a `Requires: dependency` in xNVMe's generated `.pc` file, so
  downstream `pkg-config` consumers get no hint that uPCIe must be present.
  3. The uPCIe headers are vendored under `toolbox/third-party/linux/upcie/` but are not
  installed as part of meson install.

  After installing xNVMe, any downstream project that includes `libxnvme.h` from a `.cu` file
  gets:

  `fatal error: upcie/nvme/nvme_command.h: No such file or directory`

  Installing uPCIe separately won't help, because uPCIe's own `meson.build` uses a
  single `install_headers` call with `subdir: 'upcie'` for all headers, including those under
  `nvme/`, which flattens the hierarchy and installs them as `upcie/nvme_command.h` instead
  of `upcie/nvme/nvme_command.h`.

  To resolve this we should do, as a minimum, one of:

  - Install the vendored uPCIe headers that `libxnvme_cuda.h` depends on as part of
  meson install, and add `Requires: upcie` (or equivalent include-path machinery) to
  the xNVMe `.pc` file.
  - Or explicitly document uPCIe as a required build-time dependency with a note on the
  expected installed header layout, so downstream projects know what to provide.

  The uPCIe `meson.build` flattening issue (`subdir: 'upcie'` stripping the `nvme/`
  subdirectory) is a separate upstream uPCIe bug, but it compounds the problem.

karlowich:
The public libxnvme_cuda.h pulled in <upcie/nvme/nvme_command.h> and <upcie/nvme/nvme_qpair_cuda.h>, which xNVMe vendors but does not install. Downstream projects compiling .cu files against installed xNVMe therefore failed to find these headers (issue #674).

Embed the device-side queue definition and IO helpers directly into libxnvme_cuda.h and retype them onto xNVMe spec types, so the public header no longer depends on any upcie header:

  struct nvme_qpair_cuda  -> struct xnvme_cuda_queue (flattened)
  struct nvme_command     -> struct xnvme_spec_cmd
  struct nvme_completion  -> struct xnvme_spec_cpl
  nvme_qpair_cuda_*        -> xnvme_cuda_*

Using the spec types also drops the manual phase/status bit-twiddling in favour of the xnvme_spec_status bitfields (status.p, status.sc).

The flattened xnvme_cuda_queue keeps the exact layout of the upcie struct nvme_qpair_cuda, so the internal host backend is unchanged.

Closes #674

### DIFF
--- a/include/libxnvme_cuda.h
+++ b/include/libxnvme_cuda.h
@@ -20,8 +20,6 @@
 
 #ifdef __CUDACC__
 #include <errno.h>
-#include <upcie/nvme/nvme_command.h>
-#include <upcie/nvme/nvme_qpair_cuda.h>
 
 /**
  * GPU-resident NVMe queue pair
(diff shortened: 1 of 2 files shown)


safl (approved):
LGTM

naddinadja (approved):
LGTM :)

PR #711, opened 2026-07-09
fix(upcie/dev): add _rte_term() to error paths in ctrlr_init()

naddinadja:
If the runtime environment has not been initialised when calling xnvme_be_upcie_ctrlr_init(), it will be initialised by a call to _rte_init(), however, _rte_term() was not called on error paths after the initialisation, meaning the global rte struct would not be torn down correctly. For simplification, I also add a "failed" label at the end of the function.


karlowich (approved):
LGTM

safl (approved):
LGTM

PR #714, opened 2026-07-13
doc: use @file instead of @headerfile in header doc-blocks

### LINKED ISSUE #713: Header docstring inserted into docstring of first function
The docstring of a given header shows up in the docstring of the first function in the header when looking at the xNVMe documentation at [xnvme.io](https://xnvme.io/api/c/).
For [example](https://xnvme.io/api/c/nvme/xnvme_nvm.html#xnvme-nvm-read), the following shows up in the docstring for `xnvme_nvm_read`:

"SPDX-FileCopyrightText: Samsung Electronics Co., Ltd.

SPDX-License-Identifier: BSD-3-Clause Submit, and optionally wait for completion of, a NVMe Read"

karlowich:
The file-level doc-block in each public header used the Doxygen command @headerfile, which is meant for documenting which header a class is declared in - not for documenting a file. As a result Doxygen did not treat the block as file documentation and instead attached it to the first declaration in the file, leaking the SPDX license header into the docstring of e.g. xnvme_nvm_read.

Switch these blocks to @file, matching libxnvme.h which already did so. While here, correct the file names on the blocks that had been copy-pasted with the wrong name (opts/queue referenced libxnvme_adm.h, kvs referenced libxnvme_kv.h).

Closes #713

### DIFF
--- a/include/libxnvme_adm.h
+++ b/include/libxnvme_adm.h
@@ -3,7 +3,7 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  *
- * @headerfile libxnvme_adm.h
+ * @file libxnvme_adm.h
  */
 
 /**
--- a/include/libxnvme_be.h
+++ b/include/libxnvme_be.h
@@ -3,7 +3,7 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  *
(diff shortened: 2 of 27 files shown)


naddinadja (approved):
great

PR #716, opened 2026-07-13
fix(xnvme_platform_linux): guard snprintf truncation in namespace scan

karlowich:
Check the snprintf return value when building the per-namespace sysfs path in _scan_pci_report_namespaces() and skip the entry on truncation or encoding error. This silences a -Wformat-truncation warning emitted by the debug build, where the optimizer cannot prove the composed path fits in ns_dir.


naddinadja (approved):
great

PR #717, opened 2026-07-13
 Make FreeBSD contigmem buffer size configurable

karlowich:
Loading `contigmem` intermittently failed on the FreeBSD CI with
`contigmalloc failed for buffer N` / `ENOMEM`, even though the VM had gigabytes
free. `contigmalloc(9)` cannot relocate wired pages, and the ~2GB of wired
DMA/kernel memory scattered across the heap (11 emulated NVMe controllers plus
the IOMMU) leaves too few regions large enough for a 256MB buffer.

This makes the size configurable so CI can use small buffers while other setups keep the larger default.

### DIFF
--- a/cijoe/tests/conftest.py
+++ b/cijoe/tests/conftest.py
@@ -207,7 +207,13 @@ class XnvmeDriver(object):
     def kernel_detach(cijoe):
         """Detach from kernel"""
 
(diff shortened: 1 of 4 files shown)


naddinadja (approved):
So nice that you got this fixed! :)

PR #719, opened 2026-07-14
fix(tests/map): unmap buffers and re-enable mem_map_unmap on Trixie

### LINKED ISSUE #630: Fix hack "disable libvfn mem_map_unmap test on Debian Trixie"
Improvements to the developer tooling, ci, and the codebases done on:
https://github.com/xnvme/xnvme/pull/629

Introduced a regression for which the hack: "disable libvfn mem_map_unmap test on Debian Trixie" is applied.
This is a note that it should be fixed properly.

karlowich:
Fixes #630 by removing the hack that disabled `test_mem_map_unmap` on
Debian Trixie and addressing the underlying resource leak in the test.

Closes #630


naddinadja (approved):
good find and silly we didn't notice before :)

PR #720, opened 2026-07-15
test(kvs): fix test_list under NumPy 2.x and re-enable it on Debian Trixie

### LINKED ISSUE #631: fix hack "disable Python KV test_list on Debian Trixie"
Improvements to the developer tooling, ci, and the codebases done on:
https://github.com/xnvme/xnvme/pull/629

Introduced a regression for which the hack: "disable Python KV test_list on Debian Trixie" is applied.
This is a note that it should be fixed properly.

karlowich:
Closes #631

## Summary

`test_list` in `python/tests/test_key_value.py` was disabled on the Debian Trixie
PCIe workflows (in #629) because it failed there. This fixes the actual cause —
a NumPy 2.x incompatibility — and re-enables the test.

## Root cause

`unpack_list_data()` decodes the KV List response by converting one-element
NumPy arrays to Python ints, e.g. `int(buf[:4].view(np.uint32))`. Debian Trixie
ships **NumPy 2.x** (with Python 3.12), which no longer converts a one-element
(1-D) array to a scalar implicitly and raises:

TypeError: only 0-dimensional arrays can be converted to Python scalars

NumPy 1.x accepted this, so it only shows up on Trixie. It is **backend-
independent** — the KV device and List command work correctly (the response
header decodes to 260 keys) — which is why it failed both before and after the
switch to `vfu_kvssd`, and matches the "Python 3.12+" environmental change noted
when the test was disabled.

## Fix

Index the single element before `int()`, e.g.
`int(buf[:4].view(np.uint32)[0])`, in the two conversions in
`unpack_list_data()`. This works on both NumPy 1.x and 2.x. No other test logic
changes.

### DIFF
--- a/cijoe/workflows/test-debian-trixie-iommu_disabled-pcie.yaml
+++ b/cijoe/workflows/test-debian-trixie-iommu_disabled-pcie.yaml
@@ -19,7 +19,7 @@ steps:
   with:
     # Mirror trixie's python tests minus libvfn (needs vfio-pci).
     args: '{{ config.xnvme.repository.sync.remote_path }}/python/tests --timeout=60 --timeout-method=thread -k "not fabrics
-      and not libvfn and not test_list"'
+      and not libvfn"'
     run_local: false
     random_order: false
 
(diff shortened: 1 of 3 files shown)


naddinadja (approved):
amazing

PR #661, opened 2026-04-22
upcie: add vfio-pci path

jaeyoon-choi:
## Summary
 Add VFIO support to the upcie backend

 ## Changes
 - sync the bundled upcie headers to v0.4.2 with VFIO controller support
 - select the VFIO-backed open/close path when the device is bound to vfio-pci
 - reject vfio-pci for upcie-cuda, which still expects the non-VFIO path

PR #721, opened 2026-07-15
test(cijoe): enable uPCIe in IOMMU-enabled verify workflow

### LINKED ISSUE #669: Add uPCIe to verify CI workflow.
uPCIe has support for vfio since #661, thus it should be added to the verify-debian workflow.

karlowich:
uPCIe gained vfio-pci support in #661. Under IOMMU the pcie-labelled devices are already bound to vfio-pci for the libvfn/spdk tests, so drop the "not upcie" exclusion to let the uPCIe backend run there too.

Closes #669


naddinadja (approved):
Given that I assume the tests will pass, this is great :)

PR #723, opened 2026-08-04
fix(readme): use correct URL for CI badge

naddinadja:
For a long time, we've only had a single Github Actions workflow running, so the badges in the readme has been broken. This should correctly reflect the state of our CI with the badge.

Before:
image

After:
image


karlowich (approved):
Cool

PR #724, opened 2026-08-05
fix(be/linux/libaio): keep iocb in local storage, preserve ctx->cmd

karlowich:
The `libaio` interface aliased `struct iocb` onto `&ctx->cmd` and let `io_prep_*()/io_submit()` overwrite it, clobbering the NVMe command. Unlike the `io_uring` interfaces, which build their SQE in the ring's own memory and leave `ctx->cmd` intact, this left the command unusable after submission.

The `iocb` is only read by the kernel during `io_submit()`, so hold it in a local variable instead. `ctx->cmd` now survives until completion, matching the other backends, so callers can rely on it (e.g. to track SLBA across re-submissions).

This is needed to make #591 work.


safl (approved):
lgtm.

naddinadja (approved):
niiice :)

PR #725, opened 2026-08-10
Generic CPU id CLI argument parsing (and added `--cpulist` argument)

### LINKED ISSUE #654: Extend xnvmeperf cli with "--cpu-list" argument
Tools like taskset and bdevperf allow users to define the CPU mask as a comma-separated list of CPUs or CPU ranges, e.g. `0,1-3,5,8,13-20`. The two tools handle it slightly differently:

- Taskset parses the given mask (positional argument) as a list if the flag `--cpu-list`(or `-c`) is also set, and the convention is to put this before the positional argument, e.g. `taskset --cpu-list 0-2,6 <command>`
- Bdevperf figures out itself whether the given argument is a CPU list or a hex-mask, but expects a list to be wrapped in brackets, e.g. `bdevperf -m [0-2,6]`

I suggest we add a similar feature to xnvmeperf to allow the user to use either a hex mask or a CPU list. I like the taskset approach the best, as it does not have brackets, and there is no runtime guessing of whether it's a list or hex mask. However, since our CPU mask is already a names argument, it would require either: 

1. having two separate, optional arguments where either or must be given, so `xnvmeperf --cpu-mask 0x3` or `xnvmeperf --cpu-list 0,1`

    or

2. that we keep the `--cpu-mask` argument required and have an extra flag if it's a list, e.g. `xnvmeperf --cpu-mask 0x3` or `xnvmeperf --cpu-mask 0,1 --cpu-list` (which is kinda ugly)

The problem with guessing at runtime is specifically the case where you run `xnvmeperf --cpu-mask 6`, for example. Would this mean 0x6 = 0110b, meaning we should use CPU 1 and 2, or would it mean to use CPU 6? Bdevperf using brackets for lists fixes this problem, but I think that's kinda ugly too

**TL;DR**: It would be nice to have some way of giving a CPU list instead of a CPU mask. Other tools offer this, but in slightly different ways, so we need to find a way that we like.

naddinadja:
This PR moves the parsing of the `--cpumask` to the xNVMe CLI module, such that the parsing of input of the CLI argument does not need to be rewritten if ever reused. Also, we add a `--cpulist` argument, which allows the user to provide a list of CPU ids instead of a hexmask. A such list has a combination of specific ids and id ranges, e.g. "0,4-7", which would parse to `[0,4,5,6,7]`.

This closes #654 .

### DIFF
--- a/lib/xnvme_cli.c
+++ b/lib/xnvme_cli.c
@@ -1141,6 +1141,68 @@ xnvme_cli_usage(struct xnvme_cli *cli)
 	}
 }
 
+/**
+ * Parse a CPU hex mask into a list of CPU ids and save it in args->cpus
+ */
+static int
+parse_cpumask(struct xnvme_cli_args *args, const char *hex)
+{
+	int len, count = 0, err;
+	bool nonzero = false;
+
+	len = strlen(hex);
+
+	if (len > 1 && (!memcmp(hex, "0x", 2) || !memcmp(hex, "0X", 2))) {
+		hex += 2;
+		len -= 2;
+	}
+
+	if (len == 0) {
+		err = -EINVAL;
+		xnvme_cli_perr("Error: --cpumask must be a non-zero hex value", err);
+		return err;
+	}
+
+	if (len > XNVME_CLI_CPU_MAX_ID / 4) {
+		err = -EINVAL;
+		xnvme_cli_perr("Error: --cpumask cannot define a CPU id > 1023", err);
+		return err;
+	}
+
+	for (int i = 0; i < len; i++) {
+		char c = hex[i];
+		if (!((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') ||
+		      (c >= 'A' && c <= 'F'))) {
+			err = -EINVAL;
+			xnvme_cli_perr("Error: --cpumask must be a non-zero hex value", err);
+			return err;
+		}
+		if (c != '0') {
+			nonzero = true;
+		}
+	}
+	if (!nonzero) {
+		err = -EINVAL;
+		xnvme_cli_perr("Error: --cpumask must be a non-zero hex value", err);
+		return err;
+	}
+
+	// Extract CPU indices; rightmost digit = lowest bits
+	count = 0;
+	for (int i = len - 1; i >= 0; i--) {
+		int value = xnvme_hex_digit_val(hex[i]);
+		int bit_offset = (len - 1 - i) * 4;
+		for (int b = 0; b < 4; b++) {
+			if ((value >> b) & 1) {
+				args->cpus[count++] = bit_offset + b;
+			}
+		}
+	}
+
+	args->ncpus = count;
+	return 0;
+}
+
 int
 xnvme_cli_assign_arg(struct xnvme_cli *cli, struct xnvme_cli_opt_attr *opt_attr, char *arg,
 		     enum xnvme_cli_opt_type opt_type)
@@ -1148,6 +1210,7 @@ xnvme_cli_assign_arg(struct xnvme_cli *cli, struct xnvme_cli_opt_attr *opt_attr,
 	struct xnvme_cli_args *args = &cli->args;
 	char *endptr = NULL;
 	uint64_t num = 0;
+	int err;
 
 	// Check numerical args
 	if (arg && (opt_type != XNVME_CLI_LFLG)) {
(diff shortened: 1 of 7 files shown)


lib/xnvme_cli.c
```diff
@@ -1142,13 +1142,89 @@ xnvme_cli_usage(struct xnvme_cli *cli)
 	}
 }
 
+// Returns the integer value of a hexadecimal digit character.
+static int
+hex_digit_val(char c)
```
karlowich: Maybe this is generic enough to live in `libxnvme_util.h`?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -518,7 +437,7 @@ print_run_args(struct xnvmeperf_args *args, const char *pattern)
 			if (i) {
 				printf(", ");
 			}
-			printf("%d", args->cpus[i]);
+			printf("%d", (int)args->cpus[i]);
```
safl: Instead of casting, how about using the ``PRIu16`` format-chat?

lib/xnvme_cli.c
```diff
@@ -1595,7 +1671,15 @@ xnvme_cli_assign_arg(struct xnvme_cli *cli, struct xnvme_cli_opt_attr *opt_attr,
...
 		break;
 	case XNVME_CLI_OPT_CPUMASK:
-		args->cpumask = arg ? arg : "INVALID_INPUT";
+		args->cpumask = arg;
+
+		if (args->cpumask) {
+			err = parse_cpumask(args, args->cpumask);
+			if (err) {
+				errno = -err;
+				return -1;
```
safl, line 1779: I might just be mis-remembering, but instead of ``return -1``, should this function not return either ``errno`` or negative ``errno``?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -760,7 +679,7 @@ xnvmeperf_run(struct xnvmeperf_args *args)
 		int start = i * base + (i < extra ? i : extra);
 		int count = base + (i < extra ? 1 : 0);
 
-		threads[i].cpu = args->cpus[i];
+		threads[i].cpu = (int)args->cpus[i];
```
safl: The int-cast here as well, why is that? Perhaps ``threads[N].cpu`` should be ``uint16_t`` instead of casting it?

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -1245,6 +1163,9 @@ parse_run_args(struct xnvme_cli *cli, struct xnvmeperf_args *args)
 
 	args->nqueues = cli->args.nqueues ? cli->args.nqueues : 1;
 
+	args->ncpus = (int)cli->args.ncpus;
```
safl: The int-cast here as well, why is that? Perhaps ``threads[N].cpu`` should be ``uint16_t`` instead of casting it?

safl (commented):
A couple of things inline.

karlowich (dismissed):
Looks good. Just one comment, it's not a blocker.

lib/xnvme_cli.c
naddinadja: yea good point, I've moved it and renamed it to `xnvme_hex_digit_val()`

lib/xnvme_cli.c
naddinadja: None of the other cases in the switch have calls to external functions, so they don't have an error from *somewhere else* to return, and therefore they just set errno and return -1 - and I just copied that setup. But the function's return value is only ever used in a `if (xnvme_cli_assign_arg(...)) { ..debug..; return -1; }`, so idk if it's necessary to be specific with the return value, since it's just passed on as -1 anyways?

tools/xnvmeperf/xnvmeperf.c
naddinadja: fixed! :)

tools/xnvmeperf/xnvmeperf.c
naddinadja: fixed! :)

tools/xnvmeperf/xnvmeperf.c
naddinadja: fixed! :)

lib/xnvme_cli.c
safl: It is fine then. Eventually, i guess we could refactor this, just for the sake of spreading the negative-errno return convention. But nothing to act on here/now.

safl (approved):
lgtm

karlowich (approved):
lgtm

PR #726, opened 2026-08-11
fix(cli): update argument descriptions

naddinadja:
Some arguments were either outdated or wrong:
- `uri`: Fixed wrong point(.) with colon(:) like SYS_URI description.
- `uuid`: Added "index" to description, since this value is given to `xnvme_spec_cmd_idfy`'s cdw14 (uuid in our code), which has doc-string "UUID index", and is described as "UUID Index" in the base NVMe spec.
- `rae`: Fixed from "reset" to "retain" as in the NVMe spec and rest of our code base.
- `ause` and `sanact`: Fixed sanatize -> sanitize spelling error.
- `zsa`: Add missing `Flush`, which is part of the enum `xnvme_spec_znd_cmd_mgmt_send_action`.
- `feat`: Change to cdw11, since this is where we store the value in our `struct xnvme_spec_cmd_sfeat` (and this is the only place we use it);
- `be`: Update to represent a current subset of available backends.
- `poll_io`: Updated to represent all asyncs that it's used in.

### DIFF
--- a/lib/xnvme_cli.c
+++ b/lib/xnvme_cli.c
@@ -297,7 +297,7 @@ static struct xnvme_cli_opt_attr xnvme_cli_opts[] = {
 		.opt = XNVME_CLI_OPT_URI,
 		.vtype = XNVME_CLI_OPT_VTYPE_URI,
 		.name = "uri",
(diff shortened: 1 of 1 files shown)


safl (approved):
lgtm

karlowich (approved):
lgtm

PR #729, opened 2026-08-13
Parallelize CI verification

naddinadja:
Here, I am splitting up our verification job in smaller chunks to allow us to parallelize the testing.

Essentially, for one test-suite (e.g. test-debian-trixie-iommu_enabled-pcie.yaml), do not run the steps sequentially:
- sysinfo, enumerate, hugepages, spdk_proble_diag, test_python, test_ramdisk, test_os, test_pci

But instead use GHA matrix to create parallel runners:
- sysinfo, enumerate, hugepages, spdk_proble_diag, test_python
- sysinfo, enumerate, hugepages, spdk_proble_diag, test_ramdisk 
- sysinfo, enumerate, hugepages, spdk_proble_diag, test_os
- sysinfo, enumerate, hugepages, spdk_proble_diag, test_pci

This saves us about 10-15 minutes in the runtime of our CI (without maas), but potentially more, when more tests are added in the future.

PS. adopting this will require us to change the required checks from the main branch protection rule, but I think we could just remove the verify step completely, since thsi step will always be run if the verify-maas is run (which is a required check).

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -233,7 +233,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container:
+        include:
         # Using an alpine container which is the same as upstream but with bash
         # alpine: SPDK not supported, --with-spdk=disabled
         - {os: 'alpine', dh: 'refenv/alpine-bash', ver: 'latest'}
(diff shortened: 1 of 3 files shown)


karlowich (approved):
Cool!

safl (approved):
lgtm.

PR #732, opened 2026-08-14
Multi-process mode for upcie-cuda and upcie-hip

### LINKED ISSUE #728: Multi-process mode for additional upcie backends
In PR #696 multi process mode is implemented for the upcie backend (to not bloat the PR), but an effort should be made to implement multi process mode for upcie-cuda and upcie-hip.

naddinadja:
This branch is built on top of #696 and has all of its commits. I cannot create stacked PRs from forked repositories.

This PR enables multiprocess mode for upcie-cuda and upcie-hip. Since these backends use a lot of the infrastructure from the upcie backend, the only "real" change is to use the admin queue lock on the admin paths. `--device_heap_size` is added to the CLI of xNVMe and to the homi tool.

Closes #728

### DIFF
--- a/lib/upcie/xnvme_be_upcie_cuda_gpu.c
+++ b/lib/upcie/xnvme_be_upcie_cuda_gpu.c
@@ -21,12 +21,22 @@ xnvme_cuda_queue_create(struct xnvme_dev *dev, uint16_t depth, struct xnvme_cuda
 		return -ENOMEM;
 	}
 
+	err = xnvme_be_upcie_mproc_qids_lock(state->ctrlr);
+	if (err) {
+		XNVME_DEBUG("FAILED: xnvme_be_upcie_mproc_qids_lock(); err(%d)", err);
+		cuMemFree((CUdeviceptr)qpair);
+		return err;
+	}
+
 	// The spec says that for systems where memory ordering is not guaranteed, then one should
 	// leave room in the queue to avoid races. Thus, we do so here, by allocating one more than
 	// what is needed.
 	err = nvme_controller_cuda_create_io_qpair(state->ctrlr->ctrl,
 						   (struct nvme_qpair_cuda *)qpair, depth + 1,
 						   &g_upcie_cuda_rte.cuda_heap);
+
+	xnvme_be_upcie_mproc_qids_unlock(state->ctrlr);
+
 	if (err) {
 		XNVME_DEBUG("FAILED: nvme_controller_cuda_create_io_qpair(); err(%d)", err);
 		cuMemFree((CUdeviceptr)qpair);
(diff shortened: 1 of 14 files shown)


safl (approved):
LGTM. A follow-up PR is posted here: https://github.com/xnvme/xnvme/pull/733

lib/upcie/xnvme_be_upcie_cuda_gpu.c
```diff
@@ -41,9 +51,34 @@ void
...
+
+	err = xnvme_be_upcie_mproc_qids_lock(state->ctrlr);
+	if (err) {
+		// Without the admin queue the device-side queues cannot be deleted, and
+		// releasing the GPU memory they still reference would be worse than leaking it.
+		XNVME_DEBUG("FAILED: xnvme_be_upcie_mproc_qids_lock(); err(%d)", err);
+		return;
+	}
+
+	// nvme_controller_cuda_delete_io_qpair() does not return the queue identifier to the
```
karlowich, line 65: This sounds like `nvme_controller_cuda_delete_io_qpair()` should be fixed

tools/homi.c
```diff
@@ -159,9 +167,10 @@ static struct xnvme_cli_sub g_subs[] = {
 			{XNVME_CLI_OPT_URI, XNVME_CLI_POSN},
 			{XNVME_CLI_OPT_NON_POSA_TITLE, XNVME_CLI_SKIP},
 			{XNVME_CLI_OPT_SHM_ID, XNVME_CLI_LREQ},
-			{XNVME_CLI_OPT_HOST_HEAP_SIZE, XNVME_CLI_LOPT},
 			{XNVME_CLI_OPT_ORCH_TITLE, XNVME_CLI_SKIP},
 			{XNVME_CLI_OPT_BE, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_HOST_HEAP_SIZE, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_DEVICE_HEAP_SIZE, XNVME_CLI_LOPT},
```
karlowich, line 173: I actually don't think that HOMI really needs to take a `device_heap_size`. As the 2MiB default suggests, the heap isn't used for anything.
Besides, I don't think there's any reason to use HOMI with the `upcie-cuda` / `upcie-hip` backends.
Due to the `upcie` "family" being compatible `upcie-cuda` would probably only be used for secondary processes.

karlowich (approved):
LGTM. I have two comments, but neither are blocking. Just something to note for the future, I guess.

lib/upcie/xnvme_be_upcie_cuda_gpu.c
safl: Agreed, the fix belongs in `nvme_controller_cuda_delete_io_qpair()` upstream rather than in the backend. Tracked as #735 so this PR does not have to carry it; the workaround here is correct in the meantime.

tools/homi.c
safl: Fair point, and it goes further than homi: if nothing on the primary path needs a device heap, the option should probably be dropped rather than wired through. Tracked as #736.

PR #733, opened 2026-08-16
fix(upcie/mproc): follow-ups from #696 and #732

safl:
#732 is now on `main`, so this PR is rebased onto it and stands alone: four fixes, a refactor and a docs cleanup.

## What is in it

Four fixes, a refactor and a docs cleanup from local review of #696 (merged as `b8a1964d`) and #732, `+45 -34` across 3 files:

- `fix(upcie/mproc): put per-BDF lock file under /tmp/`
  - `+2 -1` in 1 file
  - The `/tmp%s-lock` snprintf missed a slash; the lock resolved under `/` instead of `/tmp/`.
- `fix(upcie/mproc): clean up on _rte_init and ctrlr_shm_init error paths`
  - `+15 -4` across 2 files
  - Mproc state leaked when `_rte_init` failed after `mproc_rte_init` succeeded; `shm_fd`/`lock_fd` were also read uninitialised in ctrlr_shm_init's failed label.
- `refactor(upcie/mproc): replace custom completion-handling with helper`
  - `+10 -17` in 1 file
  - `_recover_aq` kept its own copy of the CQ walk; it now drives the shared helper, as suggested in review.
- `fix(upcie/mproc): only the owner enables PCI bus master`
  - `+10 -5` in 1 file
  - Secondaries called `_pci_enable_bus_master` too, without the need or the config-space write permission.
- `fix(upcie/mproc): skip qids self-copy in the primary`
  - `+6 -5` in 1 file
  - Same-address memcpy (UB per C99) because `ctrl == &shm->ctrl` in the primary.
- `docs(upcie/mproc): fix "allocates it own" and a double negative`
  - `+2 -2` in 1 file
  - Prose fixes in `docs/background/backends/upcie/mproc.md`.

## What is not

- The stale-shm foot-gun: a fresh homi occasionally came up as "Lock file already claimed" against a free `shm_id`. Not root-caused, needs a live repro.
- Freeing the qid inside uPCIe's `nvme_controller_cuda_delete_io_qpair()` rather than in the backend, from review of #732; tracked at #735.
- `_recover_aq` resyncs `head` and `phase` but not `tail`, so an owner that died between submitting and unlocking may leave the shared SQ cursor behind the device. Read from the code, not reproduced, and it predates this PR.
- Trailer discipline on `856536fa` (`992df891` before #732 was rebase-merged), now on `main` and not retroactively fixable.
- Wiring `--device_heap_size` beyond `homi start`. Review of #732 questioned whether the option is needed at all, so this may end as a removal instead; tracked at #736.

## Verified

Built and smoke-tested on warp (RTX A6000) and wave (RX 7800 XT). `homi --be upcie --shm_id 1` + one xnvmeperf secondary matched the pre-fixup baseline (1.64 M / 1.80 M IOPS); two concurrent secondaries fanned out cleanly (853 k+852 k / 898 k+898 k). `/dev/shm/xnvme-upcie-*` and `/tmp/xnvme-upcie-*` clean after primary exit on both boxes.

Those runs predate two later changes to the branch, neither of which alters what was measured. The `nvme_qpair_reap_cpl` refactor is build-and-format-checked only, and `_recover_aq` runs on owner death, a path the smoke test does not reach. The rebase onto the merged #732 dropped the seven shared commits by patch-id and left a byte-identical tree, verified with `git diff` against the pre-rebase head.

## GPU HW verification

Not on this branch. GPU HW verification requires uPCIe v0.6.0 and the DKMS `dmabuf-import` module; this PR's base sits on `main`'s pre-#727 uPCIe snapshot which uses the in-tree udmabuf path and cannot use `dmabuf-import`. All six commits here are host-mproc code paths and touch no GPU or dmabuf code, so this is orthogonal to what's under review. GPU HW verification for the mproc surface happens at #734.

### DIFF
--- a/lib/upcie/xnvme_be_upcie_mproc.c
+++ b/lib/upcie/xnvme_be_upcie_mproc.c
@@ -676,9 +676,10 @@ failed:
  *
  * Wraps xnvme_be_upcie_ctrlr_mutex_lock() for callers that allocate or release
  * I/O queue identifiers. In a secondary process ctrlr->ctrl is a process-local
- * copy of the controller, so the bitmap has to be pulled in from shared memory
- * before it is consulted. In a primary process, and when not in multi-process
- * mode, this is a no-op beyond the locking itself.
+ * copy of the controller and has to see the current bitmap before it is
+ * consulted; in a primary process ctrlr->ctrl points into the shared segment
+ * itself so no import is needed, and outside multi-process mode ctrlr->shm is
+ * NULL, so this is a no-op beyond the locking itself.
  *
  * @param ctrlr The backend controller
  *
@@ -695,7 +696,7 @@ xnvme_be_upcie_mproc_qids_lock(struct xnvme_be_upcie_ctrlr *ctrlr)
 		return err;
 	}
 
-	if (ctrlr->shm) {
+	if (ctrlr->shm && ctrlr->ctrl != &ctrlr->shm->ctrl) {
 		memcpy(ctrlr->ctrl->qids, ctrlr->shm->ctrl.qids, sizeof(ctrlr->ctrl->qids));
 	}
 
@@ -710,7 +711,7 @@ xnvme_be_upcie_mproc_qids_lock(struct xnvme_be_upcie_ctrlr *ctrlr)
 void
 xnvme_be_upcie_mproc_qids_unlock(struct xnvme_be_upcie_ctrlr *ctrlr)
 {
-	if (ctrlr->shm) {
+	if (ctrlr->shm && ctrlr->ctrl != &ctrlr->shm->ctrl) {
 		memcpy(ctrlr->shm->ctrl.qids, ctrlr->ctrl->qids, sizeof(ctrlr->shm->ctrl.qids));
 	}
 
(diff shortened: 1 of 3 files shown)


lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -228,7 +228,7 @@ _recover_aq(struct xnvme_be_upcie_ctrlr *ctrlr)
 
 	cqe = &cq[head];
 
-	while ((cqe->cid < 0xFFFF) && ((cqe->status & 0x1) == phase)) {
+	while ((cqe->status & 0x1) == phase) {
```
karlowich: `((cqe->cid < 0xFFFF) && ((cqe->status & 0x1) == phase))`
This is how it is done in `nvme_qpair_reap_cpl()` from `include/upcie/nvme/nvme_qpair.h` and the cuda equivalent.

`0xFFFF` is not supposed to be used. 

The NVMe spec (Figure 91: Command Dword 0) reads:
"Command Identifier (CID): This field specifies a unique identifier for the command when combined with the Submission Queue identifier.
The value of FFFFh should not be used as the Error Information log page (refer to section 5.2.12.1.2) uses
this value to indicate an error is not associated with a particular command."

karlowich (changes requested):
LGTM.
Just one comment.

lib/upcie/xnvme_be_upcie_mproc.c
karlowich: Actually, perhaps this function should use `nvme_qpair_reap_cpl()` instead of duplicating the code.

safl: Ack, both points taken. `_recover_aq` now drives `nvme_qpair_reap_cpl()` instead of keeping its own copy of the walk, which brings the cid check back with it. Squashed into the original commit, so this thread will go outdated.

karlowich (approved):
Great!

PR #734, opened 2026-08-16
feat(be_upcie): evolve onto uPCIe v0.6.0 dmamem with mproc integration

safl:
#733 is now on `main`, so this PR is rebased onto it and stands alone. Supersedes #727.

## Where this fits

- **#727** evolved `be_upcie` onto uPCIe v0.6.0's dmamem in place (superseded by this PR; closed).
- **#696** (merged) added multi-process for the host `--be upcie` path.
- **#732** (merged) stacked multi-process onto `upcie-cuda` and `upcie-hip`.
- **#733** (merged) collected follow-ups from reviewing #696 and #732: four fixes, a refactor and a docs cleanup.
- **#734** is what #727 becomes on top of #733: same dmamem intent, with the mproc surface integrated onto it rather than landing alongside.

## What is in it

Three commits on `main`, `+4790 -1607` across 75 files:

- `chore(upcie): vendor uPCIe v0.6.0`
  - `+3540 -438` across 49 files
  - The vendored tree is byte-identical to upstream's released v0.6.0 tag, checked with `diff -r` against the tag's `include/upcie`. The bulk of the diff.
- `feat(be_upcie): port to uPCIe v0.6.0 dmamem and add vfio-cdev attachment`
  - `+1244 -1169` across 25 files
  - The backend follows the vendored library, carrying the mproc surface through the port, collapses the per-backend command paths onto it, and groups the controller and runtime state rather than adding to two already-flat structs. See the commit body.
- `fix(xnvmeperf/cuda): guard against zero lba_nbytes in validate`
  - `+6 -0` in 1 file
  - Turns a divide-by-zero into a clear `-EINVAL`.

## What review changed

Two rounds with @karlowich, both about commit structure and duplication, and both landed as rewrites rather than commits on top:

- The port used to remove multi-process support and a follow-up commit added it back. The intermediate commit did not build: it dropped the mproc declarations from `xnvme_be_upcie.h` while `xnvme_be_upcie_mproc.c` stayed in the build. Squashed.
- The port then left three near-identical copies of each command path, one per backend, and a follow-up deleted two of them. Same shape, and this one hid because nothing builds CUDA or HIP: not this repo's CI, which has no cuda or hip job, and not a host without the toolkits. Squashed too, so the six GPU path files are only ever deleted, never ported and then deleted.
- The port added fields to two already-large structs and a follow-up regrouped them, so the same fields were read twice. Squashed for the third time, and the port now introduces them grouped: an `attach` and a `mproc` member on the controller, and `cdev`, `type1` and `mem` on the runtime.

What the generalization does: every payload now translates through a `struct dmamem`, so the paths differed only in which one. The device's dmamem is recorded in `xnvme_be_upcie_state` at `dev_open`, the single implementation reads it from there, and the GPU backends keep `dev` and `mem` and wire the shared paths into their own `xnvme_be_{admin,sync,async}`. Six files and 774 lines go away.

Also from review: the vfio helpers moved to `lib/upcie/xnvme_be_upcie_vfio.c`, taking 216 lines out of `dev.c`; `_bdf_has_vfio_cdev()` now calls the resolver instead of repeating it; a doc-comment describing mode derivation now sits on the function it describes, `xnvme_be_upcie_mode_from_driver()`; the queue offsets became one struct; the secondary attach error path no longer leaks its PRP scratch into the primary's heap; and the VFIO attachment modes are documented under the backend page rather than the multi-process one.

## Verification

GPU hardware verification needs uPCIe v0.6.0 and the DKMS `dmabuf-import` module, so neither #732 nor #733 could do it: this is where the GPU paths introduced in #732 first meet real hardware. `modprobe dmabuf_import` before running.

Everything below was re-measured on the current head, on both boxes, after the review rewrites. The measurements are no longer hand-run: they come from a cijoe workflow, kept on a separate branch since it needs iterating on its own rather than here. The generalization moved the buffer-translation path from a per-backend global to the dmamem recorded in the device state, which is what these runs exercise, so earlier figures could not simply be carried over. Nothing moved:

| box | backend | shape | now | before |
|---|---|---|---|---|
| warp | `upcie` | 1 dev, qd32 128 KiB | 5833 MiB/s flat, 5825 mproc | 5997 / 5840 |
| warp | `upcie-cuda` | 1 dev, qd32 128 KiB | 6948 MiB/s flat, 6948 mproc | 6948 / 6947 |
| warp | `upcie` | 9 dev, 512 B randread qd128, `0,32` | 16.44M flat, 16.35M mproc, 16.44M 2x mproc | 16.23M / 16.39M |
| wave | `upcie` | 1 dev, qd32 128 KiB | 7126 MiB/s flat, 7127 mproc | 7127 |
| wave | `upcie-hip` | 1 dev, qd32 128 KiB | 7127 MiB/s flat, 7127 mproc | 7127 |

Medians of three runs, release build, `performance` governor, zero failed IOs throughout. The multi-process shapes are not below the flat ones anywhere, which is the claim that matters: routing through a HOMI primary costs nothing on the data path. The attachment modes were swept separately, single-device: `iommufd` 4.03M IOPS against `type1` 4.03M.

This also closes the build gap, which was the real risk on this branch. CI has no cuda or hip job, so the GPU translation units are compiled only on these two boxes; warp built with CUDA enabled (`nvcc 13.3`) and wave with HIP enabled (`amdhip64`), so both halves of the generalization now compile as well as run.

Four things the earlier hand-run rounds established that the current runs do not re-cover, kept because they still hold and are worth knowing:

- `/dev/shm/xnvme-upcie-*` and `/tmp/xnvme-upcie-*` are clean after the primary exits.
- Sequential mproc secondaries on `upcie-cuda` are stable with zero failed IOs.
- Against #727, every comparable shape landed within its run-to-run noise band, so the dmamem evolution itself did not cost throughput.
- VFIO_TYPE1 can only open one device per IOMMU group, the defect #727 documented as upcie#56. That is why the mode sweep above is single-device: against all nine, the second device sharing a group fails at `-EIO`.

## Known limitation, deferred (#737)

Two concurrent secondaries on `--be upcie-cuda` (and by inheritance `--be upcie-hip`) collide on GPU physical memory: `xnvme_be_upcie_cuda_buf_alloc` returns CUDA memory for every `xnvme_buf_alloc`, including admin identify scratch. Two processes with independent CUDA contexts each `cuMemAlloc` at the same GPU physical offset, `dmamem_va_to_iova` returns the same PCIe address in both, and the NVMe controller DMAs both identify responses to the same physical memory. Confirmed by tracing PRP1 on both processes in the admin path.

Predates #734 and #732; #734 is simply the first place the code path is reachable, so the first place the behaviour is observable. Sequential secondaries, single-process CUDA, and all host-mproc combinations are unaffected. Failure mode is a clean `-EINVAL` from the defensive check.

Fix direction: per-process host bounce buffer on the CUDA/HIP admin path with `cuMemcpyHtoD`/`hipMemcpyHtoD` after reap. Deferred to a follow-up and tracked at #737: this is a gap in a new feature that has not shipped anywhere before, not a regression.


karlowich (changes requested):
I think the two commits for `dmamem` and `mproc` support should be either partially or fully squashed. Since the multi-process support is already present in all the uPCIe backends, it is confusing that the `mproc` support is removed in the `dmamem` commit and then added back in the next commit. This would make the diff smaller and it will be easier to tell exactly what the `dmamem` changes are.

I assume this is an artifact of `dmamem` being a separate backend in a previous PR.

safl (commented):
Squashed them fully; the branch is now three commits.

You were right about the cause, and it is worse than confusing: the `dmamem` commit did not build on its own. It dropped the mproc declarations from `xnvme_be_upcie.h` while `xnvme_be_upcie_mproc.c` stayed in `lib/upcie/meson.build`, so building at that commit fails with `struct xnvme_be_upcie_ctrlr has no member named shm` and friends. I confirmed this by building each commit in a worktree with `make config-slim`.

A partial split does not work out either. The mproc shared state sits directly on the memory model the port replaces, so any "dmamem only" commit either carries mproc through the change, which is what the squash now does, or removes it, which is what made the old split broken.

All three commits build now, checked individually with the same worktree. The tree at the head is byte-identical to what was reviewed and benchmarked, `git diff` against the pre-squash head is empty, so the numbers in the description still stand.

docs/background/backends/upcie/mproc.md
```diff
@@ -203,14 +203,25 @@ startup. Because uPCIe also needs hugepages for its DMA memory, the two cannot
...
 
+### Attachment-mode restriction
+
+Multi-process mode requires the device to be bound to `uio_pci_generic`
+(the UIO_LUT attachment mode). The `vfio-pci` paths that `--be upcie`
+otherwise supports (VFIO_CDEV via `/dev/iommu`, VFIO_TYPE1 via the
+legacy container) hold per-process fds that a secondary cannot inherit,
+so `_rte_init` returns `-ENOTSUP` up front when `opts.shm_id` is set
+under either mode. `upcie-cuda` and `upcie-hip` already required
+`uio_pci_generic` before multi-process was added, so nothing new there.
```
karlowich: "`upcie-cuda` and `upcie-hip` already required `uio_pci_generic` before multi-process was added, so nothing new there."

This sentence has a sense of time that won't age well. It should just state the requirements.

docs/background/backends/upcie/mproc.md
```diff
@@ -203,14 +203,25 @@ startup. Because uPCIe also needs hugepages for its DMA memory, the two cannot
...
+
+When the device is bound to `vfio-pci`, the environment variable
+`XNVME_UPCIE_VFIO_MODE` can force `iommufd` or `type1` selection; when
+unset, the backend prefers `iommufd` if `/dev/iommu` is available and a
+`vfio-dev` entry exists, falling back to `type1` otherwise. Neither
+setting is compatible with multi-process mode.
```
karlowich: This part seems out of place. I think the VFIO modes should documented under a different header.

lib/upcie/xnvme_be_upcie.h
```diff
@@ -22,21 +22,44 @@
 struct xnvme_queue_upcie {
 	struct xnvme_queue_base base;
 	struct nvme_qpair qpair;
-	uint8_t _rvds[168];
+	size_t sq_offset;  ///< Heap offset of the qpair's SQ, for delete
+	size_t cq_offset;  ///< Heap offset of the qpair's CQ, for delete
+	size_t prp_offset; ///< Heap offset of the qpair's per-request PRP scratch, for delete
```
karlowich: These three fields appear in `xnvme_queue_upcie`, `xnvme_be_upcie_ctrlr_shm` and `xnvme_be_upcie_ctrlr`. I think we should consider creating a struct containing them instead.

lib/upcie/xnvme_be_upcie.h
```diff
@@ -45,16 +68,23 @@ struct xnvme_be_upcie_ctrlr_shm {
...
+	char lock_name[128];                  ///< Per-BDF primary-election lock path
+	int lock_fd;                          ///< Owned by primary while it holds the controller
+	char shm_name[64];                    ///< POSIX shm name for the per-controller segment
+	int shm_fd;                           ///< Owned by primary; -1 in secondaries
+	struct xnvme_be_upcie_ctrlr_shm *shm; ///< Per-controller shm (NULL outside mproc)
 
-	char shm_name[64];
-	int shm_fd;
-	struct xnvme_be_upcie_ctrlr_shm *shm;
+	size_t aq_rpool_prp_offset; ///< Heap offset of the secondary's local admin-rpool PRPs
```
karlowich: Does it make sense to create some more structs to reduce the size and complexity of this quite large struct.

lib/upcie/xnvme_be_upcie.h
```diff
@@ -91,16 +128,33 @@ struct xnvme_be_upcie_mproc {
...
+	struct dmamem_heap heap;
+	struct vfio_container
+		type1_container;   ///< VFIO_TYPE1 mode; shared, first ctrlr sets the iommu
+	int type1_container_alive; ///< Whether the container is open
+	int type1_iommu_set;       ///< Whether VFIO_SET_IOMMU has been applied to it
+	int hp_alive;              ///< Whether hp holds an allocation
+	int dmem_alive;            ///< Whether dmem wraps hp
+	int heap_alive;            ///< Whether heap is initialized
+	struct xnvme_be_upcie_mproc *mproc; ///< NULL when not in multi-process mode
 	int is_initialized;
```
karlowich: Does it make sense to create some more structs to reduce the size and complexity of this quite large struct.

lib/upcie/xnvme_be_upcie_admin.c
```diff
@@ -16,33 +16,59 @@ xnvme_be_upcie_sync_cmd_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf
 			      void *XNVME_UNUSED(mbuf), size_t XNVME_UNUSED(mbuf_nbytes))
 {
 	struct xnvme_be_upcie_state *state = (void *)ctx->dev->be.state;
-	struct xnvme_be_upcie_ctrlr *ctrlr = state->ctrlr;
+	struct xnvme_be_upcie_ctrlr *be_ctrlr = state->ctrlr;
```
karlowich: I don't think the `ctrlr` to `be_ctrlr` adds much, but it makes the diff harder to read

lib/upcie/xnvme_be_upcie_async.c
```diff
@@ -10,32 +10,26 @@
...
 #include <xnvme_be_upcie.h>
 
-/**
- * Command Queue for asynchronous command submission and completion
- */
 int
 xnvme_be_upcie_queue_init(struct xnvme_queue *queue, int XNVME_UNUSED(opts))
 {
-	struct xnvme_queue_upcie *upcie_queue = (void *)(queue);
+	struct xnvme_queue_upcie *dq = (void *)queue;
```
karlowich: `upcie_queue` is renamed to `dq` across this entire file.
I don't think the rename adds much value, but makes the diff harder to read.
If the rename is required, I think we are better off adding it in a separate commit.

lib/upcie/xnvme_be_upcie_admin.c
karlowich: I see we use `be_ctrlr` in `xnvme_be_upcie_cuda_admin.c`, I think it's fine to follow this, but I would prefer the rename in a separate commit.

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -39,6 +38,68 @@ xnvme_be_upcie_get_driver_name(const char *bdf, char *driver_name, size_t driver
 	return 0;
 }
 
+int
+xnvme_be_upcie_resolve_vfio_cdev(const char *bdf, char *cdev_path, size_t cdev_path_len)
```
karlowich: I propose moving the vfio helpers into a separate file, since there is a considerable amount of code specific to this part.

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -39,6 +38,68 @@ xnvme_be_upcie_get_driver_name(const char *bdf, char *driver_name, size_t driver
...
+	closedir(dir);
+	return found ? 0 : -ENOENT;
+}
+
+/**
+ * Return non-zero if the target BDF has a vfio-cdev entry, i.e.
+ * /sys/bus/pci/devices/<bdf>/vfio-dev/vfio* exists.
+ */
+static int
+_bdf_has_vfio_cdev(const char *bdf)
```
karlowich: This function is almost identical to `xnvme_be_upcie_resolve_vfio_cdev()`, do we need both or can one call the other?

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -56,104 +117,375 @@ _rte_term(void)
...
+ *
+ * /dev/iommu + one IOAS + a hugepage-backed memfd imported via
+ * IOMMU_IOAS_MAP_FILE. All controllers opened afterwards attach into
+ * this same IOAS so their PRPs translate arithmetically off the shared
+ * base_iova.
+ */
+static int
+_rte_init_vfio_cdev(size_t heap_size)
+{
+	size_t hugepgsz = 2ULL * 1024 * 1024;
```
karlowich, line 74: Is it safe to hardcode the hugepage size? Is there a guarantee that it isn't 1 GiB?

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -616,30 +616,29 @@ xnvme_be_upcie_mproc_ctrlr_shm_attach(struct xnvme_dev *dev, struct xnvme_be_upc
...
 
-	err = xnvme_be_upcie_mproc_create_or_delete_io_qpair(ctrlr, &ctrlr->sync, 16, true);
+	err = xnvme_be_upcie_mproc_create_io_qpair(ctrlr, &ctrlr->sync, 16, &ctrlr->sync_sq_offset,
+						   &ctrlr->sync_cq_offset,
+						   &ctrlr->sync_prp_offset);
 	if (err) {
-		XNVME_DEBUG("FAILED: create with "
-			    "xnvme_be_upcie_mproc_create_or_delete_io_qpair(); err(%d)",
-			    err);
-		nvme_request_pool_term_prps(ctrlr->ctrl->aq.rpool, &g_upcie_rte.heap);
```
karlowich, line 642: Do we no longer need to terminate the request pool?

karlowich (changes requested):
Quite a few comments inline.
Additionally, I was expecting to see some paths become generalized.
Can't we reduce the code to a single `sync`, `async` and `admin` path now?
Perhaps I'm misunderstanding.

safl (commented):
Done, and you were right about the generalization: after the port every payload translates through a `struct dmamem`, so the three copies of each path differed only in which one. The device's dmamem is now recorded in `xnvme_be_upcie_state` at `dev_open`, one implementation reads it from there, and the GPU backends keep `dev` and `mem` and wire the shared paths into their own `xnvme_be_{admin,sync,async}`. Six files and about 800 lines gone.

That started as its own commit on top, and it repeated the mistake you caught last round in a quieter form: the port would have ported three copies and the next commit deleted two of them, and nothing would have compiled the intermediate GPU code, since neither this repo's CI nor a host without the toolkits builds CUDA or HIP. So it is squashed into the port, and the six files are only ever deleted.

Also addressed: both renames reverted, so the port diff no longer carries them. The vfio helpers moved to `lib/upcie/xnvme_be_upcie_vfio.c`, 216 lines out of `dev.c`. `_bdf_has_vfio_cdev()` now calls `xnvme_be_upcie_resolve_vfio_cdev()` instead of repeating it. The doc-comment about deriving the mode from the bound driver was describing code that had been inlined into `ctrlr_init`, so it now sits on `xnvme_be_upcie_mode_from_driver()`, which is that logic extracted. The three queue offsets became `struct xnvme_be_upcie_qpair_offsets`, which also collapses the six-parameter create/delete signatures. The mproc doc no longer has the sentence with a sense of time, and the VFIO attachment modes moved to their own section under the backend page, with the mproc page linking to it.

Two of your comments I answered rather than changed. The 2 MiB hugepage size is safe to hardcode: `dmamem_from_memfd()` turns it into an explicit `MFD_HUGE_2MB` rather than following a system default, and rejects anything that is not 2 MiB or 1 GiB, so a box with only a 1 GiB pool fails at `ftruncate` instead of misbehaving. And the request pool is terminated, at `xnvme_be_upcie_mproc_ctrlr_shm_term()`; what you pointed at was a real hole next to it, though, since the attach error path freed the rpool without releasing its PRP scratch, and in mproc that scratch lives in the primary's heap and so outlived the failing secondary. Fixed.

One left open: your suggestion to break up the two large structs. The offsets grouping is in, but the per-mode attachment state in `xnvme_be_upcie_rte` and `xnvme_be_upcie_ctrlr` is not, and I would rather do that as a follow-up than add more churn to a branch whose GPU half still needs re-measuring. Happy to do it here if you would prefer.

All three commits build individually, checked in a worktree. The performance numbers in the description predate this rewrite and are being re-run: the generalization moved the buffer-translation path, which is precisely what they measure.

docs/background/backends/upcie/mproc.md
safl: ack

docs/background/backends/upcie/mproc.md
safl: ack

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie_admin.c
safl: ack

lib/upcie/xnvme_be_upcie_async.c
safl: ack

lib/upcie/xnvme_be_upcie_admin.c
safl: ack

lib/upcie/xnvme_be_upcie_dev.c
safl: ack

lib/upcie/xnvme_be_upcie_dev.c
safl: ack

lib/upcie/xnvme_be_upcie_dev.c
safl: Safe to hardcode: `dmamem_from_memfd()` maps the size onto an explicit `MFD_HUGE_2MB` rather than following a system default, and rejects anything that is not 2 MiB or 1 GiB. A box with only a 1 GiB pool fails at `ftruncate` rather than misbehaving.

lib/upcie/xnvme_be_upcie_mproc.c
safl: It is terminated, in `xnvme_be_upcie_mproc_ctrlr_shm_term()`. You did point at a real hole next to it though: the attach error path freed the rpool without releasing its PRP scratch, and in mproc that scratch lives in the primary's heap, so it outlived the failing secondary. Fixed.

lib/upcie/xnvme_be_upcie.h
```diff
@@ -19,42 +19,95 @@
...
 	_Atomic bool is_initialized; ///< Set by primary once the controller is fully opened
 	pthread_mutex_t aq_mutex;    ///< Process-shared mutex for admin queue access
 	char driver_name[32];
+	struct xnvme_be_upcie_qpair_offsets sync_offsets; ///< Heap offsets of the sync qpair
 	struct nvme_controller ctrl; ///< Embedded controller; pointer fields use primary's VA
 };
 
 /**
  * Shared controller state, one per physical controller, managed by cref.
  */
```
karlowich: This comment belongs to`xnvme_be_upcie_ctrlr`

lib/upcie/xnvme_be_upcie.h
```diff
@@ -64,12 +117,26 @@ struct xnvme_be_upcie_ctrlr {
...
 
-	uint8_t _rvds[120];
+	/**
+	 * Where data buffers for this device live
+	 *
+	 * The host heap for `upcie`, the GPU heap for `upcie-cuda` and
+	 * `upcie-hip`. Assigned by the backend's dev_open, and the reason the
+	 * admin, sync and async paths are shared across the three: they differ
+	 * only in which dmamem a payload translates through.
+	 */
```
karlowich: This docstring should merge with the docstring above `xnvme_be_upcie_state`.

lib/upcie/xnvme_be_upcie_cuda.c
karlowich: These should just point to `g_xnvme_be_upcie_*`. I don't think the structs below are necessary.

lib/upcie/xnvme_be_upcie_hip.c
karlowich: Same comment as for the CUDA

lib/upcie/xnvme_be_upcie_vfio.c
```diff
@@ -0,0 +1,246 @@
...
+ * Attach a controller's group to the RTE's type1 container, and on the
+ * first controller also set_iommu(TYPE1) + MAP_DMA the hostmem hugepage
+ * + init the shared dmamem_heap.
+ *
+ * Rolls back the group open (and any RTE-side state it turned on) on
+ * failure. The container itself and the hostmem hugepage remain owned
+ * by the RTE.
+ */
+int
+xnvme_be_upcie_type1_attach_and_maybe_map(struct xnvme_be_upcie_ctrlr *ctrlr, const char *bdf)
```
karlowich: This function name is weird. `and_maybe_map` does not add anything other than confusion.
I think that part can just be removed.

karlowich (changes requested):
Much better.

Still some comments inline. Don't get confused by the comments on lines you didn't change, I chose these as they allow me to be clearest about my point.

Finally, the commit message of `feat(be_upcie): port to uPCIe v0.6.0 dmamem and add vfio-cdev attachment` has gotten too long in my opinion.

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie_cuda.c
safl: ack

lib/upcie/xnvme_be_upcie_hip.c
safl: ack

lib/upcie/xnvme_be_upcie_vfio.c
safl: ack

lib/upcie/xnvme_be_upcie_cuda.c
```diff
@@ -2,16 +2,25 @@
...
+/**
+ * Backend wiring for `upcie-cuda`
+ *
+ * The admin, sync and async paths are the ones implemented for `upcie`; they
+ * pick up this backend's device memory through xnvme_be_upcie_state.dmem,
+ * assigned by dev_open. Only dev and mem are cuda-specific.
+ */
 #include <xnvme_be.h>
+#include <xnvme_be_nosys.h>
+#ifdef XNVME_BE_UPCIE_CUDA_ENABLED
```
karlowich: Let's move the ifdef to the top as it was before. We can also drop the `xnvme_be_nosys.h`.

lib/upcie/xnvme_be_upcie_hip.c
```diff
@@ -2,16 +2,25 @@
...
+/**
+ * Backend wiring for `upcie-hip`
+ *
+ * The admin, sync and async paths are the ones implemented for `upcie`; they
+ * pick up this backend's device memory through xnvme_be_upcie_state.dmem,
+ * assigned by dev_open. Only dev and mem are hip-specific.
+ */
 #include <xnvme_be.h>
+#include <xnvme_be_nosys.h>
+#ifdef XNVME_BE_UPCIE_HIP_ENABLED
```
karlowich: Let's move the ifdef to the top as it was before. We can also drop the `xnvme_be_nosys.h`.

karlowich (changes requested):
Just two final nitpicks.

The only other thing is, I wish we could encapsulate/abstract the uio/vfio_cdev/vfio_type1 stuff more.
`xnvme_be_upcie_dev.c` is so dense to read. However, I can't exactly pinpoint what I would change.

lib/upcie/xnvme_be_upcie_cuda.c
safl: ack

lib/upcie/xnvme_be_upcie_hip.c
safl: ack

karlowich (approved):
LGTM

PR #739, opened 2026-08-19
Update CIJOE nvmf scripts for RDMA use cases

jswaro:
NOTE: Description generated by GitHub Copilot. 

## Summary

Extends the cijoe NVMe target scripts (`nvme_target_start`, `nvme_target_probe`, `nvme_target_stop`) to support RDMA in addition to TCP, and renames their CLI flags with an `nvme-` prefix to avoid collisions with other cijoe arguments. Each script also gains a `--transport-name` flag so callers can pin `cijoe.run()` calls to a specific configured transport instead of relying on the first one found.

## Changes

- Renamed CLI flags across all three scripts:
  - `--provider` → `--nvme-provider`
  - `--traddr` → `--nvme-traddr`
  - `--trsvcid` → `--nvme-trsvcid`
  - `--trtype` → `--nvme-trtype`
  - `--adrfam` → `--nvme-adrfam`
- Added `rdma` as a supported value for `--nvme-trtype`, alongside the existing `tcp` default, restricted to a fixed choice list.
- Added `--transport-name` to all three scripts and threaded it through every `cijoe.run()` call so setup, probe, and teardown can target a specific cijoe-configured transport.
- Load the correct kernel module (`nvme_{trtype}` / `nvmet_{trtype}`) based on the selected transport type instead of hardcoding `nvme_tcp` / `nvmet_tcp`.
- Added a comment noting RDMA requires kernel modules and supporting software (for example `libibverbs`) configured on both initiator and target, and that failures here may indicate incomplete RDMA setup rather than a script defect.
- Removed an unused provider reference from `nvme_target_probe.py` and workflow YAML files (`transports-tcp-linux.yaml`, `transports-tcp-spdk.yaml`).
- Updated [docs/tutorial/transports/index.md](docs/tutorial/transports/index.md) flag reference table and prose to match the renamed and new flags.

## Dependencies / Risks

- Callers of these cijoe scripts (workflows, other scripts, CI configs) that pass the old flag names (`--provider`, `--traddr`, `--trsvcid`, `--trtype`, `--adrfam`) must be updated to the `--nvme-*` names.
- RDMA path is untested in environments without RDMA hardware/drivers; failures there may reflect environment setup rather than a code defect.


cijoe/scripts/nvme_target_start.py
```diff
@@ -95,20 +102,25 @@ def _start_spdk(args, cijoe):
     nvmf_tgt = spdk_path / "build" / "bin" / "nvmf_tgt"
     nvmf_tgt_src = spdk_path / "app" / "nvmf_tgt"
 
-    # nvme_tcp is the kernel host driver loaded for the initiator-side
+    if cijoe.getconf("overrides.prefer_developer_build", False):
+        xnvme_driver = Path(xnvme_repos) / "builddir" / "toolbox" / "xnvme-driver"
+    else:
+        xnvme_driver = "xnvme-driver"  # use version found on path
```
safl: I can see what this is for, but I would rather we did not take an override of this kind.

If the concern behind it is not being able to purge what was installed afterwards, xNVMe covers that three ways. The first is meson itself: `meson install -C builddir` puts it on the system, and meson can take it off again, though its uninstall is an internal entry point rather than a first-class command, which is why the Makefile wraps both as `make install` and `make uninstall`. The second is apt, through `make build-deb install-deb uninstall-deb`. The third is yum, through `make build-rpm install-rpm uninstall-rpm`. Whichever suits your system, xNVMe comes off it cleanly again, so a developer build is not needed to stay safe.

There is also a cost to carrying it: every site that invokes an installed artefact has to learn about the override and then be kept in sync with the others. Dropping it avoids that, and leaves this patch as purely the RDMA fix it says it is.

safl (commented):
This looks excellent, and the `--transport` plumbing is complete across all four `cijoe.run` sites. There is one slightly controversial thing, which I have commented on inline.

The short of it: we would rather encourage an actual system install than a developer build run out of the repository. Once the libraries and tools are deployed on the system they can behave differently from a build tree, and catching those differences early has consistently been preferable to discovering them later. The Makefile has helpers to make that painless in both directions, so nothing is left behind on the system afterwards.

With the override dropped this is good to go from my side.

---

FYI, and unrelated to the above: `xnvme-driver` itself is something we have been considering how to replace for a long time. It bundles several separate concerns into one script, and those have since grown into their own tools, each doing one part: [devbind](https://github.com/xnvme/devbind) for binding devices to `uio_pci_generic` or `vfio-pci`, [hugepages](https://github.com/xnvme/hugepages) for reserving hugepages, and [iommu](https://github.com/safl/iommu) for the IOMMU side.

Perhaps one of those would suit the use-case you reached for the override for, better than pointing at a build-tree copy of `xnvme-driver`?

cijoe/scripts/nvme_target_start.py
jswaro: I dropped the commit.

cijoe/scripts/nvme_target_start.py
```diff
@@ -53,6 +54,12 @@ def add_args(parser: ArgumentParser):
         default="ipv4",
         help="Address family",
     )
+    parser.add_argument(
+        "--transport",
```
safl: Two options here now say "transport" and mean different things: `--trtype` is the NVMe fabrics transport, `--transport` is which host to run on. Someone will pass `--transport rdma`. `--transport-name`, matching the keyword it feeds, would avoid it, and renaming is free before this lands.

cijoe/scripts/nvme_target_start.py
```diff
@@ -95,20 +102,20 @@ def _start_spdk(args, cijoe):
...
 
-    # nvme_tcp is the kernel host driver loaded for the initiator-side
+    # nvme_tcp/nvme_rdma is the kernel host driver loaded for the initiator-side
     # `nvme discover` in the probe step; nvmf_tgt itself runs in userspace and
     # the PCIe device is rebound to vfio-pci by xnvme-driver, so nvmet and
-    # nvmet_tcp are not needed on the SPDK target path.
+    # nvmet_tcp/nvmet_rdma are not needed on the SPDK target path.
     drivers = [
-        "modprobe nvme_tcp",
+        f"modprobe nvme_{args.trtype}",
```
safl: In addition to `nvme_rdma` and `nvmet_rdma` (line 160), the RDMA fabric also needs the core stack and a provider loaded, right? Might be worth a comment here noting that the RDMA path expects the fabric to already be up, so a failure there does not read as a script bug.

cijoe/scripts/nvme_target_start.py
```diff
@@ -150,9 +157,9 @@ def _start_linux(args, cijoe):
         "xnvme-driver reset",
         "modprobe nvme",
         "modprobe nvmet",
-        "modprobe nvmet_tcp",
+        f"modprobe nvmet_{args.trtype}",
     ]
-    err, cmd = _run_all(cijoe, drivers)
+    err, cmd = _run_all(cijoe, drivers, args.transport)
```
safl: Positional here and on line 183, keyword on line 113. Same call, two spellings.

safl (commented):
There is still a mention of `overrides.prefer_developer_build` in the description; it appears nowhere in the diff.

cijoe/scripts/nvme_target_start.py
jswaro: If the goal is to avoid duplication of terminology, we can call it 'host', which would be inline with what we are talking about. The collision is between the cijoe terminology, and NVMeoF terminology regarding the word transport. Since NVMe is more fixed, then I'd suggest using `--host` and translating that into the script into the cijoe `transport` parameter. 

```python 
    parser.add_argument(
        "--host",
```

```
    err, cmd = _run_all(cijoe, drivers, args.host)
```

Thoughts?

cijoe/scripts/nvme_target_start.py
jswaro:
```
    # RDMA requires kernel modules and supporting software (e.g. libibverbs) to be
    # installed and configured on both the initiator and target. An error here may
    # reflect incomplete RDMA setup rather than a fault in this script.
```

How about this added directly above the ```drivers = ....``` line?

cijoe/scripts/nvme_target_start.py
jswaro: I'll fix that

cijoe/scripts/nvme_target_start.py
safl: I am back-paddling a bit here... ideally then the **cijoe** terminology would be used consistently everywhere, but it is probably not. It should probably be ``transport`` or ``transport-name`` since that is what is in the cijoe-config. Then perhaps use e.g. ``nvme-trtype`` / ``--nvme-trtype`` then that is close the spec-defined names. How about that?

jswaro: That makes sense to me, and is clear regarding the naming conflicts so that the user knows very clearly what is being provided by which argument. Want me to make the change? 

Changing the argument name at this point means the tutorials need to be updated.

karlowich (dismissed):
LGTM

cijoe/scripts/nvme_target_start.py
```diff
@@ -95,20 +102,23 @@ def _start_spdk(args, cijoe):
...
         "xnvme-driver",
     ]
-    err, cmd = _run_all(cijoe, drivers)
+    err, cmd = _run_all(cijoe, drivers, args.transport)
     if err:
         log.error("FAILED: driver issue: %s (errno=%d)", cmd, err)
         return err
 
-    cijoe.run("pkill -f nvmf_tgt || true")
+    cijoe.run("pkill -f nvmf_tgt || true", args.transport)
```
safl: `cijoe.run` is `run(self, cmd, cwd=None, env={}, transport_name=None)`, so this
passes `args.transport` as `cwd`. With `--transport` given, the pkill runs on
the default transport and `cwd` is set to the transport's name: `SSHTransport`
turns that into `cd <name>; pkill -f nvmf_tgt || true`, `LocalTransport` hands
it to `subprocess`. Either way the stale `nvmf_tgt` on the target host is not
killed, and `_start_spdk` goes on to start a new one.

This came out of the "Positional here and on line 183, keyword on line 113"
comment, but those are two different calls. `_run_all(cijoe, commands,
transport)` takes the transport third, so positional is right in the two
`_run_all` cases; `cijoe.run` takes it fourth, so it needs the keyword here.

CI does not cover it: `--transport` defaults to `None`, so `cwd` is `None` and
the mistake is invisible unless the option is used.

safl (commented):
Reviewed with a coding agent. The finding below comes from reading
`cijoe.core.command.run()` and `cijoe.core.transport`; the RDMA path was not
exercised. Push back where it is wrong.

cijoe/scripts/nvme_target_start.py
safl: Let's go with ``transport_name`` / ``--transport-name`` matching the argument to the **cijoe** ``run()`` function.

cijoe/scripts/nvme_target_start.py
jswaro: I'll fix that.

cijoe/scripts/nvme_target_start.py
safl: And for the NVMe args. spell them out as ``nvme_trtype`` / ``--nvme-trtype``, ``nvme_trsvcid`` / ``--nvme-trsvcid`` etc.

cijoe/scripts/nvme_target_stop.py
```diff
@@ -4,7 +4,7 @@
 ======================================
 
 Tear down whichever NVMe TCP target was set up by ``nvme_target_start``.
```
safl: Still "NVMe TCP" mentions here despite it being generalized.

Note made here since it is not possible to anchor a comment when there is no change.

cijoe/scripts/nvme_target_probe.py
```diff
@@ -17,28 +17,34 @@
 
 def add_args(parser: ArgumentParser):
     parser.add_argument(
-        "--provider",
+        "--nvme-provider",
```
safl, line 20: Still "NVMe TCP" mentions here despite it being generalized.

Note made here since it is not possible to anchor a comment when there is no change.

safl (commented):
Minor things inline. The description still lists `overrides.prefer_developer_build` under what the SPDK path respects, and the commit behind it is no longer on the branch, so it describes functionality this no longer has. On the commit messages: `85cc3cb4`'s body says the scripts take a transport from the cijoe config "if --transport is provided on the command line"; the option it adds is `--transport-name`. `6d88a4ad`'s subject scopes it to the nvme-target workflows, but it also renames the options in `nvme_target_probe.py`, `nvme_target_start.py` and `nvme_target_stop.py`, which is the larger part of that commit.

The inline comments are two docstrings that still describe a TCP-only target, a line of prose apiece.

What was verified: the three commits were read, along with the resulting `nvme_target_start.py`, `nvme_target_stop.py`, `nvme_target_probe.py`, the workflows under `cijoe/workflows/` and the pages under `docs/tutorial/transports/`. Nothing was built or run; neither the TCP nor the RDMA path was exercised.

cijoe/scripts/nvme_target_stop.py
jswaro: Fixed

cijoe/scripts/nvme_target_probe.py
jswaro: Fixed

docs/tutorial/transports/index.md
```diff
@@ -62,13 +62,13 @@ linux/index
...
-| `--trsvcid`    | `4420`        | Transport service id (TCP port).             |
-| `--trtype`     | `tcp`         | Transport type.                              |
-| `--adrfam`     | `ipv4`        | Address family.                              |
+| Argument            | Default       | Description                                  |
+|---------------------|---------------|----------------------------------------------|
+| `--nvme-provider`   | `spdk`        | `spdk` or `linux`.                           |
+| `--nvme-traddr`     | `127.0.0.1`   | Transport address (IP) for the listener.     |
+| `--nvme-trsvcid`    | `4420`        | Transport service id (TCP port).             |
+| `--nvme-trtype`     | `tcp`         | Transport type.                              |
+| `--nvme-adrfam`     | `ipv4`        | Address family.                              |
```
safl, line 71: `--transport-name` is missing from this table, which the line above introduces as the flags the three scripts share. It is added to all three in this PR.

cijoe/scripts/nvme_target_probe.py
```diff
@@ -17,28 +17,34 @@
...
-        "--trsvcid",
+        "--nvme-trsvcid",
         type=str,
         default="4420",
-        help="Transport service id (TCP port)",
+        help="Transport service id (Port) of the NVMe listener",
     )
     parser.add_argument(
-        "--trtype",
+        "--nvme-trtype",
```
safl: `--nvme-trtype` takes any string here, while `nvme_target_start.py` gives the same flag `choices=["tcp", "rdma"]`. A typo is caught when starting the target and passed through to `nvme discover` when probing it. Add the same `choices=`.

cijoe/scripts/nvme_target_stop.py
```diff
@@ -70,9 +79,9 @@ def _stop_linux(args, cijoe):
 def main(args, cijoe):
     """Tear down an NVMe TCP target using the selected provider."""
```
safl, line 80: Still says "NVMe TCP target" after the module docstring dropped TCP.

safl (commented):
Three things inline.

Read, not run: no target was brought up on either provider and no workflow was executed.

cijoe/scripts/nvme_target_probe.py
safl, line 20: `main()` never reads `args.nvme_provider`, so `transports-tcp-linux.yaml` and `transports-tcp-spdk.yaml` pass `nvme-provider` to this step and it does nothing. It predates the rename, but the table in `docs/tutorial/transports/index.md` documents it as one of the flags the three scripts share, so it reads as meaningful. Drop it from this script and from the workflow steps that pass it to it, or use it.

jswaro: Dropping it. Use of this flag should come later, if at all.

docs/tutorial/transports/index.md
jswaro: Added it to the table with a reference to the cijoe default behavior.

docs/tutorial/transports/index.md
```diff
@@ -62,13 +62,14 @@ linux/index
...
-| `--provider`   | `spdk`        | `spdk` or `linux`.                           |
-| `--traddr`     | `127.0.0.1`   | Transport address (IP) for the listener.     |
-| `--trsvcid`    | `4420`        | Transport service id (TCP port).             |
-| `--trtype`     | `tcp`         | Transport type.                              |
-| `--adrfam`     | `ipv4`        | Address family.                              |
+| Argument            | Default     | Description                                                        |
+|---------------------|-------------|--------------------------------------------------------------------|
+| `--nvme-provider`   | `spdk`      | `spdk` or `linux`.                                                 |
+| `--nvme-traddr`     | `127.0.0.1` | Transport address (IP) for the listener.                           |
+| `--nvme-trsvcid`    | `4420`      | Transport service id (TCP port).                                   |
```
safl: `--nvme-trtype` now takes `rdma` as well, so "(TCP port)" is stale; "(port)" covers both.

docs/tutorial/transports/index.md
```diff
@@ -62,13 +62,14 @@ linux/index
...
+| `--nvme-trtype`     | `tcp`       | Transport type.                                                    |
+| `--nvme-adrfam`     | `ipv4`      | Address family.                                                    |
+| `--transport-name`  | None        | CIJOE Transport to use. Defaults to first-found, if not specified. |
```
safl: Trailing whitespace after the closing `|`; `make format-all` should catch it, and the row is wider than the header's last column by the same two characters.

cijoe/scripts/nvme_target_probe.py
```diff
@@ -17,28 +17,28 @@
...
```
safl, line 32: `nvme_target_start.py` gives `--nvme-trtype` `choices=["tcp", "rdma"]`; add the same here so a typo fails at parse time rather than inside `nvme discover`.

safl (commented):
Close, but not mergeable yet: aaae8ba2 ("Remove unused provider reference") has no `Signed-off-by:`. Add it when folding, and fold that commit into a7e39ac2 while at it, since it undoes part of what a7e39ac2 added.

The inline comments are three small ones, a docs row, a docs cell and an argparse `choices`.

On the commit messages, carried over from the last round: a7e39ac2's subject still says "nvme-target workflows", and the commit's larger part is the rename of the options in the three scripts and the tutorial pages; name that instead. On the description: it is a plan written before the work ("Validation Plan", "Requirements"), not a statement of what was run. Replace the validation section with what was actually exercised. CI covers the TCP workflows on both providers; was the RDMA path brought up anywhere? If not, say so in one sentence, that is enough. The Copilot note can go once the section says what was done.

Reviewed with a coding agent. Read the four commits, the resulting scripts, the workflows and the tutorial pages, and checked each open item from the last round against the head; CI is green on it. Nothing built or run; neither the TCP nor the RDMA path was exercised here. Push back where it is wrong.

cijoe/scripts/nvme_target_probe.py
jswaro: Fixed

docs/tutorial/transports/index.md
jswaro: Fixed

docs/tutorial/transports/index.md
jswaro: Fixed.

docs/tutorial/transports/index.md
```diff
@@ -62,13 +62,14 @@ linux/index
...
-| `--provider`   | `spdk`        | `spdk` or `linux`.                           |
-| `--traddr`     | `127.0.0.1`   | Transport address (IP) for the listener.     |
-| `--trsvcid`    | `4420`        | Transport service id (TCP port).             |
-| `--trtype`     | `tcp`         | Transport type.                              |
-| `--adrfam`     | `ipv4`        | Address family.                              |
+| Argument            | Default     | Description                                                        |
+|---------------------|-------------|--------------------------------------------------------------------|
+| `--nvme-provider`   | `spdk`      | `spdk` or `linux`.                                                 |
+| `--nvme-traddr`     | `127.0.0.1` | Transport address (IP) for the listener.                           |
+| `--nvme-trsvcid`    | `4420`      | Transport service id (port).                                   |
```
safl, line 69: This row is a few characters short of the others; padding the description column keeps the table aligned in the source.

safl (commented):
Looks good; two small things.

On the commit subjects: `cijoe(nvmf):` is not one of the Conventional Commits types the history uses; `ci(cijoe):` would fit. Three of the five bodies are empty, and the flag rename is the one where a line on the why would help most.

cijoe/scripts/nvme_target_probe.py
safl, line 32: `choices=["tcp", "rdma"]` was dropped here in this push but kept on the same flag in `nvme_target_start.py:48`, and the description still says the flag is restricted to a choice list. Either restore it here or drop it in `start` as well.

safl (changes requested):
Not mergeable yet, for the reason from 09-08: 59a74252 ("Remove unused provider reference") still has no `Signed-off-by:`, and it is still its own commit rather than folded into 73dc29c0, which it partly undoes. On the commit message of 73dc29c0: no body; it is the flag rename, the one commit whose why (the collision with cijoe's own arguments) is not visible from the diff. The subjects are `ci(cijoe):` now, as asked.

One inline comment, small: `--nvme-trtype` lost its `choices` in `nvme_target_probe.py` this push while `nvme_target_start.py` kept it.

For a follow-up, not here: the description's risk item about external callers of the old flag names is right in principle, but nothing in `cijoe/` or `docs/` still uses them, so a line saying so would spare the next reader the search.

Read, not run: the full diff against `main` and the delta since the 09-10 review, a grep of the head for the old flag names and for every caller of the three scripts, and each commit's trailers. No target was brought up on either provider and no workflow was executed; the RDMA path is untouched.

Assisted-by: Claude Code:claude-fable-5-1

cijoe/scripts/nvme_target_probe.py
jswaro: I'll fix that. Something with my rebase process isn't working as expected.

safl (dismissed):
The scripts, workflows and per-script tables are right and the history is folded; approving. What is left is wording in docs/tutorial/transports/index.md: the prose above the tables still says all three scripts are switched via --nvme-provider (probe no longer takes it), one line is unwrapped, and --nvme-trsvcid is described as a TCP port now that rdma is accepted. Rather than another round of comments, the edits are in one commit you can cherry-pick and squash into 943a0eef: ceb9650b on safl/xNVMe, branch nvmf-scripts-args-docs (`git fetch https://github.com/safl/xNVMe.git nvmf-scripts-args-docs && git cherry-pick ceb9650b`). Take it under your own sign-off.

Read, not run: the diff since the previous head (docs only), each table against the scripts' add_args(), the folded commits against the previous four. No target was brought up and no workflow executed.

Assisted-by: Claude Code:claude-fable-5-1

safl (approved):
lgtm

karlowich (approved):
Great!

I rebased it on the latest main. As soon as the CI is finished this can go in.

PR #743, opened 2026-08-21
homi: a non-invasive status probe, plus the docs and systemd unit to run homi as a service and a fix

safl:
Asking whether a HOMI primary is up meant `pgrep`, which says a process exists
but not that its controllers are open, or attaching, which allocates a heap of
its own, needs the hugepages and driver bindings a monitoring tool has no
business requiring, and increments the very refcount being reported. This adds a
probe that reads uPCIe's shared bookkeeping directly, `homi status` to format
it, and what is needed to run a primary as a service: a systemd unit, its
install, and the operator documentation.

`homi` remains a plain binary. `homi start` is untouched, `status` is added
alongside it, and the unit is additive and optional; `-Dsystemd=disabled` keeps
a build from writing to a machine's unit directory at all.

## Worth a careful look

Role election moves from `flock` to OFD locks, because `flock` cannot be tested
without being taken, so a probe would briefly hold the lock a process elects on.
Both segments gain a magic and version stamp, checked on attach as well as on
probe, since this changes their layout.

Readers take no lock. The controller count is stored with release ordering after
the identifier it covers, and decremented before entries move, so a reader
cannot count a slot being overwritten. It can still catch an identifier
mid-rewrite, which then fails to open and is reported unreadable rather than
trusted.

Underneath sit two changes that stand alone. A reap loop was bounded by a bitmap
dimension rather than the range the allocator admits, so its counter wrapped and
left a process unkillable during teardown. And `xnvme_adm_gfeat_nqueues()` puts
the NSQA and NCQA conversion in one place; it is a helper rather than a
one-to-one mapping of the command, which its docstring says explicitly.

## Verification

CI covers this: the IOMMU-disabled `pcie` job runs `test_mproc` with `--shm_id`,
which selects uPCIe against emulated NVMe on `uio_pci_generic`. The
IOMMU-enabled job excludes uPCIe from that step, since multi-process mode
requires `uio_pci_generic`.

The tests added here cover what the probe reports in each state it can be in: a
held runtime, an id nobody claimed, debris left by a primary that was killed,
and a segment this build cannot read. Two guard against a plausible-looking
constant, since the count has to move with what a secondary asks for and the
totals have to match what the controller answers for itself. One bounds how long
a primary takes to stop, which is what the reap-loop wrap broke. All nine pass
on warp, 9x NVMe on `uio_pci_generic`.

Not exercised anywhere: the unit under systemd itself, only the probe it gates
on.


lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -322,8 +322,13 @@ xnvme_be_upcie_mproc_free_all_queues(struct xnvme_be_upcie_ctrlr *ctrlr)
...
 	}
 
-	for (uint16_t qid = 1; qid < NVME_QID_BITMAP_WORDS * BITS_PER_WORD; qid++) {
-		if (nvme_qid_is_allocated(shm->ctrl.qids, qid)) {
+	/* Counted in a wider type than the qid itself: the bitmap spans
+	 * NVME_QID_BITMAP_WORDS * BITS_PER_WORD entries, which can exceed what
+	 * a uint16_t holds, and such a counter would then never reach the bound
+	 * and would wrap instead. Stop at the highest qid the allocator admits,
+	 * which is what the bitmap exists to track. */
+	for (uint32_t qid = 1; qid < NVME_QID_MAX; ++qid) {
```
naddinadja: Without the history of changing from `NVME_QID_BITMAP_WORDS * BITS_PER_WORD` to `NVME_QID_MAX`, this comment doesn't make a lot of sense. I think it makes more sense to have such information in the commit-message, and inline comments should imo describe the current state and not why it differs from a previous state :)

lib/upcie/xnvme_be_upcie_dev.c
```diff
@@ -590,6 +590,45 @@ xnvme_be_upcie_dev_close(struct xnvme_dev *XNVME_UNUSED(dev))
...
+		return;
+	}
+
+	/* Only whoever brought this controller up may issue admin commands, and
+	 * the per-controller lock says who that is; an attaching secondary
+	 * leaves the descriptor at -1. Tighter than the runtime's primary role,
+	 * which is about the shm_id group rather than this controller. */
+	if (state->ctrlr->mproc.lock_fd < 0) {
+		return;
+	}
```
naddinadja: The comment in this function seem pretty overkill - code seems pretty self explanatory

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -2,16 +2,54 @@
...
+ * Both of the runtime's locks use this, the per-runtime one that elects the
+ * role and the per-controller one that says who owns a device, so there is one
+ * locking mechanism to reason about rather than two with different semantics.
+ *
+ * Returns 1 when F_OFD_GETLK finds the lock held, 0 on success, negative errno
+ * on failure. The lock is released when the descriptor is closed.
+ */
+static int
+_rte_lock_op(int fd, int cmd, short type)
+{
```
naddinadja, line 24: Doc-strings again: a very long explanation for choice of lock in the docstring for this internal helper-function. I feel like this choice is better documented elsewhere (e.g. in commit messages or docs)

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -2,16 +2,54 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
+/* Also needed by the stubs below, which report -ENOSYS where multi-process
+ * mode is unavailable; only Linux picks it up transitively. */
+#include <errno.h>
```
naddinadja: I don't think we need to document why we include errno.h

lib/upcie/xnvme_be_upcie_mproc.c
```diff
@@ -322,8 +322,13 @@ xnvme_be_upcie_mproc_free_all_queues(struct xnvme_be_upcie_ctrlr *ctrlr)
 		return;
 	}
 
-	for (uint16_t qid = 1; qid < NVME_QID_BITMAP_WORDS * BITS_PER_WORD; qid++) {
-		if (nvme_qid_is_allocated(shm->ctrl.qids, qid)) {
+	/* Counted in a wider type than the qid itself: the bitmap spans
+	 * NVME_QID_BITMAP_WORDS * BITS_PER_WORD entries, which can exceed what
+	 * a uint16_t holds, and such a counter would then never reach the bound
+	 * and would wrap instead. Stop at the highest qid the allocator admits,
+	 * which is what the bitmap exists to track. */
```
karlowich: This comment adds nothing in the context of the code.

lib/upcie/xnvme_be_upcie_mproc.c
karlowich: There seems to be a bit of a "double fix" going on. The `uint32_t` change is not needed when switching to `qid < NVME_QID_MAX`.
`NVME_QID_MAX` is `0xFFFF` which is `65535`. The max value of `uint16_t` is `65535`

include/xnvme_dev.h
```diff
@@ -47,4 +47,26 @@ xnvme_dev_alloc(struct xnvme_dev **dev);
...
+ * @param dev Device handle obtained with xnvme_dev_open()
+ * @param nsq Filled with NSQA + 1, the number of I/O submission queues
+ * @param ncq Filled with NCQA + 1, the number of I/O completion queues
+ *
+ * @return On success, 0 is returned and both counts are filled. On error,
+ *         negative errno is returned and neither is written; zero would be a
+ *         plausible answer rather than an absent one.
+ */
+int
+xnvme_dev_nqueues(struct xnvme_dev *dev, uint32_t *nsq, uint32_t *ncq);
```
karlowich: I don't think this is placed correctly in the API. 
We already have `xnvme_adm_gfeat()`, so to me this should be `xnvme_adm_gfeat_nqueues()`.
That follows the pattern of `xnvme_adm_idfy()` vs `xnvme_adm_idfy_ctrlr()` etc.

This also means we can drop this paragraph:
"Asks rather than requests: Set Features could negotiate a larger allocation,
 but a controller may refuse. Costs an admin command, and the answer does not
 change while the controller is up."

karlowich (changes requested):
This PR is doing too much and the framing is wrong in my opinion.
The title "Inspect a multi-process runtime without attaching to it" is solving a problem that the PR introduces. 
The PR introduces a huge "undercover" change in the middle which is moving from a regular binary to a systemd service. Thus, creating the problem that the PR is solving.
I think this should be broken into multiple PRs, so we can discuss the large changes individually.

(Also a few inline comments, which may or may not be relevant depending on the direction we choose)

tools/homi.c
```diff
@@ -156,6 +158,146 @@ sub_start(struct xnvme_cli *XNVME_UNUSED(cli))
...
+	return (int)rte.nctrlrs;
+}
+
+static int
+sub_status(struct xnvme_cli *cli)
+{
+	const int shm_id = (int)cli->args.shm_id;
+	int running, held, all_up = 0;
+
+	running = xnvme_mproc_primary_alive((uint32_t)shm_id);
```
naddinadja: Why are we casting the `shm_id` back and forth?
I see that the `cli->args.shm_id` is an `uint64_t`, while `xnvme_opts.shm_id` is an `uint32_t` (woopsie, that's my fault). We should probably streamline this to only `uint32_t`, and then I would not cast it to an `int` here. You only really use it as int for printing once (which can also be done with PRIu32)

naddinadja (changes requested):
I have a couple comments about code-comments/doc-strings. Inline-documentation can be helpful, especially if there something very important to be aware of when inspecting and modifying the code, and also if the code is quite complex. However, I often experience that Claude inserts commentary which is way above what is necessary. 

I see that in commit `2f15a3a` ("refactor/be/upcie): name a controller...") you shorten a lot of the comments, which is great, but maybe not related to the actual commit :)

lib/upcie/xnvme_be_upcie_mproc.c
safl: ack

lib/upcie/xnvme_be_upcie_dev.c
safl: ack

lib/upcie/xnvme_be_upcie_mproc.c
safl: ack

lib/upcie/xnvme_be_upcie_mproc.c
safl: ack

lib/upcie/xnvme_be_upcie_mproc.c
safl: ack

lib/upcie/xnvme_be_upcie_mproc.c
safl: ack

include/xnvme_dev.h
safl: ack

tools/homi.c
safl: ack

PR #745, opened 2026-08-22
Register application-owned memory: registry-backed dmamem and mem_map

safl:
A dmamem describes one contiguous range and resolves an offset from a base,
which suits a heap uPCIe allocated and nothing else. `xnvme_mem_map()` cannot
be built on that: the memory belongs to the caller, there are arbitrarily many
regions, and the command paths must not branch on which kind they were handed.
The registry indexes granules absolutely, so one dmamem spans the heap and
every registered region at once. That is what a consumer holding its own GPU
tensors needs, llama.cpp's loader being the one driving this.

Supersedes #663 by @fmma and builds on it: that PR's `mem_map`/`mem_unmap` and
its CUDA implementation carry over, and what changes is the mechanism beneath
them, so the changes land in the generic `xnvme_be_upcie_{admin,sync,async}.c`
rather than the CUDA-specific ones and HIP comes along at no extra cost.
`examples/xnvme_cuda_mem_map.cu` carries over unchanged and keeps its
authorship.

Vendors uPCIe at v0.7.0, byte-identical to the tip of the stack under review
there: safl/upcie#58, safl/upcie#59 and safl/upcie#60.

The vendored import is its own leading commit, as asked. It does not compile on
its own: the drop removes `dmamem_from_hostmem_lut()` and its CUDA and HIP
counterparts, which the backend still calls until the next commit ports it. That
is deliberate. Keeping them together would mean reviewing a 50-file vendored
import and a backend rewrite as a single diff, and a reviewable branch is worth
more here than a bisectable one. Exactly one commit is affected; the next
restores the build and every commit after it is green.

## Worth a careful look

**Whether the registry-backed translation still does what #663 intended for the
CUDA path**, since the mechanism under a working implementation was replaced.

**The reservation.** Sizing the table by the address space rather than by one
region costs 512 MiB of virtual reservation at a 2 MiB granule. It is
demand-paged, but `ulimit -v` refuses it and `vm.overcommit_memory=2` charges
it whatever `MAP_NORESERVE` says, so `XNVME_UPCIE_VA_BITS` bounds it. Lowering
it narrows what can be registered.

**Where the split runs.** The first commit moves onto the registry and the
second adds `mem_map`/`mem_unmap` to the CUDA and HIP backends, which until
then keep the `nosys` slots they have on `main`. The vendored headers cannot be
separated from the port that follows them: the vendored change removes
`dmamem_from_hostmem_lut()`, `dmamem_from_cuda_lut()`, `dmamem_from_hip_lut()`,
`phys_lut` and `hugepgsz`, all of which the old backend uses, so either half
alone fails to compile. Every commit here was built on both GPU boxes to show
the ones that can be separated are.

## Verification

CI covers the host path. `test_upcie` runs the backend against emulated NVMe on
`uio_pci_generic`, and picks up the new `test_mptr_unregistered`, which checks
that metadata memory the registry has no record of is refused rather than
submitted. That case writes rather than reads on purpose: both take the same
path, but without the check a write has the controller read physical address
zero where a read has it write there. It skips itself behind an IOMMU, where
the mapping supplies the address and there is nothing to fail to resolve.

What CI has no toolchain for is the GPU backends, which is most of this change.
Run by hand against real controllers instead, on this branch, every exit code
asserted:

- warp, NVIDIA RTX A6000, CUDA, controller on `uio_pci_generic`:
  `xnvmeperf verify --be upcie` and `--be upcie-cuda` both 64 IOs with 0
  mismatches, `xnvmeperf cuda-verify` 8 GPU-submitted IOs with 0 mismatches,
  `xnvme_cuda_mem_map` 64 GPU-submitted reads into a `mem_map`'d `cudaMalloc`
  buffer, and `xnvme_tests_map mptr_unregistered` refused with `-EINVAL`.
- wave, AMD with ROCm, same shape: `xnvmeperf verify --be upcie` and `--be
  upcie-hip` both 64 IOs with 0 mismatches, `xnvme_hip_mem_map` reading into a
  `mem_map`'d `hipMalloc` buffer, and `mptr_unregistered` refused with
  `-EINVAL`.

Not exercised anywhere: the GPU paths under CI, since no runner has a GPU or
the toolchains; GPU-initiated submission on ROCm, since `xnvmeperf` has no
`hip-verify` counterpart to `cuda-verify`; `mem_map` for host memory, which
stays `nosys` on plain `upcie`; and Windows, which keeps the `-ENOTSUP` stub.


karlowich (changes requested):
Why is this based on #743?
The two are targeting different parts of the uPCIe backend. Are they in conflict?
Also, I would prefer to split "feat(be/upcie): move onto the registry-backed dmamem, and add mem_map" into two. The vendored uPCIe changes should be separate from the changes to xNVMe.

lib/upcie/xnvme_be_upcie.h
```diff
@@ -216,6 +216,15 @@ extern struct xnvme_be_async g_xnvme_be_upcie_async;
...
 
 // Attachment: driver probing, mode selection and vfio wiring (xnvme_be_upcie_vfio.c)
+/**
+ * Address-space width the DMA-address table is sized for; 0 for the default.
+ *
+ * Read from XNVME_UPCIE_VA_BITS. See the definition for what it costs and when
+ * lowering it is warranted.
+ */
+int
+xnvme_be_upcie_va_bits(void);
```
fmma: This declaration should be moved to another line. It currently sits under a `xnvme_be_upcie_vfio.c` comment, but its implementation lives in `xnvme_be_upcie_dev.c`. I would move it next to the other functions implemented in `_dev.c` a bit further down.

fmma (dismissed):
LGTM with some small comments. Verified with opends.

There is some documentation drift: `docs/background/backends/upcie/cuda.md` says `mem_map` / `mem_unmap` returns ENOSYS which is no longer the case.

lib/upcie/xnvme_be_upcie.h
safl: ack

lib/upcie/xnvme_be_upcie_cuda_mem.c
```diff
@@ -46,6 +46,43 @@ xnvme_be_upcie_cuda_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void
...
 
+/**
+ * Register a caller-allocated device buffer for DMA
+ *
+ * Resolves through the same registry as the heap, so a buffer handed over here
+ * is usable exactly as one from xnvme_buf_alloc(). Registering the same range
+ * twice is cheap: the chunks it covers are refcounted.
+ */
+int
+xnvme_be_upcie_cuda_mem_map(const struct xnvme_dev *XNVME_UNUSED(dev), void *vaddr, size_t nbytes,
```
karlowich, line 57: How about adding a `memmap` function to the `upcie` backend as well? Perhaps it can be done in a general way, so we don't have to duplicate the code 3 times. Doing this would allow us to leverage the xNVMe memmap/unmap tests that we already have.

cijoe/tests/logic/test_xnvme_tests_map.py
```diff
@@ -10,3 +10,31 @@ def test_mem_map_unmap(cijoe, device, be_opts, cli_args):
 
     err, _ = cijoe.run(f"xnvme_tests_map mem_map_unmap {cli_args} --count 31")
     assert not err
+
+
+@xnvme_parametrize(labels=["pcie"], opts=["be"])
+def test_mptr_unregistered(cijoe, device, be_opts, cli_args):
```
karlowich: Isn't this a broader thing than just `mptr` and `upcie`?
How does SPDK react if you give it "unknown" memory?
I think this is a good test that's a little too specific to the uPCIe implementation.

karlowich (changes requested):
This looks good.
I have two suggestions inline.

I'm unsure if the examples introduced really give us anything new.
The original `xnvme_cuda` example shows how to use xNVMe in a CUDA Kernel.
I don't think we need examples of how to use the `xnvme_memmap()` API.

lib/upcie/xnvme_be_upcie_cuda_mem.c
safl: ack

cijoe/tests/logic/test_xnvme_tests_map.py
safl: Not yet: I have not measured what SPDK does with unregistered memory, so I do not know whether the assertion generalises or whether each backend needs its own. Leaving the case uPCIe-specific until that is measured rather than widening it on a guess.

safl: Measured on warp, which changes my previous answer: SPDK refuses it too.

| backend | registered mbuf | unregistered mbuf |
|---|---|---|
| uPCIe, uio_pci_generic | accepted | refused, `-22` EINVAL |
| SPDK, vfio-pci | accepted | refused, `-5` EIO |

So the property is not uPCIe-specific and the case can be broadened, as you suspected.

Getting there turned up something about the test itself. It reports success whenever the write is refused, and warp's namespace has `nbytes_oob: 0`, so a write carrying an mptr could be refused for having no metadata at all rather than for the buffer being unregistered. On that reading both results were worthless. The table above is from a control I added in 54eacc7d: the same write with a registered mbuf, which is accepted on both backends, so registration really is what the refusal is about. Where the control is refused too, both cases now report themselves inconclusive and print the namespace's out-of-band size rather than passing quietly.

I have left the parametrisation alone for now, since widening it is your call on which backends should be held to this.

fmma (dismissed):

lib/upcie/xnvme_be_upcie_mem.c
```diff
@@ -64,6 +64,95 @@ xnvme_be_upcie_buf_vtophys(const struct xnvme_dev *XNVME_UNUSED(dev), void *buf,
...
+ * Unregister what xnvme_be_upcie_mem_map() registered
+ *
+ * @return 0 on success, negative errno on failure.
+ */
+static int
+xnvme_be_upcie_mem_unmap(const struct xnvme_dev *XNVME_UNUSED(dev), void *XNVME_UNUSED(vaddr))
+{
+	/* iommufd unmaps by IOVA and length, neither of which a caller handing
+	 * back an address has. */
+	return -ENOSYS;
```
karlowich: Hmm. In this state, the hostmem implementation doesn't give us much, as the memmap test would fail.
I didn't consider that host memory works differently than the device memory equivalents.
How about returning success on unmap with a debug message explaining that unmap is a no-op?
That way we can enable the `test_mem_map_unmap` for upcie.

karlowich: Alternatively, I propose dropping the hostmem part until it is integrated with dmamem and can use the same API

tests/map.c
```diff
@@ -80,6 +80,55 @@ test_mem_map_unmap(struct xnvme_cli *cli)
...
  */
+/**
+ * Control for mptr_unregistered: the same write, with a registered mbuf
+ *
+ * The negative case reports success whenever the write is refused, which a
+ * namespace carrying no metadata would also do, on any backend. Pairing it with
+ * this says which of the two the refusal was about.
+ */
+static int
+test_mptr_registered(struct xnvme_cli *cli)
```
karlowich: Let's test this in "test_xnvme_tests_map.py"

karlowich: To follow your earlier comment, perhaps the test should call both `mptr_registered` and `mptr_unregistered` in the same test to weed out false positives

cijoe/tests/logic/test_xnvme_tests_map.py
```diff
@@ -10,3 +10,31 @@ def test_mem_map_unmap(cijoe, device, be_opts, cli_args):
...
+    # The registry translates only where the device consumes physical
+    # addresses. Behind an IOMMU the address comes from the mapping instead, so
+    # there is nothing for an unregistered buffer to fail to resolve to and the
+    # check deliberately does not apply.
+    uri = device["uri"]
+    err, state = cijoe.run(
+        f"sh -c 'basename $(readlink -f /sys/bus/pci/devices/{uri}/driver)'"
+    )
+    if err or "uio_pci_generic" not in state.output():
+        pytest.skip(reason="Requires uio_pci_generic; with vfio the IOMMU translates")
```
karlowich: Let's broaden this. If I'm reading what you wrote in the comment correctly, this should work with anything in `labels=["pcie"]`

karlowich (changes requested):
This is starting to look really good. Just a few comments before the finish line

lib/upcie/xnvme_be_upcie.h
safl: Moved. It now sits under its own `xnvme_be_upcie_dev.c` heading, which is where it is implemented.

lib/upcie/xnvme_be_upcie_cuda_mem.c
safl: Done, in one implementation shared by both GPU backends: they differ only in which dmamem the memory belongs to. See the sibling thread on `xnvme_be_upcie_mem.c` for why the host backend does not get one.

lib/upcie/xnvme_be_upcie_mem.c
safl: Agreed, so I took your alternative: the host `mem_map`/`mem_unmap` are back to `xnvme_be_nosys_*` until hostmem shares the registry-backed path. The host heap is adopted rather than discovered (`dmamem_from_hostmem_registry` passes a NULL populate callback), so mapping caller memory would mean reading /proc/self/pagemap, which it does not do. Returning success from a no-op unmap seemed the worse of the two. `host.md` and the `xnvme_mem_map()` docstring now say so.

cijoe/tests/logic/test_xnvme_tests_map.py
safl: Both of these are now one test, `test_mptr_registration`, parametrised on `labels=["pcie"]` rather than uPCIe only. It runs `mptr_registered` first and `mptr_unregistered` second, since neither says anything alone: a namespace with no out-of-band space refuses a write carrying an mptr whatever the buffer is, so the refusal only means something once the control is accepted. The C test already signals that case as INCONCLUSIVE, and the python test skips on it. Both subcommands also needed `XNVME_CLI_ASYNC_OPTS` rather than `XNVME_CLI_ADMIN_OPTS`; they do an I/O write, so they were rejecting the `--sync`/`--async` that cijoe passes.

The uPCIe-only skip that remains is narrower: uPCIe translates through the registry only where the device consumes physical addresses, so behind an IOMMU there is nothing for an unregistered buffer to fail to resolve to. Other backends are asked regardless.

On my guest only the uPCIe permutation actually ran and passed; spdk and libvfn were not exercised, because that guest has no vIOMMU (`/sys/kernel/iommu_groups` is empty) so neither backend can open a device there at all. Those permutations are left to CI.

cijoe/tests/logic/test_xnvme_tests_map.py
safl: Both of these are now one test, `test_mptr_registration`, parametrised on `labels=["pcie"]` rather than uPCIe only. It runs `mptr_registered` first and `mptr_unregistered` second, since neither says anything alone: a namespace with no out-of-band space refuses a write carrying an mptr whatever the buffer is, so the refusal only means something once the control is accepted. The C test already signals that case as INCONCLUSIVE, and the python test skips on it. Both subcommands also needed `XNVME_CLI_ASYNC_OPTS` rather than `XNVME_CLI_ADMIN_OPTS`; they do an I/O write, so they were rejecting the `--sync`/`--async` that cijoe passes.

The uPCIe-only skip that remains is narrower: uPCIe translates through the registry only where the device consumes physical addresses, so behind an IOMMU there is nothing for an unregistered buffer to fail to resolve to. Other backends are asked regardless.

tests/map.c
safl: Done. `mptr_registered` is now called from `test_mptr_registration` in `test_xnvme_tests_map.py`, in the same test as `mptr_unregistered` so the pair weeds out the false positive you were pointing at.

include/libxnvme_mem.h
```diff
@@ -13,7 +13,9 @@
  * command payloads
  *
  * @note nbytes must be greater than zero and a multiple of minimal granularity
- * @note Unmap the buffer using xnvme_mem_unmap()
+ * @note Unmap the buffer using xnvme_mem_unmap(), where the backend implements
+ * it. Not every backend does; -ENOSYS says so rather than that the call was
+ * malformed.
```
karlowich: This is outdated

include/libxnvme_mem.h
```diff
@@ -27,6 +29,9 @@ xnvme_mem_map(const struct xnvme_dev *dev, void *vaddr, size_t nbytes);
 /**
  * Unmap the given IO buffer mapped with xnvme_mem_map()
  *
+ * @note Not every backend implements this; -ENOSYS says the mapping cannot be
+ * taken back, not that the call was malformed.
+ *
```
karlowich: This is outdated

docs/background/backends/upcie/host.md
```diff
@@ -30,7 +30,10 @@ DMA buffer space needed by the application.
 
 - **No `buf_realloc`.** Buffer reallocation is not implemented and returns
   `ENOSYS`.
-- **No memory mapping** (`mem_map` / `mem_unmap`). These return `ENOSYS`.
+- **No memory mapping** (`mem_map` / `mem_unmap`). These return `ENOSYS`. The
+  host heap is adopted rather than discovered, so there is no way to make
+  memory the caller allocated addressable without reading pagemap; the GPU
+  backends can, and do.
```
karlowich: I think we can use the old version of this text. The description here is very close uPCIe implementation details, I don't think the xNVMe docs reader will know what "adopted rather than discovered" means.

karlowich (changes requested):
Let's squash "test(map): add the control the unregistered-mbuf case was missing" into "test(be/upcie): refuse metadata memory the registry has no record of"

include/libxnvme_mem.h
safl: Right, both were written for the state where the host backend had `mem_map` but not `mem_unmap`, and that asymmetry went away when I dropped the host `mem_map`. Reverted; `include/libxnvme_mem.h` is back to what it says on `main`, so this branch no longer touches it.

include/libxnvme_mem.h
safl: Right, both were written for the state where the host backend had `mem_map` but not `mem_unmap`, and that asymmetry went away when I dropped the host `mem_map`. Reverted; `include/libxnvme_mem.h` is back to what it says on `main`, so this branch no longer touches it.

docs/background/backends/upcie/host.md
safl: Agreed, and the old text is now correct again anyway, since the host backend no longer implements either call. Reverted to `main`'s wording, so `host.md` is untouched by this branch. The "adopted rather than discovered" reasoning stays in the commit message, where a reader is looking at the uPCIe internals rather than at the backend docs.

tests/map.c
```diff
@@ -68,6 +70,106 @@ test_mem_map_unmap(struct xnvme_cli *cli)
...
+ *
+ * Memory outside the registry translates to zero, so submitting anyway points
+ * the controller at physical address zero, which without an IOMMU is host
+ * memory it may write.
+ *
+ * Writes rather than reads on purpose: without the check a write has the
+ * controller read address zero where a read has it write there.
+ */
+/**
+ * Control for mptr_unregistered: the same write, with a registered mbuf
```
karlowich: I see this `test_mptr_registered()` is really generic and the subcommand has no relation `tests/map.c` outside of being the contol. Perhaps the two functions should be merged in the C code? Given the amount of boilerplate, it looks like we will actually have fewer lines of code as a result. That will also give a more straight forward python test on top.

karlowich (changes requested):

tests/map.c
safl: Merged, and you were right about the line count: 24 fewer lines across the two files. The shared part is now `mptr_write()`, which allocates the dbuf, submits the write with whatever mbuf it is handed, and reports accepted or refused; the two cases differ only in where the mbuf came from. One subcommand, `mptr_registration`, runs the registered write first and the unregistered one second, so the python test is a single invocation with the INCONCLUSIVE skip and an assert.

It also turned up a real bug in the process: the squash had left two stacked docstrings above `test_mptr_registered`, the first of which described the unregistered case. Doxygen would have attached the wrong one. There is now one docstring on the merged function.

Verified on a QEMU guest: the uPCIe permutation passes, and the output shows both halves ran, "accepted with a registered mbuf, refused with err(-22)". spdk and libvfn still fail there at `xnvme_dev_open()` with -6, unchanged by this and not caused by it, because that guest has no vIOMMU; those permutations are left to CI.

karlowich (approved):
GREAT!

PR #747, opened 2026-08-25
feat(be/linux): NVMe FLUSH and FUA support in the io_uring and libaio backends

yonggilsong:
The opcode dispatch in the io_uring and libaio async backends handled only READ and WRITE; FLUSH fell through to the default case and was rejected with -ENOSYS. This breaks any consumer whose device reports a volatile write cache, e.g. serving a namespace through ublk makes the kernel issue FLUSH, and mkfs on top of it fails with EIO. Similarly, the FUA bit set on a write command was silently ignored.

Scoped down per review to the FLUSH/FUA work and its tests; the SQPOLL fix with its regression test, the sqe-leak fix in cmd_iov, the poll_sq/poll_io options and the qdepth fix move to a follow-up PR. Four commits:

1. **feat(be/linux): support NVMe FLUSH in the io_uring and libaio backends** — FLUSH → `IORING_OP_FSYNC` / `IO_CMD_FSYNC` in the single-buffer and vectored paths; full fsync, matching psync. On an IOPOLL ring FLUSH is rejected at submission-time with -ENOSYS (IORING_OP_FSYNC has no iopoll-handler), matching the pre-change behaviour, and before acquiring an sqe.
2. **feat(be/linux): honour the FUA bit in io_uring and libaio writes** — `ctx->cmd.nvm.fua` on an NVM write → `RWF_DSYNC`; on a raw block device the kernel then issues the write with REQ_FUA. NVM write opcode only: 'fua' is an NVMe command-field, the FS opcodes define no such field and the file-API has no way of setting it. The header documents that honouring 'fua' is backend-dependent and that unset fields survive context reuse.
3. **tests(ioworker): add a verify-flush subcommand** — FUA-write, FLUSH through the queue, read-compare; the FLUSH goes through the vectored submission-path when `--vec-cnt` is given, so both `cmd_io` and `cmd_iov` dispatches are exercised. Meson entries with `--async emu` walk a real opcode dispatch (the plain `1GB` entry resolves to the nil backend, which ignores the opcode).
4. **tests(cijoe): cover NVM FLUSH through the async backends in ioworker** — verify-flush across the backend matrix, skipping the async-interfaces without an NVM FLUSH implementation (posix, kqueue, iocp, iocp_th).

## Verification

On Ubuntu 24.04, kernel 6.8.12, against a file-backed target, the ramdisk backend, and a Samsung SSD 970 EVO 1TB (as root):

- `meson test` — pass/fail set identical to `main` on the same machine.
- Every commit in the series builds.
- Queue-level smoke driving write, FUA-write and FLUSH through io_uring and libaio (single-buffer and vectored): all complete without error.
- `xnvme_tests_ioworker verify-flush 1GB --async emu|thrpool`, with `--vec-cnt 4`: pass.
- Real-NVMe: `verify-flush` over io_uring, libaio and io_uring_cmd, plus vectored variants: pass (detail in the comments below).

Known gap, deferred: vectored READ/WRITE on an SQPOLL queue still loses `IOSQE_FIXED_FILE` through the prep-helpers — that is the pre-existing bug whose fix moved to the follow-up PR; the FLUSH branch added here re-assigns `sqe->flags` after its prep-helper and is unaffected.

### DIFF
--- a/cijoe/tests/logic/test_xnvme_tests_ioworker.py
+++ b/cijoe/tests/logic/test_xnvme_tests_ioworker.py
@@ -195,3 +195,23 @@ def test_verify_sync_iovec_direct(cijoe, device, be_opts, cli_args):
         f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 7 --vec-cnt 4 --direct 1"
     )
     assert not err
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
+def test_verify_flush(cijoe, device, be_opts, cli_args):
+    # Async-interfaces without an implementation of NVM FLUSH: FreeBSD 'kqueue'
+    # and Windows 'ioring' reject it with -ENOSYS, Windows 'iocp'/'iocp_th'
+    # handle only the FS FLUSH opcode
+    if be_opts["async"] in ["kqueue", "iocp", "iocp_th", "ioring"]:
+        pytest.skip(reason=f"[async={be_opts['async']}] does not implement NVM FLUSH")
+
+    # The FUA-writes make this the first fabrics case sustaining I/O beyond the
+    # keep-alive timeout, and the spdk backend only services the admin-queue --
+    # and thereby keep-alive -- inside admin commands, so the target reaps the
+    # controller mid-run and the queue dies with -ENXIO. A backend concern, not
+    # a FLUSH one; skip until the backend keeps the connection alive
+    if "fabrics" in device["labels"]:
+        pytest.skip(reason="[fabrics] spdk backend sends no keep-alive during I/O")
+
+    err, _ = cijoe.run(f"xnvme_tests_ioworker verify-flush {cli_args}")
+    assert not err
--- a/include/libxnvme_nvm.h
+++ b/include/libxnvme_nvm.h
@@ -34,6 +34,20 @@ xnvme_nvm_read(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t slba, uint16_t
 /**
  * Submit a NVMe Write command
  *
+ * NOTE: fields which are not arguments here, such as 'ctx->cmd.nvm.fua', are
+ * left untouched; set them on the context before calling this. Neither this
+ * function nor ::xnvme_queue_get_cmd_ctx clears them, so a context which is
+ * reused for another command carries them over. Reset it with
+ * ::xnvme_cmd_ctx_clear when that is not intended.
+ *
+ * NOTE: whether 'ctx->cmd.nvm.fua' is honoured is backend-dependent, and it
+ * applies to NVM writes only. Backends passing the command through to an NVMe
+ * device (ioctl, io_uring_cmd, spdk, libvfn, upcie) carry the bit in the
+ * command itself. The Linux io_uring and libaio backends translate it to
+ * RWF_DSYNC, which on a raw block device makes the kernel issue the write
+ * with REQ_FUA. The remaining backends (psync, POSIX-aio, ramdisk, FreeBSD,
+ * Windows, macOS) ignore it.
+ *
  * @param ctx Pointer to command context (::xnvme_cmd_ctx)
  * @param nsid Namespace Identifier
  * @param slba The LBA to start the write at
@@ -80,6 +94,10 @@ xnvme_nvm_write_zeroes(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t slba, 
  *
  * This populates the command fields in the given context, it does not submit
  * the command. Submit it with e.g. ::xnvme_cmd_pass.
+ *
+ * NOTE: only the fields given as arguments are populated; the remaining ones,
+ * such as 'ctx->cmd.nvm.fua', are left as they were. Reset a reused context
+ * with ::xnvme_cmd_ctx_clear when carrying them over is not intended.
  *
  * @param ctx Pointer to command context (::xnvme_cmd_ctx)
  * @param opcode Opcode for the NVMe command
--- a/lib/xnvme_be_linux_async_libaio.c
+++ b/lib/xnvme_be_linux_async_libaio.c
@@ -11,6 +11,7 @@
 #ifdef XNVME_BE_LINUX_LIBAIO_ENABLED
 #include <errno.h>
 #include <libaio.h>
+#include <linux/fs.h>
 #include <stdatomic.h>
 #include <sys/eventfd.h>
 #include <unistd.h>
@@ -199,6 +200,13 @@ _linux_libaio_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes, 
 		return -ENOSYS;
 	}
 
+	// Honour the FUA bit of an NVM write; on a raw block device RWF_DSYNC
+	// makes the kernel issue the write with REQ_FUA. Only for the NVM opcode:
+	// 'fua' is an NVMe command-field, it does not exist for the FS opcodes
+	if (ctx->cmd.nvm.fua && ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_WRITE) {
+		iocb.aio_rw_flags = RWF_DSYNC;
+	}
+
 	iocb.data = (unsigned long *)ctx;
 
 	if (queue->efd != -1) {
@@ -265,6 +273,13 @@ _linux_libaio_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, size_t dvec
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d", ctx->cmd.common.opcode);
 		return -ENOSYS;
+	}
+
+	// Honour the FUA bit of an NVM write; on a raw block device RWF_DSYNC
+	// makes the kernel issue the write with REQ_FUA. Only for the NVM opcode:
+	// 'fua' is an NVMe command-field, it does not exist for the FS opcodes
+	if (ctx->cmd.nvm.fua && ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_WRITE) {
+		iocb.aio_rw_flags = RWF_DSYNC;
 	}
 
 	if (queue->efd != -1) {
--- a/lib/xnvme_be_linux_async_libaio.c
+++ b/lib/xnvme_be_linux_async_libaio.c
@@ -188,7 +188,11 @@ _linux_libaio_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes, 
 		io_prep_pread(&iocb, state->fd, dbuf, dbuf_nbytes, ctx->cmd.nvm.slba);
 		break;
 
-		// TODO: determine how to handle fsync
+	case XNVME_SPEC_NVM_OPC_FLUSH:
+	case XNVME_SPEC_FS_OPC_FLUSH:
+		// NOTE: full fsync, matching the psync backend
+		io_prep_fsync(&iocb, state->fd);
+		break;
 
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d", ctx->cmd.common.opcode);
@@ -252,7 +256,11 @@ _linux_libaio_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, size_t dvec
 		io_prep_preadv(&iocb, state->fd, dvec, dvec_cnt, ctx->cmd.nvm.slba);
 		break;
 
-		// TODO: determine how to handle fsync
+	case XNVME_SPEC_NVM_OPC_FLUSH:
+	case XNVME_SPEC_FS_OPC_FLUSH:
+		// NOTE: full fsync, matching the psync backend
+		io_prep_fsync(&iocb, state->fd);
+		break;
 
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d", ctx->cmd.common.opcode);
--- a/lib/xnvme_be_linux_async_liburing.c
+++ b/lib/xnvme_be_linux_async_liburing.c
@@ -12,6 +12,7 @@
 #include <pthread.h>
 #include <errno.h>
 #include <liburing.h>
+#include <linux/fs.h>
 #include <sys/eventfd.h>
 #include <unistd.h>
 #include <xnvme_queue.h>
@@ -338,7 +339,12 @@ xnvme_be_linux_liburing_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbu
 	// NOTE: we only ever register a single file, the raw device, so the
 	// provided index will always be 0
 	sqe->fd = queue->poll_sq ? 0 : state->fd;
-	sqe->rw_flags = 0;
+	// Honour the FUA bit of an NVM write; on a raw block device RWF_DSYNC
+	// makes the kernel issue the write with REQ_FUA. Only for the NVM opcode:
+	// 'fua' is an NVMe command-field, it does not exist for the FS opcodes
+	sqe->rw_flags = (ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_WRITE && ctx->cmd.nvm.fua)
+				? RWF_DSYNC
+				: 0;
 	sqe->user_data = (unsigned long)ctx;
 	// sqe->__pad2[0] = sqe->__pad2[1] = sqe->__pad2[2] = 0;
 
@@ -409,6 +415,13 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
 	/* fall through */
 	case XNVME_SPEC_FS_OPC_WRITE:
 		io_uring_prep_writev(sqe, fd, dvec, dvec_cnt, ctx->cmd.nvm.slba << ssw);
+		if (ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_WRITE && ctx->cmd.nvm.fua) {
+			// Honour the FUA bit of an NVM write; on a raw block device
+			// RWF_DSYNC makes the kernel issue the write with REQ_FUA.
+			// Only for the NVM opcode: 'fua' is an NVMe command-field, it
+			// does not exist for the FS opcodes
+			sqe->rw_flags = RWF_DSYNC;
+		}
 		break;
 
 	case XNVME_SPEC_NVM_OPC_READ:
--- a/lib/xnvme_be_linux_async_liburing.c
+++ b/lib/xnvme_be_linux_async_liburing.c
@@ -294,6 +294,19 @@ xnvme_be_linux_liburing_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbu
 		opcode = IORING_OP_READ;
 		break;
 
+	case XNVME_SPEC_NVM_OPC_FLUSH:
+	case XNVME_SPEC_FS_OPC_FLUSH:
+		// NOTE: IORING_OP_FSYNC has no iopoll-handler; on a ring set up with
+		// IORING_SETUP_IOPOLL it would be accepted here and only fail with
+		// -EINVAL at completion-time. Reject it at submission-time instead
+		if (queue->poll_io) {
+			XNVME_DEBUG("FAILED: FLUSH is not supported on an IOPOLL queue");
+			return -ENOSYS;
+		}
+		// NOTE: full fsync, matching the psync backend
+		opcode = IORING_OP_FSYNC;
+		break;
+
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d for async", ctx->cmd.common.opcode);
 		return -ENOSYS;
@@ -302,6 +315,18 @@ xnvme_be_linux_liburing_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbu
 	sqe = io_uring_get_sqe(&queue->ring);
 	if (!sqe) {
 		return -EAGAIN;
+	}
+
+	if (opcode == IORING_OP_FSYNC) {
+		// NOTE: the prep-helper zeroes 'addr', 'off', 'len' and 'fsync_flags'.
+		// io_fsync_prep() rejects a non-zero 'addr'; 'off'/'len' would select a
+		// range for vfs_fsync_range(), zero/zero means the entire file
+		io_uring_prep_fsync(sqe, queue->poll_sq ? 0 : state->fd, 0);
+		// NOTE: set after the prep-helper; io_uring_prep_rw() zeroes 'flags'
+		sqe->flags = queue->poll_sq ? IOSQE_FIXED_FILE : 0;
+		sqe->user_data = (unsigned long)ctx;
+
+		goto submit;
 	}
 
 	sqe->opcode = opcode;
@@ -316,6 +341,8 @@ xnvme_be_linux_liburing_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbu
 	sqe->rw_flags = 0;
 	sqe->user_data = (unsigned long)ctx;
 	// sqe->__pad2[0] = sqe->__pad2[1] = sqe->__pad2[2] = 0;
+
+submit:
 
 	if (queue->batching) {
 		goto exit;
@@ -355,6 +382,16 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
 		return -ENOTSUP;
 	}
 
+	// NOTE: IORING_OP_FSYNC has no iopoll-handler; on a ring set up with
+	// IORING_SETUP_IOPOLL it would be accepted at submission-time and only fail
+	// with -EINVAL at completion-time. Reject it up-front instead, and do so
+	// before acquiring an sqe, such that none is left dangling in the ring
+	if (queue->poll_io && (ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_FLUSH ||
+			       ctx->cmd.common.opcode == XNVME_SPEC_FS_OPC_FLUSH)) {
+		XNVME_DEBUG("FAILED: FLUSH is not supported on an IOPOLL queue");
+		return -ENOSYS;
+	}
+
 	sqe = io_uring_get_sqe(&queue->ring);
 	if (!sqe) {
 		return -EAGAIN;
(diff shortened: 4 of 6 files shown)


safl (commented):
Action items, detail in the inline comments:

- [x] Add a basic test exercising FLUSH through the async backends (`lib/xnvme_be_linux_async_liburing.c:298`)
- [x] Split the `IOSQE_FIXED_FILE` / SQPOLL fix into its own commit (`lib/xnvme_be_linux_async_liburing.c:441`)
- [x] Say in the header that honouring `fua` is backend-dependent (`include/libxnvme_nvm.h:37`)
- [x] Decide whether `nvm.fua` should be read for the FS write opcodes (`lib/xnvme_be_linux_async_libaio.c:207`)
- [x] State in the description what was run to verify this, and on what
- [x] Optional: the vectored `default:` leaves an SQE consumed but unsubmitted (`lib/xnvme_be_linux_async_liburing.c:434`)

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -294,6 +295,19 @@ xnvme_be_linux_liburing_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbu
 		opcode = IORING_OP_READ;
 		break;
 
+	case XNVME_SPEC_NVM_OPC_FLUSH:
```
safl, line 298: A basic test exercising this path would be highly appreciated. Nothing in `tests/` or `cijoe/tests/` covers NVM FLUSH through an async backend today, so the `-ENOSYS` this fixes would go unnoticed if it came back, here or in the libaio path at `xnvme_be_linux_async_libaio.c:192`.

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -381,11 +424,22 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
...
+
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d for async", ctx->cmd.common.opcode);
 		return -ENOSYS;
 	}
 
+	// NOTE: set after the prep-helpers; io_uring_prep_rw() zeroes 'flags', so
+	// assigning this before the dispatch above would drop IOSQE_FIXED_FILE and
+	// make an SQPOLL queue target the actual fd 0 instead of the registered file
+	sqe->flags = queue->poll_sq ? IOSQE_FIXED_FILE : 0;
```
safl: This line fixes a distinct pre-existing bug, and it is folded into a FLUSH feature commit. `io_uring_prep_rw()` zeroes `sqe->flags`, so the old assignment before the prep helpers dropped `IOSQE_FIXED_FILE`, and a vectored submission on an SQPOLL queue targeted real file descriptor 0 instead of the registered file. Worth its own commit so `git log` can find it, and it needs its own test, since nothing here covers SQPOLL with vectored I/O.

include/libxnvme_nvm.h
```diff
@@ -34,6 +34,12 @@ xnvme_nvm_read(struct xnvme_cmd_ctx *ctx, uint32_t nsid, uint64_t slba, uint16_t
 /**
  * Submit a NVMe Write command
  *
+ * NOTE: fields which are not arguments here, such as 'ctx->cmd.nvm.fua', are
```
safl, line 37: The note explains context reuse but not that honouring `fua` is backend-dependent. After this change `io_uring` and `libaio` translate it to `RWF_DSYNC`, while `psync`, the POSIX-aio backend, `ramdisk`, FreeBSD, Windows and macOS ignore it silently; passthru backends carry it in the command and are unaffected. A consumer that sets `fua` gets durability on some backends and silence on others, and this header is where that belongs.

lib/xnvme_be_linux_async_libaio.c
```diff
@@ -188,13 +189,24 @@ _linux_libaio_cmd_io(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf_nbytes,
...
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d", ctx->cmd.common.opcode);
 		return -ENOSYS;
 	}
 
+	// Honour the FUA bit of the write; on a raw block device RWF_DSYNC makes
+	// the kernel issue the write with REQ_FUA
+	if (ctx->cmd.nvm.fua && (ctx->cmd.common.opcode == XNVME_SPEC_NVM_OPC_WRITE ||
+				 ctx->cmd.common.opcode == XNVME_SPEC_FS_OPC_WRITE)) {
+		iocb.aio_rw_flags = RWF_DSYNC;
```
safl, line 207: `ctx->cmd.nvm.fua` is read for `XNVME_SPEC_FS_OPC_WRITE` as well, where the caller had no reason to set it. Combined with `xnvme_queue_get_cmd_ctx()` not clearing the context, a stale bit from an earlier command silently turns a file write into `RWF_DSYNC`. The header note documents the hazard, which is honest, and the same argument applies to `xnvme_be_linux_async_liburing.c:344`. It is a performance surprise rather than a correctness one, since `DSYNC` is the stronger guarantee.

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -381,11 +424,22 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
...
 		break;
 
+	case XNVME_SPEC_NVM_OPC_FLUSH:
+	case XNVME_SPEC_FS_OPC_FLUSH:
+		// NOTE: full fsync, matching the psync backend
+		io_uring_prep_fsync(sqe, fd, 0);
+		break;
+
 	default:
 		XNVME_DEBUG("FAILED: unsupported opcode: %d for async", ctx->cmd.common.opcode);
```
safl, line 446: Optional, and pre-existing rather than introduced here. This `default:` returns `-ENOSYS` after `io_uring_get_sqe()`, so the SQE is consumed but never submitted and the next `io_uring_submit()` sends it with stale content. Mentioning it only because the FLUSH path above was deliberately written to avoid exactly this, so the asymmetry stands out.

lib/xnvme_be_linux_async_liburing.c
yonggilsong: Added in the two test commits: 'tests(ioworker)' introduces a verify-flush subcommand (write, FLUSH through the queue, read-compare) and 'tests(cijoe)' runs it across the backend matrix with explicit --async, covering io_uring, libaio and the passthru/sync-mixin paths; async=posix is skipped as it still rejects FLUSH.

lib/xnvme_be_linux_async_liburing.c
yonggilsong: Split out as the first commit of the series, with the user-visible SQPOLL symptom leading the message. The cijoe commit adds 'verify --poll_sq 1 --vec-cnt 4' restricted to async=io_uring as its regression test; the poll options are now declared on the ioworker subs to make that reachable.

include/libxnvme_nvm.h
yonggilsong: Done. The note now states that honouring fua is backend-dependent (passthru backends carry the bit in the command, io_uring/libaio translate it to RWF_DSYNC, the rest ignore it) and that it applies to NVM writes only.

lib/xnvme_be_linux_async_libaio.c
yonggilsong: Restricted to XNVME_SPEC_NVM_OPC_WRITE in both backends. fua is a field of the NVMe write command; the FS opcodes are a separate namespace defining no such field, and the file-API has no way of setting it, so a set bit combined with an FS write could only be stale state from a reused context. If per-write durability for files turns out to be wanted, a file-API flag mapping to RWF_DSYNC would be the way, matching how the kernel defines it for regular files - out of scope here.

lib/xnvme_be_linux_async_liburing.c
yonggilsong: Taken as its own fix commit: the vectored path now dispatches on the opcode before io_uring_get_sqe(), mirroring the single-buffer path, so the default-case rejection no longer leaves an acquired sqe in the ring.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -195,3 +195,36 @@ def test_verify_sync_iovec_direct(cijoe, device, be_opts, cli_args):
...
+
+@xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
+def test_verify_sqpoll_iovec(cijoe, device, be_opts, cli_args):
+    if be_opts["async"] != "io_uring":
+        pytest.skip(reason="poll_sq (SQPOLL) is io_uring-only")
+
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify {cli_args} --poll_sq 1 --vec-cnt 4 --nlb 7"
+    )
+    assert not err
```
karlowich: I think this is unrelated to the flush work.

tests/ioworker.c
```diff
@@ -527,6 +615,28 @@ static struct xnvme_cli_sub g_subs[] = {
 			{XNVME_CLI_OPT_NON_POSA_TITLE, XNVME_CLI_SKIP},
 			{XNVME_CLI_OPT_VEC_CNT, XNVME_CLI_LOPT},
 			{XNVME_CLI_OPT_NLB, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_QDEPTH, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_POLL_SQ, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_POLL_IO, XNVME_CLI_LOPT},
```
karlowich: I think you're mixing features and fixes here.
I think the qdepth fix should be separate and the same goes for the POLL_SQ / POLL_IO additions.
Perhaps these should be added as a separate PR instead, just to keep things focused.
Besides, if I recall correctly, POLL_SQ / POLL_IO requires reloading the kernel NVMe driver which I don't think we support in our CI (yet).

karlowich (changes requested):
The work on flush looks good, but there are some other things bundled with it that I would prefer to have separate.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -195,3 +195,36 @@ def test_verify_sync_iovec_direct(cijoe, device, be_opts, cli_args):
         f"xnvme_tests_ioworker verify-sync {cli_args} --nlb 7 --vec-cnt 4 --direct 1"
     )
     assert not err
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
+def test_verify_flush(cijoe, device, be_opts, cli_args):
+    if be_opts["async"] == "posix":
```
safl: This skips `async=posix` but not `async=kqueue`. The FreeBSD job `verify (freebsd, 14, pcie, test_os)` does fail on this branch, on `test_verify_flush` with `async=kqueue`, and it seems to me that it comes down to the same thing: FreeBSD has no FLUSH implementation either. Both FLUSH opcodes fall into the `default:` in `lib/xnvme_be_freebsd_async.c`, in `_cmd_io` and in `_cmd_iov`, under a commented-out `aio_fsync` TODO, returning the same `-ENOSYS` this PR fixes for io_uring and libaio.

Skipping on the capability rather than naming posix alone would also cover Windows `iocp`/`iocp_th`, which handle `XNVME_SPEC_FS_OPC_FLUSH` but not `XNVME_SPEC_NVM_OPC_FLUSH`.

tests/ioworker.c
```diff
@@ -397,6 +398,72 @@ start_workers(struct iowork *work)
...
+	default:
+		xnvme_cmd_ctx_set_cb(ctx, on_completion, work);
+		xnvme_queue_put_cmd_ctx(work->queue, ctx);
+		return err;
+	}
+
+	while (!state.done) {
+		err = xnvme_queue_poke(work->queue, 0);
+		if (err < 0) {
+			return err;
```
safl, line 472: The submit-failure path just above restores `on_completion` and puts the context back. This return does neither, so it leaks the `xnvme_cmd_ctx` and leaves the queue-wide callback overridden with a `cb_arg` pointing at this frame's `state`.

Nothing dereferences that today: the caller goes to `final()` and `xnvme_queue_term()`, and neither Linux backend's term drains the queue or runs completions, so the stale callback is never invoked. So this is an inconsistent error path rather than a live bug. Worth making symmetric with the path above regardless, since that asymmetry is what would turn it into one if teardown ever started draining.

Related, a few lines up: the `-EBUSY`/`-EAGAIN` retry ignores a negative return from `xnvme_queue_poke()` and can spin indefinitely.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -195,3 +195,36 @@ def test_verify_sync_iovec_direct(cijoe, device, be_opts, cli_args):
...
+def test_verify_flush_iopoll_rejected(cijoe, device, be_opts, cli_args):
+    if be_opts["async"] != "io_uring":
+        pytest.skip(reason="poll_io (IOPOLL) is io_uring-only")
+
+    # IORING_OP_FSYNC has no iopoll-handler; the backend rejects FLUSH on an
+    # IOPOLL queue at submission-time, so the run must fail
+    err, _ = cijoe.run(
+        f"xnvme_tests_ioworker verify-flush {cli_args} --poll_io 1 --direct 1"
+    )
+    assert err
```
safl: Any nonzero exit satisfies this: a verify mismatch, a `--direct` open failure, or the write phase failing would all pass, and none of them show the submission-time IOPOLL rejection the comment describes. The same command without `--poll_io 1`, asserted to succeed, would separate the two. `test_verify_flush` is not that control, since it runs without `--direct` and on `labels=["dev"]` rather than `["bdev"]`.

safl (commented):
Three things worth a look. The FreeBSD job is failing on this branch, and the cause looks to be in the new test rather than in the backend change; the Linux code paths came through green.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Done — the skip now covers the async-interfaces without an NVM FLUSH implementation: posix, kqueue, and iocp/iocp_th.

tests/ioworker.c
yonggilsong: Both fixed: the poke-failure paths now restore the queue-wide callback and put the context back, same as the submit-failure path, and a negative poke return breaks the retry and wait loops.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: The IOPOLL case moves to the follow-up PR along with the poll options; the control run you suggest will be part of it.

tests/ioworker.c
yonggilsong: Split out as announced in the thread — this PR is down to FLUSH/FUA and its tests.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Moved to the follow-up PR together with the fix it exercises.

lib/xnvme_be_linux_async_liburing.c
```diff
@@ -381,6 +431,12 @@ xnvme_be_linux_liburing_cmd_iov(struct xnvme_cmd_ctx *ctx, struct iovec *dvec, s
 		io_uring_prep_readv(sqe, fd, dvec, dvec_cnt, ctx->cmd.nvm.slba << ssw);
 		break;
 
+	case XNVME_SPEC_NVM_OPC_FLUSH:
+	case XNVME_SPEC_FS_OPC_FLUSH:
+		// NOTE: full fsync, matching the psync backend
+		io_uring_prep_fsync(sqe, fd, 0);
+		break;
```
safl, line 443: `sqe->flags` is assigned before this switch, and `io_uring_prep_fsync()` goes through `io_uring_prep_rw()`. On liburing versions where that helper clears `flags`, this drops `IOSQE_FIXED_FILE`, and since `fd` is 0 under `poll_sq` the vectored FLUSH then fsyncs real file descriptor 0 rather than the device.

Worth qualifying: I checked six liburing headers here and four do not clear `flags` while two do, so whether it bites depends on the version built against. It is the same ordering hazard you diagnosed and fixed for the single-buffer path a couple of hunks up, with the NOTE explaining it. The readv/writev instances are pre-existing and deferred to the follow-up, which is fine, but this one is new. Setting `sqe->flags` after `io_uring_prep_fsync()` here would match what `cmd_io` already does.

tests/ioworker.c
```diff
@@ -397,6 +398,78 @@ start_workers(struct iowork *work)
...
+		goto retry;
+
+	default:
+		goto failed;
+	}
+
+	while (!state.done) {
+		err = xnvme_queue_poke(work->queue, 0);
+		if (err < 0) {
+			goto failed;
```
safl: The `failed:` label is right for the two paths that have not submitted yet, but reaching it from here is not: the command was submitted and `state.done` is still 0, so `xnvme_queue_put_cmd_ctx()` puts a context back on the free list while it is still the io_uring `user_data` for an in-flight command, and leaves `outstanding` non-zero going into `final()`.

Returning `err` directly from this loop is enough. Nothing invokes the completion afterwards, since queue teardown does not drain.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -195,3 +195,15 @@ def test_verify_sync_iovec_direct(cijoe, device, be_opts, cli_args):
...
     )
     assert not err
+
+
+@xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
+def test_verify_flush(cijoe, device, be_opts, cli_args):
+    # Async-interfaces without an implementation of NVM FLUSH: 'posix' and
+    # FreeBSD 'kqueue' reject it with -ENOSYS, Windows 'iocp'/'iocp_th' handle
+    # only the FS FLUSH opcode
+    if be_opts["async"] in ["posix", "kqueue", "iocp", "iocp_th"]:
```
safl: Two small things in this list. `ioring` belongs in it: `lib/xnvme_be_windows_async_ioring.c` handles only READ and WRITE and returns `-ENOSYS` from its `default:`, and `xnvme_be_combinations.py` declares a `be=ioring, async=ioring` bdev combination. It will not turn a job red today, since the Windows CI job selects ramdisk only, but the gap is there and the comment claims to enumerate them.

In the other direction, `posix` matches nothing: no combination declares `async: ["posix"]`, only `mem: ["posix"]`.

tests/meson.build
```diff
@@ -39,6 +39,7 @@ tests = {
   ],
   'ioworker.c': [
     ['verify', ['verify', '1GB']],
+    ['verify_flush', ['verify-flush', '1GB']],
```
safl, line 42: This is the only test that runs on every PR, and it gives no coverage of the change: the `1GB` URI resolves to the nil backend, whose `cmd_io` ignores `ctx->cmd.common.opcode` entirely and completes everything, so this would pass unchanged with the whole FLUSH dispatch reverted.

Adding a second entry with `--async emu` fixes that, and I checked it runs and walks a real opcode dispatch.

lib/xnvme_be_linux_async_libaio.c
```diff
@@ -11,6 +11,7 @@
 #ifdef XNVME_BE_LINUX_LIBAIO_ENABLED
 #include <errno.h>
 #include <libaio.h>
+#include <linux/fs.h>
```
safl, line 14: This include is not needed and it is noisy: the file already defines `_GNU_SOURCE` and includes `libxnvme.h`, which reaches glibc's `bits/uio-ext.h` and defines the whole `RWF_*` set. Compiling this file with `-Wall -Wextra` gives eight `redefined` warnings (`RWF_HIPRI`, `RWF_DSYNC`, `RWF_SYNC`, `RWF_NOWAIT`, `RWF_APPEND`, `RWF_NOAPPEND`, `RWF_ATOMIC`, `RWF_DONTCACHE`), and dropping the include compiles clean. Same in the liburing file, where `liburing/io_uring.h` pulls `linux/fs.h` anyway.

If musl is the worry, an `#ifndef RWF_DSYNC` fallback would be lighter than the kernel header.

tests/ioworker.c
```diff
@@ -531,6 +625,22 @@ static struct xnvme_cli_sub g_subs[] = {
...
+			{XNVME_CLI_OPT_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_URI, XNVME_CLI_POSA},
+
+			{XNVME_CLI_OPT_NON_POSA_TITLE, XNVME_CLI_SKIP},
+			{XNVME_CLI_OPT_VEC_CNT, XNVME_CLI_LOPT},
+			{XNVME_CLI_OPT_NLB, XNVME_CLI_LOPT},
+
+			XNVME_CLI_ASYNC_OPTS,
+		},
+	},
```
safl, line 659: Two coverage gaps worth naming somewhere, here or in the description.

Nothing sets `ctx->cmd.nvm.fua`, so the four new `RWF_DSYNC` assignments have no test at all. A `verify` variant that sets it on the write phase would at least catch a submission-time rejection on every backend.

And `iowork_flush()` submits with `xnvme_cmd_pass()`, which always lands in `cmd_io`, so the new FLUSH branches in `cmd_iov` are never reached; `--vec-cnt 4` changes only the data phase. That means the vectored FLUSH path is unexercised even by the manual run reported in the description.

safl (commented):
Thanks for splitting the fixes out; that reads much better. The head has moved
to four commits since the description was written, and the description still
says two and still describes the sqe->flags move that is no longer here, so it
is worth refreshing.

The kqueue skip is fixed. Two things below are correctness rather than polish,
and the IOPOLL rejection lost its test rather than gaining a tighter one.

I built the branch but this machine has neither liburing nor libaio, so those
two files were not compiled by my build and I have no NVMe hardware;
everything below is from reading, except where I say otherwise.

---

lib/xnvme_be_linux_async_libaio.c
safl: Correcting the justification here. The eight warnings I quoted are not reproducible in a real build: both definition sites are system headers, and gcc and clang suppress macro-redefinition warnings sited in system headers unless `-Wsystem-headers` is given. I got them from a synthetic compile, not from building xNVMe, and I should have said so or checked.

The substance is unchanged: the include is redundant, since `libxnvme.h` reaches `bits/uio-ext.h` which defines `RWF_DSYNC`, the only symbol these files take from `linux/fs.h`, and liburing's own header includes it anyway. So dropping it is still worth doing, just as tidying rather than to silence anything.

tests/meson.build
safl: Correcting the framing here; the technical point stands but two claims around it were wrong.

"The only test that runs on every PR" is wrong twice over. Nothing in CI runs `meson test` at all, so this entry runs on zero PRs. And your cijoe `test_verify_flush` does run on every PR, through `make verify-ramdisk`, whose selection matches the ramdisk_emu and ramdisk_thrpool combinations, and the ramdisk sync backend has a real FLUSH case. So the change is better covered than I implied.

I also said this would pass unchanged with the feature reverted. On my machine it does not pass at all: `verify-flush 1GB` fails deterministically, as does `verify 1GB` on unmodified `main`, which looks like a pre-existing nil-backend problem that CI never sees precisely because meson tests are not run there.

What remains is the original point: the `1GB` URI resolves to the nil backend, whose `cmd_io` ignores the opcode entirely, so this particular entry proves nothing about FLUSH. An `--async emu` entry does walk a real dispatch and passes.

karlowich (dismissed):
This looks good to me.
(That is, I don't have anything besides Simon's comments)

lib/xnvme_be_linux_async_liburing.c
yonggilsong: Fixed — the FLUSH branch in cmd_iov now re-assigns sqe->flags after io_uring_prep_fsync(), matching cmd_io. The readv/writev instances stay deferred to the follow-up as you note.

tests/ioworker.c
yonggilsong: Fixed — the post-submission poke-failure now returns directly; only the not-yet-submitted paths restore the callback and return the context.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Fixed — ioring added, posix dropped.

tests/meson.build
yonggilsong: Added two '--async emu' entries, plain and vectored.

lib/xnvme_be_linux_async_libaio.c
yonggilsong: Dropped from both files; compiles clean without it.

tests/ioworker.c
yonggilsong: Both covered now rather than named: verify-flush sets the FUA bit on its writes, and its FLUSH goes through xnvme_cmd_pass_iov() when --vec-cnt is given, so the cmd_iov branches are exercised. The description carries the remaining known gap (vectored SQPOLL data-I/O, deferred with its fix).

safl (commented):
Withdrawn, please skip this one; it was wrong on both counts and I do not want
it costing you time.

It claimed the `sqe->flags` zeroing in `io_uring_prep_rw()` is not
version-dependent. It is. Of the six liburing headers here only two clear
`sqe->flags`; the other four set `opcode`, `fd`, `off`, `addr` and `len` and
nothing else. The earlier finding on this PR had it right and I generalised
from a single header.

It also re-raised the readv/writev instances as though they were open, when
they were called out as pre-existing, deferred to the follow-up, and agreed as
deferred. Nothing for you to do there.

The one small thing that survives is in the description rather than the code,
and it is in the comment below this one.

safl (commented):
Correcting my previous comment. Two things in it were wrong.

"It is not version-dependent" is wrong, and it contradicts a count made
earlier on this PR that was right. I have six liburing headers here and only
two of them clear `sqe->flags` in `io_uring_prep_rw()`; the other four set
`opcode`, `fd`, `off`, `addr` and `len` and nothing else. I generalised from
the first header I opened. So whether the hazard bites does depend on the
version built against, exactly as stated before.

The second is that I re-raised the readv/writev instances as though they were
open. They are not. They were called out as pre-existing, deferred to the
follow-up, and you confirmed the deferral. That deferral is fine and I should
not have reopened it.

What does still want doing is smaller and is in the description rather than
the code. It says "Two commits" where the head has four, and it still says
that the first commit "moves the `sqe->flags` assignment in the vectored
io_uring path to after the prep-helpers". The branch does not do that. The
assignment before the switch is still there, and only the FSYNC case
re-assigns after its prep-helper. So the description credits the PR with the
fix that was deferred, which is the one place a reader would go looking to
find out whether it is in.

Reviewed with a coding agent. The header count above was rerun rather than
recalled; the rest is from reading the branch.

safl (dismissed):
LGTM, as mentioned, we address additional things as follow-up after #755

safl (approved):
Great!

naddinadja (approved):
LGTM :)

PR #748, opened 2026-08-25
draft: NVMe over Fabrics backend for xNVMe

jswaro:
I'm opening this pull request as a way for early visibility into the change.


lib/nvmf/xnvme_be_nvmf_async.c
```diff
@@ -0,0 +1,139 @@
...
+ * thus allocated and de-allocated by:
+ *
+ * The XNVME_BE_NVMF specific context is a NVMF qpair and it is carried inside:
+ *
+ * xnvme_queue->be_rsvd
+ */
+int
+xnvme_be_nvmf_queue_init(struct xnvme_queue *q, int XNVME_UNUSED(opts))
+{
+	struct xnvme_be_nvmf_queue *queue = (struct xnvme_be_nvmf_queue *)q;
```
karlowich: Looks like you forgot to create and connect the qpair here.

lib/nvmf/xnvme_be_nvmf_admin.c
```diff
@@ -0,0 +1,143 @@
...
+	req = xnvme_be_nvmf_req_alloc(qpair->req_pool, false, (void *)ctx);
+	if (!req) {
+		XNVME_DEBUG("FAILED: xnvme_be_nvmf_req_alloc()");
+		return -ENOSPC;
+	}
+
+	/* Set the command identifier (CID) to the request's CID */
+	ctx->cmd.common.cid = req->cid;
+
+	pthread_mutex_lock(&state->lock);
```
karlowich: Why do you need to lock here? Most xNVMe functions aren't considered thread-safe. We put this responsibility on the caller.

lib/nvmf/xnvme_be_nvmf_dev.c
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Maximum number of attempts to probe for a device matching the provided URI.
+ *
+ * This is used to retry probing for a device in case of transient errors or
+ * delays in the NVMe-oF subsystem.
+ */
+#define XNVME_BE_NVMF_MAX_PROBE_ATTEMPTS 3
+
+#define FOR_EACH_NVMF_TRANSPORT(transport)                                          \
```
karlowich: Would it simplify things if the transport was provided by the user instead.
I see no problem in having multiple backends "nvmf-tcp", "nvmf-rdma" etc.
They can share most of the code and have separate `xnvme_be_dev` if necessary.
The different version would be defined in `xnvme_be_nvmf.c`.
For reference, we do something similar for the different `upcie` versions.
See `xnvme_be_upcie.c`.

lib/nvmf/xnvme_be_nvmf.h
```diff
@@ -0,0 +1,65 @@
...
+#define XNVME_BE_NVMF_ADMIN_QUEUE_ID    0
+#define XNVME_BE_NVMF_SYNC_QUEUE_ID     1
+#define XNVME_BE_NVMF_IO_QUEUE_ID_START XNVME_BE_NVMF_SYNC_QUEUE_ID
+
+#define NVME_CMD_CAPSULE_SIZE sizeof(struct xnvme_spec_cmd_common)
+#define NVME_CPL_CAPSULE_SIZE sizeof(struct xnvme_spec_cpl)
+
+struct xnvme_be_nvmf_queue {
+	struct xnvme_queue_base base;
+	uint64_t completions_pending;
```
karlowich: You can use `xnvme_queue_base.outstanding` instead.

lib/nvmf/xnvme_be_nvmf_qpair.h
```diff
@@ -0,0 +1,159 @@
...
+int
+xnvme_be_nvmf_qpair_create(struct xnvme_be_nvmf_ctrlr *ctrlr, struct xnvme_be_nvmf_qpair_attr *attr,
+			   struct xnvme_be_nvmf_qpair **qpair);
+
+int
+xnvme_be_nvmf_connect_qpair(struct xnvme_be_nvmf_qpair *qpair);
+int
+xnvme_be_nvmf_disconnect_qpair(struct xnvme_be_nvmf_qpair *qpair);
+int
+xnvme_be_nvmf_destroy_qpair(struct xnvme_be_nvmf_qpair *qpair);
```
karlowich: Is there any benefit to separating create/destroy and connect/disconnect?
From my brief look, I think they could easily be merged.
Would that simplify the flow a little bit?

lib/nvmf/rdma/xnvme_be_nvmf_rdma_qpair.c
```diff
@@ -0,0 +1,572 @@
...
+static struct xnvme_be_nvmf_qpair_ops g_xnvme_be_nvmf_rdma_qpair_ops = {
+	.connect = _connect_rdma_qpair_sync,
+	.disconnect = _disconnect_rdma_qpair_sync,
+	.destroy = _rdma_destroy,
+	.reg_mr = _rdma_register_memory,
+	.dereg_mr = _rdma_deregister_memory,
+
+	.cmd_io = _rdma_cmd_io,
+	.cmd_iov = _rdma_cmd_iov,
+	.send_capsule = _rdma_send_capsule,
```
karlowich: When you have `send_capsule()`, why do you need `cmd_io()`/`cmd_iov()`?
I think the async/sync interface could just translate to `send_capsule()` like your admin interface

karlowich (commented):
Great progress on this!
I have left a few comments inline. I have only skimmed the code, so there might be obvious answers to my questions.

lib/nvmf/xnvme_be_nvmf_async.c
karlowich: After reading more, I see that the I/O interface ends in `ENOSYS` right now, so my comment is probably irrelevant.

lib/nvmf/xnvme_be_nvmf_admin.c
karlowich: In general, I see you have many locks and atomic operations in the code.
I don't think that is necessary at this stage. Of course they don't hurt, but you might be able to simplify things by assuming single threaded execution for now.

PR #755, opened 2026-08-26
tools: add qublk, a ublk server backed by xNVMe

yonggilsong:
This moves qublk — the ublk server backed by xNVMe — into the tree as `tools/qublk`, integrated with the xNVMe build system and the `xnvme_cli` framework, per the xnvmeperf pattern. It creates `/dev/ublkbN` and serves the kernel's READ/WRITE/FLUSH (and FUA writes) through the xNVMe async API, so any xNVMe backend can be exposed as a regular block device.

Eight commits, restructured per review:

1. **feat(tools/qublk): add ublk server backed by xnvme** — the server itself: control-path, per-queue I/O threads on io_uring cmd, FLUSH/FUA handling, defensive validation of I/O sizes.
2. **fix(tools/qublk): widen qdepth and nqueues to 32 bits** — the fields matched neither the UAPI nor the CLI width, so out-of-range values were silently narrowed into in-range ones; renamed to the xNVMe convention while at it.
3. **fix(tools/qublk): io-path and teardown robustness** — SQE-exhaustion no longer spins the dispatch-retry, the remote-flush wait is bounded on stop, teardown issues STOP_DEV before joining the I/O threads (matching ubdsrv) and the drain dispatches rather than discards, `_GNU_SOURCE` guarded.
4. **build(tools/qublk): skip the build on kernel headers without ioctl-encoded ublk commands** — probes for `UBLK_U_CMD_GET_FEATURES`; fixes the Debian bookworm jobs (kernel 6.1 headers).
5. **feat(cli): add dev-id and max-io-bytes options** — two new shared CLI options (LOPT), with range validation; queue count reuses the existing `--nqueues`.
6. **refactor(tools/qublk): use the xnvme_cli helpers** — replaces the standalone getopt CLI with `xnvme_cli`; validation now happens at full width before any narrowing.
7. **docs(tools/qublk): document the qublk tool** — usage page under `docs/tools/`.
8. **test(tools/qublk): add cijoe functional and filesystem tests** — block-level (dd, multi-queue, max-io-bytes) and XFS-level (format/mount/remount, file ops, persistence across a qublk restart) cases; the daemon runs with a fixed `--dev-id` and the session refuses a pre-existing node; skips where ublk or `mkfs.xfs` is unavailable.

**Dependency note**: the filesystem-level cases (commit 6) need NVMe FLUSH in the Linux async backends, i.e. #747 — mkfs on a served namespace makes the kernel issue FLUSH, which io_uring/libaio reject with -ENOSYS without it. With #747 merged first (or its branch merged in locally), the whole suite is green; without it, the fs cases fail on the Linux async backends. The block-level cases and the tool itself do not depend on it.

**CI coverage**: the trixie-based verify jobs do run the qublk cases (their guests have `ublk_drv`); the filesystem cases currently skip there for lack of `mkfs.xfs` in the guest image — adding `xfsprogs` to the provisioning would light them up, which I can do as a follow-up. The bookworm jobs skip the tool at configure-time per commit 4.

## Verification

On Ubuntu 24.04, kernel 6.8.12, as root against a Samsung SSD 970 EVO 1TB and the ramdisk backend:

- `meson test`: 41 pass (the two new CLI option cases included), failures identical to `main` on the same machine.
- cijoe run of `test_qublk.py` + `test_qublk_fs.py` on a build including #747: 88 passed, 47 skipped, 0 failed — including mkfs/mount/remount and the restart-persistence case.
- FUA path checked via ftrace `block_rq_issue`: with `fua=1` both io_uring and libaio issue the write followed by a flush to the device; without the fix only the bare write went down.
- `--dev-id` and `--max-io-bytes` out-of-range values are rejected at startup.

### DIFF
--- a/cijoe/tests/tools/qublk_session.py
+++ b/cijoe/tests/tools/qublk_session.py
@@ -1,0 +1,130 @@
+"""
+qublk session-helper
+====================
+
+qublk is a blocking daemon; it serves a ublk block-device until signalled. Exercising it
+therefore cannot be done by running a command and inspecting its status, the way the
+one-shot command-line tools are tested. Instead, each test-case runs a single shell
+session which launches qublk in the background, waits for the block-device to appear,
+runs a payload against it, and tears it down with SIGINT.
+
+* qublk_session(cijoe, uri, be, payload, ...): run 'payload' against a served device
+
+* qublk_teardown(cijoe, mountpoint): remove leftovers from a failed test-case
+
+Three things the session-script must get right, each of which has bitten before:
+
+* The lines are joined by newline, not by ';'. The background launch already ends in
+  '&', and a ';' following it is a shell syntax error.
+
+* The qublk log goes to a mktemp path. A fixed path in /tmp breaks the next run as soon
+  as one is left behind by another user.
+
+* The payload runs in a subshell with 'set -e', so the first failing line decides the
+  status. Without it a multi-line payload would report only its last line, and e.g. a
+  failed mkfs followed by a successful mount would pass. The subshell keeps 'set -e'
+  away from the teardown, which must run regardless.
+"""
+
+import pytest
+
+UBLK_NODE = "/dev/ublkb0"
+UBLK_CONTROL = "/dev/ublk-control"
+
+DEFAULT_TIMEOUT_TICKS = 50  # 0.2s each, so 10s for the block-device to appear
+
+
+def require_ublk(cijoe):
+    """
+    Skip unless ublk can actually be used
+
+    qublk needs root and the 'ublk_drv' module. Neither is available in the unprivileged
+    containers running the ramdisk test-suite, and those select on the 'ramdisk' keyword,
+    which matches the ramdisk-backend parametrization of these very test-cases. Without
+    this guard they would be collected there and fail for reasons having nothing to do
+    with qublk.
+    """
+
+    err, _ = cijoe.run("test $(id -u) -eq 0")
+    if err:
+        pytest.skip("qublk requires root")
+
+    cijoe.run("modprobe ublk_drv 2>/dev/null || true")
+
+    err, _ = cijoe.run(f"test -c {UBLK_CONTROL}")
+    if err:
+        pytest.skip(f"qublk requires the ublk_drv module ({UBLK_CONTROL} is absent)")
+
+
+def require_mkfs_xfs(cijoe):
+    """Skip unless mkfs.xfs is available; the fs-level cases format with XFS"""
+
+    err, _ = cijoe.run("command -v mkfs.xfs")
+    if err:
+        pytest.skip("the filesystem-level cases require mkfs.xfs (xfsprogs)")
+
+
+def qublk_script(uri, be, args, payload, node=UBLK_NODE, mountpoint=None):
+    """
+    Produce the shell session serving 'node' from the given device
+
+    'payload' is a list of shell lines run while the block-device is served; its exit status
+    becomes the exit status of the session. When 'mountpoint' is given it is unmounted during
+    teardown regardless of what the payload did.
+    """
+
+    umount = [
+        f"umount {mountpoint} 2>/dev/null || umount -l {mountpoint} 2>/dev/null || true"
+    ]
+
+    return "\n".join(
+        [
+            "set -u",
+            "id -un",
+            "modprobe ublk_drv || echo MODPROBE-FAILED",
+            # A leftover device from a crashed run would satisfy the readiness
+            # wait while the qublk under test got another id; refuse to start
+            f"if [ -b {node} ]; then echo PREEXISTING-DEVICE; exit 1; fi",
+            "log=$(mktemp)",
+            f"qublk run {uri} --be {be} --dev-id 0 {args} > $log 2>&1 &".replace(
+                "  >", " >"
+            ),
+            "pid=$!",
+            f"for i in $(seq 1 {DEFAULT_TIMEOUT_TICKS}); "
+            f"do [ -b {node} ] && break; sleep 0.2; done",
+            f"if [ ! -b {node} ]; then echo MISSING-DEVICE; cat $log; "
+            "kill -INT $pid 2>/dev/null; exit 1; fi",
+            "(",
+            "set -e",
+            *payload,
+            ")",
+            "rc=$?",
+            *(umount if mountpoint else []),
+            "kill -INT $pid 2>/dev/null",
+            "wait $pid",
+            f"if [ -b {node} ]; then echo LEFTOVER-DEVICE; rc=1; fi",
+            "cat $log",
+            "rm -f $log",
+            "exit $rc",
+        ]
+    )
+
+
+def qublk_session(cijoe, uri, be, payload, args="", node=UBLK_NODE, mountpoint=None):
+    """Run 'payload' against the ublk block-device served by qublk"""
+
+    script = qublk_script(uri, be, args, payload, node=node, mountpoint=mountpoint)
+
+    return cijoe.run(f"bash -c '{script}'")
+
+
+def qublk_teardown(cijoe, mountpoint=None, node=UBLK_NODE):
+    """Ensure a failed test-case does not leave a mount or a ublk device behind"""
+
+    if mountpoint:
+        cijoe.run(
+            f"umount {mountpoint} 2>/dev/null || umount -l {mountpoint} 2>/dev/null || true"
+        )
+
+    cijoe.run("pkill -INT qublk || true")
+    cijoe.run(f"for i in $(seq 1 25); do [ -b {node} ] || break; sleep 0.2; done")
--- a/docs/tools/qublk/index.rst
+++ b/docs/tools/qublk/index.rst
@@ -1,0 +1,54 @@
+.. _sec-tools-qublk:
+
+qublk
+#####
+
+**qublk** is a ublk server backed by **xNVMe**. It creates a ``/dev/ublkbN``
+block device, receives ``READ`` / ``WRITE`` / ``FLUSH`` requests from the Linux
+``ublk_drv`` over ``io_uring``, and services them through the **xNVMe**
+asynchronous command interface. Thus, any **xNVMe** backend, such as
+:ref:`sec-backends-upcie` or **io_uring**, becomes usable as an ordinary block
+device.
+
+``REQ_FUA`` and ``REQ_PREFLUSH`` are supported, as are multiple ublk hardware
+queues via ``--nqueues``.
+
+.. literalinclude:: qublk_usage.out
+   :language: bash
+
+Requirements
+============
+
+**qublk** is Linux-only and is built only when ``liburing`` and
+``<linux/ublk_cmd.h>`` are available. At runtime it requires:
+
+* ``root`` privileges
+
+* the ublk driver, loaded with ``modprobe ublk_drv``
+
+* for user space backends such as **uPCIe**, a device bound to
+  ``uio_pci_generic`` and hugepages configured; see :ref:`sec-backends-upcie`
+
+``run`` — Serve a block-device
+==============================
+
+Opens the given device URI, adds a ublk device, and serves it until
+``SIGINT`` / ``SIGTERM``, upon which it performs a clean teardown
+(``STOP_DEV`` followed by ``DEL_DEV``).
+
+When ``--dev-id`` is not given, the kernel assigns the device identifier.
+When ``--max-io-bytes`` is not given, the per-IO buffer size defaults to the
+smaller of 1MiB and the controller ``MDTS``.
+
+.. literalinclude:: qublk_run_usage.out
+   :language: bash
+
+Example — NVMe block device via io_uring::
+
+   qublk run /dev/nvme0n1 --be io_uring --qdepth 64
+
+Example — user space NVMe via uPCIe, with multiple hardware queues::
+
+   qublk run 0000:01:00.0 --be upcie --qdepth 64 --nqueues 4
+
+While **qublk** is running, ``/dev/ublkb0`` is the resulting block device.
--- a/tools/qublk/ctrl.c
+++ b/tools/qublk/ctrl.c
@@ -1,0 +1,198 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#define _GNU_SOURCE
+#include "ctrl.h"
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <liburing.h>
+
+#define CTRL_DEV "/dev/ublk-control"
+
+static int
+ctrl_uring_cmd(int ctrl_fd, uint32_t cmd_op, const struct ublksrv_ctrl_cmd *cmd)
+{
+	struct io_uring ring;
+	struct io_uring_params p = {
+		.flags = IORING_SETUP_SQE128 | IORING_SETUP_CQE32,
+	};
+	struct io_uring_sqe *sqe;
+	struct io_uring_cqe *cqe;
+	int rc, res;
+
+	rc = io_uring_queue_init_params(2, &ring, &p);
+	if (rc < 0) {
+		fprintf(stderr, "io_uring_queue_init_params(ctrl): %s\n", strerror(-rc));
+		return rc;
+	}
+
+	sqe = io_uring_get_sqe(&ring);
+	if (!sqe) {
+		io_uring_queue_exit(&ring);
+		return -ENOMEM;
+	}
+
+	io_uring_prep_rw(IORING_OP_URING_CMD, sqe, ctrl_fd, NULL, 0, 0);
+	sqe->cmd_op = cmd_op;
+	memset(sqe->cmd, 0, 80);
+	memcpy(sqe->cmd, cmd, sizeof(*cmd));
+
+	rc = io_uring_submit_and_wait(&ring, 1);
+	if (rc < 0) {
+		fprintf(stderr, "io_uring_submit_and_wait(ctrl 0x%x): %s\n", cmd_op,
+			strerror(-rc));
+		io_uring_queue_exit(&ring);
+		return rc;
+	}
+
+	rc = io_uring_peek_cqe(&ring, &cqe);
+	if (rc < 0) {
+		fprintf(stderr, "io_uring_peek_cqe(ctrl 0x%x): %s\n", cmd_op, strerror(-rc));
+		io_uring_queue_exit(&ring);
+		return rc;
+	}
+
+	res = cqe->res;
+	io_uring_cqe_seen(&ring, cqe);
+	io_uring_queue_exit(&ring);
+
+	if (res < 0) {
+		fprintf(stderr, "ublk ctrl cmd 0x%x failed: %s\n", cmd_op, strerror(-res));
+		return res;
+	}
+	return 0;
+}
+
+int
+qublk_ctrl_open(struct qublk_dev *dev)
+{
+	dev->ctrl_fd = open(CTRL_DEV, O_RDWR | O_CLOEXEC);
+	if (dev->ctrl_fd < 0) {
+		fprintf(stderr, "open(%s): %s\n", CTRL_DEV, strerror(errno));
+		return -errno;
+	}
+	return 0;
+}
+
+void
+qublk_ctrl_close(struct qublk_dev *dev)
+{
+	if (dev->ctrl_fd >= 0) {
+		close(dev->ctrl_fd);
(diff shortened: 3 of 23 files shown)


safl (commented):
Read through this one carefully; it is a lot of new code and mostly reads
well. Three things below are blocking in my view, and the first is not about
the code at all.

I built it twice, once with no liburing where the meson warning skips the tool
and the tree still builds, and once against a locally built liburing 2.8 where
qublk compiles and links. Ran `qublk --help`, `qublk run --help` and the
argument rejections. There is no ublk driver and no NVMe device here, so
nothing about the served block device, the FLUSH and FUA paths, teardown or
the cijoe cases was exercised; all of that is from reading.

---

tools/qublk/meson.build
```diff
@@ -0,0 +1,28 @@
+# SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
```
safl, line 1: Licensing is the one I would settle before the code. I diffed the import against `karlowich/qublk` and it is faithful: the sources differ only by the SPDX header, formatting, and one functional hunk. But that repository carries no LICENSE file and no license headers on any source file, and this adds `SPDX-License-Identifier: BSD-3-Clause` with `SPDX-FileCopyrightText: Samsung Electronics Co., Ltd` to code written by someone else.

Nothing in the PR records who granted that licence, or that the copyright is Samsung's rather than the upstream author's. Could you get karlowich to confirm in writing before this lands? I cannot resolve it from the artefacts.

tools/qublk/ctrl.c
```diff
@@ -0,0 +1,198 @@
...
+{
+	const struct xnvme_geo *geo = dev->geo;
+	uint8_t lba_shift = dev->lba_shift;
+	uint64_t dev_sectors_512 = geo->tbytes >> 9;
+	struct ublk_params params = {
+		.len = sizeof(params),
+		.types = UBLK_PARAM_TYPE_BASIC,
+		.basic =
+			{
+				.attrs = dev->has_vwc ? (UBLK_ATTR_VOLATILE_CACHE | UBLK_ATTR_FUA)
```
safl, line 151: This advertises `UBLK_ATTR_FUA` whenever the controller reports a volatile write cache, but on `main` today neither `xnvme_be_linux_async_liburing.c` nor `xnvme_be_linux_async_libaio.c` has any `fua` or `RWF_DSYNC` handling, and no FLUSH case at all; that is what #747 adds and it is still open.

So merged in this order, qublk tells the kernel the device honours FUA while a FUA write has its bit dropped with no error. The description mentions the dependency but frames it as the fs cases failing, and the silent durability loss is the part a reviewer needs to see. Either land #747 first, or set `UBLK_ATTR_FUA` only for backends that honour it.

cijoe/tests/tools/qublk_session.py
```diff
@@ -0,0 +1,117 @@
...
+  as one is left behind by another user.
+
+* The payload runs in a subshell with 'set -e', so the first failing line decides the
+  status. Without it a multi-line payload would report only its last line, and e.g. a
+  failed mkfs followed by a successful mount would pass. The subshell keeps 'set -e'
+  away from the teardown, which must run regardless.
+"""
+
+import pytest
+
```
safl, line 30: This hardcodes `/dev/ublkb0` while the tool is launched without `--dev-id`, so the kernel picks the id. The readiness wait is `[ -b /dev/ublkb0 ]`, which succeeds immediately if a `ublkb0` from a crashed run or another consumer already exists, while the qublk under test got id 1. The payload then runs against someone else's device, and in `test_qublk_fs.py` that payload is `mkfs.xfs -f`.

The tool already has what is needed: pass `--dev-id 0` and fail the session up front if the node exists, or parse the id out of qublk's own `/dev/ublkb%d ready` line.

tools/qublk/qublk.h
```diff
@@ -0,0 +1,77 @@
...
+struct qublk_dev;
+struct qublk_queue;
+
+struct qublk_io {
+	uint16_t tag;
+	void *buf;
+	const struct ublksrv_io_desc *iod;
+	struct qublk_queue *q;
+	/*
+	 * NVMe orders FLUSH/FUA only within a single submission queue, so
```
safl, line 28: This comment is the premise the whole cross-queue machinery rests on, and I do not think it matches the spec. Flush is scoped to the namespace and to commands the controller has completed prior to its submission, not to the submission queue it was issued on.

If that reading is right, one FLUSH on the local queue suffices, and the remote-flush pool, the MSG_RING encoding, `handle_msg_do_flush`, `handle_msg_flush_done`, `msg_post_flush_done` and the barrier counters are all unnecessary, roughly 200 of io.c's 880 lines plus a FLUSH per queue and two ring messages per queue on every flush. The FUA case is harder still: flushing other queues has no bearing on the durability of this write. Could you cite the device or spec text that requires the fan-out, or drop it?

tools/qublk/io.c
```diff
@@ -0,0 +1,880 @@
...
+
+	if (tag >= q->depth) {
+		fprintf(stderr, "qublk: bogus tag %u in cqe\n", tag);
+		return -EINVAL;
+	}
+	io = &q->ios[tag];
+
+	if (cqe->res == UBLK_IO_RES_OK) {
+		rc = dispatch(q, io);
+		while (rc == -EBUSY || rc == -EAGAIN) {
```
safl, line 507: This retry conflates two different `-EAGAIN`s. `dispatch` returns whatever `submit_commit_and_fetch` returned, and that returns `-EAGAIN` when `io_uring_get_sqe` is exhausted. Poking the xnvme queue does not free an io_uring SQE, so the loop re-runs the same rejection forever. Give SQE exhaustion its own return code, or submit the ring inside the retry.

tools/qublk/io.c
```diff
@@ -0,0 +1,880 @@
...
+	rf = rflush_alloc(q);
+	if (!rf) {
+		fprintf(stderr, "qublk: q%d remote-flush pool exhausted\n", q->q_id);
+		msg_post_flush_done(q, orig_qid, orig_tag, -ENOMEM);
+		return;
+	}
+	rf->orig_q_id = orig_qid;
+	rf->orig_tag = orig_tag;
+
+	for (;;) {
```
safl, line 312: This spins waiting for a command context with no `dev->stop` check, unlike every other wait loop in the file. A backend that stops completing hangs this thread through teardown, and `qublk_io_thread_join()` then hangs the process with the ublk device still added.

tools/qublk/main.c
```diff
@@ -0,0 +1,217 @@
...
+	const char *uri = cli->args.uri;
+	const char *be = cli->args.be;
+	sigset_t blk;
+	uint64_t feat = 0;
+	uint32_t want_max_io = 0, cap_max;
+	int sig;
+
+	// Options are optional; only override the defaults for the ones actually given
+	if (cli->given[XNVME_CLI_OPT_QDEPTH]) {
+		dev.depth = (uint16_t)cli->args.qdepth;
```
safl: These narrow to `uint16_t` before validating, so out-of-range values are accepted as different values. I ran it: `--qdepth 65600` becomes 64 and sails past the power-of-two check into `/dev/ublk-control`, and `--nqueues 65537` silently becomes 1. Validate the `uint32_t` before the cast.

tools/qublk/main.c
```diff
@@ -0,0 +1,217 @@
...
+		dev.stop = 1;
+		qublk_io_thread_join(&dev);
+		goto err_io;
+	}
+
+	fprintf(stderr, "qublk: /dev/ublkb%d ready (Ctrl-C to stop)\n", dev.dev_id);
+
+	sigwait(&blk, &sig);
+	fprintf(stderr, "qublk: stopping (signal %d)\n", sig);
+	dev.stop = 1;
```
safl, line 174: This sets `dev.stop`, joins the I/O threads, then issues `STOP_DEV`. ubdsrv issues `STOP_DEV` while the queue threads are still servicing, because `del_gendisk()` waits on in-flight requests. Compounding it, the drain loop below counts CQEs and advances the ring but discards them rather than dispatching, so any ublk request delivered after `dev->stop` is never committed. I could not test what the kernel does with those, which is why I would rather the order matched ubdsrv's.

docs/tools/qublk/index.rst
```diff
@@ -0,0 +1,54 @@
...
+#####
+
+**qublk** is a ublk server backed by **xNVMe**. It creates a ``/dev/ublkbN``
+block device, receives ``READ`` / ``WRITE`` / ``FLUSH`` requests from the Linux
+``ublk_drv`` over ``io_uring``, and services them through the **xNVMe**
+asynchronous command interface. Thus, any **xNVMe** backend, such as
+:ref:`sec-backends-upcie` or **io_uring**, becomes usable as an ordinary block
+device.
+
+``REQ_FUA`` and ``REQ_PREFLUSH`` are supported, as are multiple ublk hardware
```
safl, line 13: This claims `REQ_PREFLUSH` support, but `UBLK_IO_F_PREFLUSH` appears nowhere in the code; `dispatch` branches only on `UBLK_IO_OP_FLUSH` and `UBLK_IO_F_FUA`, so a PREFLUSH write is served as a plain write. Since this is a durability claim rather than a feature-list nicety, either handle the flag or drop the line.

Also worth stating in the description: nothing in CI runs any of this. The cijoe step runs `make verify-ramdisk` in containers with no `/dev/ublk-control`, so every qublk case skips, and the two new meson entries cover the CLI plumbing only. The manual Ubuntu run is good, but the conventions ask for the gap to be named.

tools/qublk/main.c
```diff
@@ -0,0 +1,217 @@
+// SPDX-FileCopyrightText: Samsung Electronics Co., Ltd
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#define _GNU_SOURCE
```
safl, line 6: A bare `#define _GNU_SOURCE` here, and in `ctrl.c` and `io.c`, adds three `redefined` warnings to every build, since meson already sets it project-wide. The convention elsewhere in the tree is the `#ifndef _GNU_SOURCE` guard, as in `tools/xnvmeperf/xnvmeperf.c`.

Smaller things while I am here: `dev.depth > QUBLK_MAX_QUEUE_DEPTH` permits exactly 4096 but `xnvme_queue_init` requires under 4096, so `--qdepth 4096` fails only after `ADD_DEV` has created the device; `return -errno` after `xnvme_dev_open()` returns 0 if errno happens to be 0; and a SIGKILLed qublk leaves the ublk device behind with no `del` sub-command to recover it.

tools/qublk/meson.build
safl: Retracting this. I was wrong, and I should have looked harder before raising it.

Upstream `karlowich/qublk` does carry a licence: its root `meson.build` declares `license: 'BSD-3-Clause'`, which is exactly the identifier this PR stamps. My comment was literally true that there is no LICENSE file and no per-file headers, but the conclusion I drew from that, that nothing records who granted the licence, does not follow and is not correct. The upstream author's own build file grants it.

The Samsung copyright line is unremarkable too, since karlowich is a Samsung address and a long-standing contributor here.

So there is no blocker, and I am sorry for putting one on the board. The faithfulness half of the comment still holds: I diffed the imported files against upstream and they match apart from the SPDX header, formatting, and the one functional hunk in `io.c`, which is the point made separately about the import commit message.

docs/tools/qublk/index.rst
safl: Retracting the REQ_PREFLUSH half of this. `UBLK_IO_F_PREFLUSH` does not exist in the kernel UAPI at all, so a ublk server cannot receive such a flag. Because qublk advertises `UBLK_ATTR_VOLATILE_CACHE`, the block layer decomposes a REQ_PREFLUSH write into a separate `UBLK_IO_OP_FLUSH` followed by the write, and qublk serves that FLUSH. The documentation line is correct as written and I was asking you to handle a flag that is never delivered.

The second half of the comment stands: nothing in CI exercises any of this, since the cijoe step runs in containers without `/dev/ublk-control`, and the description would be better for saying so.

tools/qublk/main.c
karlowich: This is a mistake in my original code. Queue Depth naturally should be 32 bits as the spec defines it.
@yonggilsong can you make a separate commit fixing that before this commit?

karlowich: Additionally, perhaps the field should be named qdepth to match xNVMe convention

tools/qublk/main.c
```diff
@@ -36,103 +34,71 @@ lba_shift_of(uint32_t lba_nbytes)
...
-			return 2;
-		}
+	int sig;
+
+	// Options are optional; only override the defaults for the ones actually given
+	if (cli->given[XNVME_CLI_OPT_QDEPTH]) {
+		dev.depth = (uint16_t)cli->args.qdepth;
+	}
+	if (cli->given[XNVME_CLI_OPT_NQUEUES]) {
+		dev.nr_queues = (uint16_t)cli->args.nqueues;
```
karlowich: Same here, let's change `nr_queues` to `nqueues` and `uint32_t` to match what xnvme_cli gives.

tools/qublk/main.c
```diff
@@ -36,103 +34,71 @@ lba_shift_of(uint32_t lba_nbytes)
...
-	if (!is_pow2(dev.depth) || dev.depth > QUBLK_MAX_QUEUE_DEPTH) {
-		fprintf(stderr, "qublk: --depth must be power of 2, <= %u\n",
-			QUBLK_MAX_QUEUE_DEPTH);
-		return 2;
+
+	if (!xnvme_is_pow2(dev.depth) || dev.depth > QUBLK_MAX_QUEUE_DEPTH) {
+		xnvme_cli_perr("Error: --qdepth must be a power of 2 and within limits", -EINVAL);
+		return -EINVAL;
 	}
 	if (dev.nr_queues < 1 || dev.nr_queues > UBLK_MAX_NR_QUEUES) {
```
karlowich: `nr_queues` is unsigned so the `<1` case is impossible to reach.
I know this comes from the original qublk, but since we are changing the surrounding lines we might as well clean this up too.

karlowich (changes requested):
Found a few things on the first pass, see inline.
Also can you squash the two commits: `test(tools/qublk): add cijoe functional tests` and test(tools/qublk): add filesystem-level cijoe tests`
The first introduces lines the next removes.

tools/qublk/main.c
yonggilsong: Done — widened to 32 bits in a separate fix commit right after the import, before the refactor; the refactor now assigns at full width with no narrowing.

yonggilsong: Renamed to qdepth in the same commit.

tools/qublk/main.c
yonggilsong: Done — nqueues, uint32_t, same commit.

tools/qublk/main.c
yonggilsong: Dropped.

tools/qublk/ctrl.c
yonggilsong: Agreed on the ordering hazard. The plan of record is to land #747 first — it has an approval and the FLUSH/FUA support there is what qublk was written against; the description now spells out the silent-FUA consequence of merging in the other order rather than only the failing fs cases.

cijoe/tests/tools/qublk_session.py
yonggilsong: Fixed — the daemon now runs with --dev-id 0 and the session refuses to start when the node already exists.

tools/qublk/io.c
yonggilsong: Fixed — the retry now also submits the ring; poking the xnvme queue cannot cure an exhausted SQ, as you say.

tools/qublk/io.c
yonggilsong: Fixed — the wait is bounded on dev->stop and fails the flush with -ENODEV.

tools/qublk/main.c
yonggilsong: Fixed via the widening commit — both fields are now uint32_t, validated at full width before use, and --qdepth 65600 is rejected.

tools/qublk/main.c
yonggilsong: Reordered to match ubdsrv: STOP_DEV is issued while the queue threads still service, and the drain now dispatches the completions it reaps instead of discarding them.

tools/qublk/main.c
yonggilsong: All addressed: _GNU_SOURCE is guarded in the three sources, --qdepth is capped at half the ublk maximum so xnvme_queue_init cannot fail after ADD_DEV, and the error path no longer returns -errno blind. The missing 'del' recovery sub-command is real and I would prefer to add it as a follow-up rather than grow this PR.

tools/qublk/qublk.h
yonggilsong: I think your reading of the spec is right: Flush is scoped to the namespace and to commands the controller has completed prior to its submission, regardless of which SQ they arrived on. And the block layer only delivers UBLK_IO_OP_FLUSH after the writes it must cover have completed, so by then those writes are completed commands on whichever xnvme queue carried them — one FLUSH on the local queue covers them. Same for FUA: RWF_DSYNC/REQ_FUA makes the write itself durable, other queues have no bearing on it.

@karlowich the fan-out came over from the original qublk — do you recall a device quirk or an ordering concern that motivated it? If not, I will drop the cross-queue machinery (remote-flush pool, MSG_RING encoding and the barrier counters) down to a single local-queue FLUSH. Happy to do it in this PR or as an immediate follow-up, whichever suits the demo timeline better.

safl (commented):
Superseded, please skip this one. Two of the four items here were already
raised and already answered: the Flush scope at `qublk.h:28`, which you agreed
with and which is waiting on karlowich, and the `UBLK_ATTR_FUA` ordering,
which is settled by landing #747 first. Re-reading them costs you time for
nothing.

The two that were actually new, plus a description gap, are in the comment
below this one.

safl (commented):
Three things, none of them a merge blocker, and none corrupts data or crashes
at the defaults of one queue at depth 64. Two are new; the third is a
description gap. Both code items live in the cross-queue machinery, so if that
comes out per the qublk.h thread they come out with it.

1. `ring_entries_for()` in `tools/qublk/io.c:84` asks for `depth * (2 *
   nqueues - 1)` SQEs. `io_uring_setup` rejects anything above
   `IORING_MAX_ENTRIES`, which is 32768, since `IORING_SETUP_CLAMP` is not
   set. `--qdepth` and `--nqueues` are each validated on their own but not
   against each other, so `--nqueues 16 --qdepth 2048` fails to start, as does
   `--nqueues 32 --qdepth 1024`, and so on past roughly `2 * nqueues * qdepth
   > 32768`. What the user sees is `io_uring_queue_init(qN): Invalid
   argument`, which points at the wrong thing, and the `-EINVAL` retry in
   `init_ring()` cannot help since it re-tries with the same entry count. That
   retry exists for kernels without `SINGLE_ISSUER`/`DEFER_TASKRUN`, so it
   makes a sizing error look like a feature probe. The defaults are
   unaffected, so this only bites someone who tunes.

2. `on_xnvme_complete()` and `barrier_complete_iod()` both discard what
   `submit_commit_and_fetch()` returns, and it returns `-EAGAIN` when no SQE
   is available. If that ever happens the ublk request is never committed and
   the block layer waits forever. It is unreachable today only because the
   ring is sized for the worst case, which is the same sizing that produces
   the cliff above, so clamping the entry count to fix that would turn an
   error into a silent hang. They want handling together.

3. The description still says six commits where the head has eight, and it
   still does not say that nothing in CI exercises any of this. That second
   part was raised twice and is the one the conventions ask for explicitly.
   The manual Ubuntu run is good and is already written up; what is missing is
   the sentence saying the cijoe step runs in containers without
   `/dev/ublk-control`, so every qublk case skips there.

Reviewed with a coding agent. The sources were read rather than built or run,
and the entry-count thresholds are arithmetic against the kernel constants
rather than something reproduced. Push back where it is wrong.

safl (dismissed):
LGTM, once #747 is in let's rebase and get green lights on CI ensuring there are no regressions, then we can continue additional changes in follow-ups.

safl (approved):
lgtm, with the list of things to address in follow-ups.

tools/qublk/qublk.h
karlowich: Hmmm, I was having a bunch of problems when I was developing this originally.
I can't recall exactly why I added it, but it is entirely possible that this cross-queue machinery was an attempt at solving a problem caused by something else.
That being said, I don't think we can get rid of all of it, as some operations seem to rely on coming after a flush, so some kind of ordering is required.
We should experiment with how much we can remove without breaking anything.
Let's do that in a follow-up.

karlowich (approved):
Great!
Let's get this in, and then iterate on it in follow-up PRs

PR #756, opened 2026-08-27
feat: Reach GPU memory through an enforcing IOMMU

### LINKED ISSUE #735: upcie-cuda: free the qid in nvme_controller_cuda_delete_io_qpair() rather than in the backend
From review of #732, https://github.com/xnvme/xnvme/pull/732#discussion_r3796101002.

`xnvme_cuda_queue_destroy()` in `lib/upcie/xnvme_be_upcie_cuda_gpu.c` copies the device-side queue-pair back to the host with `cuMemcpyDtoH()` purely to read `qpair.qid`, so it can call `nvme_qid_free()` itself. It does that because `nvme_controller_cuda_delete_io_qpair()` does not return the identifier to the pool, unlike its host counterpart `nvme_controller_delete_io_qpair()`. Without it, a create/destroy cycle drains the identifier space.

The fix belongs in uPCIe rather than here: `nvme_controller_cuda_delete_io_qpair()` should release the qid the way the host path does, and the HIP equivalent alongside it. Once that lands and is vendored, this backend drops the `cuMemcpyDtoH()`, the `cu_err` bookkeeping and the `nvme_qid_free()` call, and the destroy path stops depending on a device-to-host copy that can fail.

Not blocking #732; the workaround is correct, just misplaced.

karlowich:
This vendors https://github.com/safl/upcie/pull/66
Closes #735

`upcie-cuda` and `upcie-hip` accepted only `uio_pci_generic`, since the device-memory
registry puts physical addresses in the translation table, which a device behind
a translating IOMMU cannot use. `vfio-pci` now works too.

The out-of-tree `iommu-map-pa` module inserts the VRAM into the controller's
IOMMU domain, at IOVAs taken from a window the IOAS is told to stay out of, and
the table holds those IOVAs. Nothing else about the backend moves: the same
registry enumerates the addresses and the command paths resolve payloads through
the per-device dmamem as before.

## Several controllers

A mapping reaches one IOMMU domain. The window is divided into a slice per
controller and each maps the heap into its own domain. 
With the default 1 GiB CUDA heap, each controller's slice is ~2 GiB, and the 64 GiB window fits 31 slices.

## Behaviour change

The heap stays process-wide, but the table reaching it does not: under an
enforcing IOMMU a buffer handed to `xnvme_mem_map()` is registered against the
device it was passed, so using it on another controller means registering it
there too. `xnvme_buf_alloc()` buffers are unaffected, since every controller maps the
whole heap. `uio_pci_generic` is unaffected throughout, since physical addresses read
the same from every controller, so one table still serves them all.


safl (commented):
This is in good shape and I see nothing that should block it. It builds clean,
it works on hardware, and it costs nothing: nine drives still reach 16.48M IOPS
on host memory and 15.8M on VRAM under an enforcing IOMMU, and `xnvmeperf verify
--be upcie-cuda` reports 0 mismatches. Two things I would like resolved first,
and neither is a change of approach.

The one that wants a change is the guard refusing a second controller, at
`lib/upcie/xnvme_be_upcie_cuda_dev.c:41` and `lib/upcie/xnvme_be_upcie_hip_dev.c:39`:

    if (g_upcie_cuda_rte.map.alive && strcmp(g_upcie_cuda_rte.map.bdf, bdf)) {

It does not appear to fire. On this branch I ran
`xnvmeperf run 0000:c4:00.0 0000:c3:00.0 --be upcie-cuda` in one process, both
bound to vfio-pci, every IOMMU group reporting DMA and iommu-map-pa loaded, and
both controllers were served, at 1.94M and 1.97M IOPS. No `device memory is
mapped for` line appeared anywhere in the debug output. So either `map.alive` is
not set in the configuration this is meant to cover, or a second controller is
fine and the restriction is unnecessary. I did not chase which, since that is
better answered by whoever wrote it, but it is worth settling before merge: a
guard that does not guard is worse than none, because it reads as a constraint
somebody may later build on.

For a follow-up rather than here: the vendoring commit says v0.9.0, but upstream
uPCIe main is still `VERSION ?= 0.8.0` with no v0.9.0 tag, so this vendors a
version that does not exist yet. Separately, on AMD the import this path depends
on currently lands in GTT rather than VRAM, because the dmabuf-import module
attaches as its own misc device and amdgpu will not hand VRAM to an importer
with no PCI parent; safl/upcie#80 addresses that. Neither affects the CUDA path
here, which does reach device memory.

## Verified

Built and run on an NVIDIA box (RTX A6000, CUDA, kernel 7.0.0-28-generic) with
`iommu.passthrough=0 iommu.strict=1`, all 65 IOMMU groups reporting DMA:
`xnvmeperf verify --be upcie-cuda` clean, `lblk read` clean, nine-drive
throughput as above, and the two-controller case described above.

Not exercised: the HIP path. The AMD box reaches this code, but the underlying
import lands in system memory there for the reason given above, so running it
would not have tested what this change is for. The docs commits were read, not
rendered, and nothing here was tested under uio or type1.

safl (dismissed):
Taking this as is. The guard at lib/upcie/xnvme_be_upcie_cuda_dev.c:41 and lib/upcie/xnvme_be_upcie_hip_dev.c:39 not engaging is a follow-up rather than something to hold this for; the details are in the review above.

Issue #757, opened 2026-08-28
Follow-up work deferred out of #747 and #755


safl:
Work deferred out of #747 (FLUSH and FUA in the Linux async backends) and #755
(qublk), collected so it is not lost once both land. Nothing here blocks
either PR. Items are grouped by the PR they come from; several are only worth
doing in a particular order, which is noted where it matters.

## From #747

1. Hoist the `sqe->flags` assignment in `xnvme_be_linux_liburing_cmd_iov()` to
   after the opcode switch. The readv and writev cases set it before their
   prep-helpers, so on liburing versions whose `io_uring_prep_rw()` clears
   `flags` an SQPOLL vectored read or write loses `IOSQE_FIXED_FILE` and
   addresses process file descriptor 0 rather than the registered file.
   Whether it bites is version-dependent: of the liburing headers available
   for inspection, 2.5 clears `flags` while 2.8, 2.9, 2.11 and 2.14 set only
   `opcode`, `fd`, `off`, `addr` and `len`. Deferred by agreement during
   review.

2. Restore the SQE-leak fix in the same function. The `default:` case returns
   `-ENOSYS` after `io_uring_get_sqe()` has already acquired an SQE, so the
   next `io_uring_submit()` sends it with stale content. This was raised in
   review and fixed by dispatching on the opcode before acquiring the SQE,
   then lost when the series was restructured; only the IOPOLL FLUSH rejection
   stayed hoisted.

3. Add the IOPOLL regression test with a control run, and the `--poll_sq` and
   `--poll_io` options it needs on the ioworker subcommands. The assertion
   wants a matching run without `--poll_io 1` asserted to succeed, so that it
   distinguishes a submission-time IOPOLL rejection from any other nonzero
   exit. Note that POLL_SQ and POLL_IO may need a kernel NVMe driver reload
   that CI does not currently support.

4. The ioworker qdepth fix, split out of #747 as unrelated to flush.

5. Implement NVM FLUSH on the platforms that still reject it, or record the
   gap deliberately. Both FLUSH opcodes fall into `default:` in
   `lib/xnvme_be_freebsd_async.c`, in `_cmd_io` and `_cmd_iov`, under a
   commented-out `aio_fsync` TODO. On Windows, `iocp` and `iocp_th` handle
   `XNVME_SPEC_FS_OPC_FLUSH` but not the NVM opcode, and `ioring` handles only
   READ and WRITE while `xnvme_be_combinations.py` declares an ioring bdev
[...]

PR #758, opened 2026-08-31
fix(be/linux/liburing): SQPOLL and sqe-leak fixes in the vectored path, follow-up to #747

yonggilsong:
The follow-up deferred from #747, per the scope split agreed there: the SQPOLL fix in the vectored io_uring path with its regression test, the sqe-leak fix in cmd_iov, and the qdepth/poll_sq/poll_io options on the ioworker verify subcommands.

Four commits:

1. fix(be/linux/liburing): assign sqe->flags after the prep-helpers in cmd_iov

   On liburing versions where io_uring_prep_rw() zeroes 'flags', the assignment made before the opcode-dispatch is wiped by every prep-helper, so IOSQE_FIXED_FILE never reaches the sqe and an SQPOLL queue targets the actual file descriptor 0 instead of the registered file. As discussed on #747, the assignment is hoisted to just after the dispatch, covering the readv/writev cases as well as FSYNC, and the FSYNC-case re-assignment that compensated locally is dropped.

2. fix(be/linux/liburing): reject unknown opcodes before acquiring an sqe in cmd_iov

   The default-case returned -ENOSYS with the sqe already acquired; io_uring_get_sqe() advances the submission-queue tail, so the abandoned sqe goes out, carrying stale content, with the next io_uring_submit(). Dispatch first, acquire after, mirroring the single-buffer path.

3. tests(ioworker): declare qdepth, poll_sq and poll_io on the verify subcommands

   The worker already reads --qdepth when given, but no subcommand declared it, so it could not actually be passed. --poll_sq/--poll_io give the cijoe suite a handle on the SQPOLL and IOPOLL queue-setups.

4. tests(cijoe): cover SQPOLL submission and IOPOLL flush-rejection in ioworker

   Two io_uring-only bdev cases: a vectored verify on an SQPOLL queue (regression for commit 1), and a verify-flush on an IOPOLL queue, which must fail since the backend rejects FLUSH at submission-time there.

## Verification

Real hardware — Ubuntu 24.04, kernel 6.8.12, as root against a Samsung SSD 970 EVO 1TB (/dev/nvme1n1), all with `--async io_uring`:

- verify --poll_sq 1: pass
- verify --poll_sq 1 --vec-cnt 4 --nlb 7: pass
- verify --poll_sq 1 --vec-cnt 8 --nlb 3 --qdepth 8: pass
- verify-flush --poll_sq 1 --vec-cnt 4 --nlb 7: pass
- verify-flush --direct 1 (control): pass; verify-flush --poll_io 1 --direct 1: rejected at submission-time as intended
- The rootfs of this machine sits on NVMe, so `modprobe -r nvme` refuses with "Module nvme is in use" and test_verify_iopoll skips cleanly without touching driver-state.

QEMU guest (debian-trixie, cijoe test-debian-trixie-iommu_enabled-pcie, test_os narrowed to `-k "sqpoll or iopoll"`): 6 passed, 40 skipped (non-io_uring combinations), 0 failed. test_verify_iopoll performed the driver reload and restore on both io_uring bdev combinations; after the run `/sys/module/nvme/parameters/poll_queues` reads 0 again and a follow-up verify over io_uring on the same device passes.

Ramdisk without root: verify/verify-flush/verify-sync with --qdepth over the emu and thrpool async-interfaces pass, and an out-of-range --qdepth is rejected. clang-format and black clean.

Not exercised: the libaio backend is untouched by this PR.

## Coverage of #757

This PR resolves items 1-4 of #757: the sqe->flags hoist (1), the restored SQE-leak fix (2), the IOPOLL regression test with a control run plus the --poll_sq/--poll_io/--qdepth options (3), and the qdepth fix (4). The driver-reload concern noted in item 3 is handled by the test reloading and restoring the driver itself.


cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -215,3 +215,30 @@ def test_verify_flush(cijoe, device, be_opts, cli_args):
 
     err, _ = cijoe.run(f"xnvme_tests_ioworker verify-flush {cli_args}")
     assert not err
+
+
+@xnvme_parametrize(labels=["bdev"], opts=["be", "sync", "async", "admin"])
+def test_verify_sqpoll_iovec(cijoe, device, be_opts, cli_args):
```
karlowich: Let's take this opportunity to expand the test suite.
Can you add a few more cases to this test to expand the coverage beyond the single iovec case?

Also, it would be great to create a similar test exercising IOPOLL in different configurations. This will require reloading the NVMe driver with `poll_queues` enabled. Since this is only needed by the single test we can make it local and undo it after.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -215,3 +215,30 @@ def test_verify_flush(cijoe, device, be_opts, cli_args):
...
+    if be_opts["async"] != "io_uring":
+        pytest.skip(reason="poll_sq (SQPOLL) is io_uring-only")
+
+    # Vectored I/O on an SQPOLL queue; regression-coverage for the sqe->flags
+    # assignment in cmd_iov, where a dropped IOSQE_FIXED_FILE makes the kernel
+    # treat the registered-file index 0 as the actual file descriptor 0
```
karlowich: I don't think we need this justifaction

karlowich (changes requested):
Looks good, just a few comments about the test

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Done in the re-push. test_verify_sqpoll now runs four variants: the single-buffer path, two vectored configurations (`--vec-cnt 4 --nlb 7`, and `--vec-cnt 8 --nlb 3 --qdepth 8`), and verify-flush over SQPOLL, which reaches the prep-helper path with a non-read/write opcode.

test_verify_iopoll is new and actually exercises IOPOLL: it reloads the nvme driver with `poll_queues=2` locally in the test and restores the original setting afterwards, waiting for the block-device to reappear in both directions. It skips without touching driver-state when the parameter is absent (driver built into the kernel) or when the module cannot be unloaded (e.g. a rootfs sitting on NVMe); when polled queues are already enabled nothing is reloaded. Verified in a QEMU guest (debian-trixie): both io_uring bdev combinations pass, `poll_queues` reads 0 again after the run, and the device remains usable. On a machine whose rootfs is on NVMe the test skips cleanly, checked on real hardware.

The rejection-case also gained the control run asked for in #757 item 3: the same invocation without `--poll_io` is asserted to succeed first. With this re-push the PR covers items 1-4 of #757.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Dropped in the re-push.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
```diff
@@ -1,7 +1,68 @@
...
+        assert not err
+        assert _wait_for_device(cijoe, device["uri"])
+
+        yield
+    finally:
+        cijoe.run("modprobe -r nvme")
+        cijoe.run(
+            "modprobe nvme" + (f" poll_queues={current}" if current.isdigit() else "")
+        )
+        _wait_for_device(cijoe, device["uri"])
```
safl: `_wait_for_device()` is called for `device["uri"]` alone, so the restore
returns once `/dev/nvme0n1` is back while the other namespaces are still
enumerating. `cijoe/tests/logic/` is collected before `cijoe/tests/tools/`, so
`test_fdp.py` and `test_lblk.py` run next, against those namespaces.

Run 33464365426 fails 43 cases there, and re-running the job reproduced them.
Every failure is on a namespace other than the one waited for: `nvme1n1`,
`nvme5n1`, `nvme6n1`, `nvme7n1`, `nvme10n1`, `ng2n1`, `ng5n1`, `ng6n1`,
`ng7n1`, `ng10n1`. There are none on `/dev/nvme0n1` or `/dev/ng0n1`. The same
job passed on `b24ff9dd`.

Wait for every device the configuration declares rather than the one under
test. The return value is also discarded here, where the `try` asserts it, and
the restore drops any nvme module parameter other than `poll_queues`.

safl (commented):
Reviewed with a coding agent. The finding below comes from reading the logs of
run 33464365426 and of its re-run; the driver reload was not reproduced
locally, and nothing else in the change was exercised. Push back where it is
wrong.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
karlowich: Great

cijoe/tests/logic/test_xnvme_tests_ioworker.py
karlowich: Great

cijoe/configs/debian-trixie-iommu_enabled.toml
```diff
@@ -56,10 +56,11 @@ system_args.raw = """\
...
 
-[kvssd]
-binary = "{{ local.env.HOME }}/guests/vfu_kvssd"
-url = "https://github.com/safl/vfio-user-kvssd/releases/download/v0.1.11/vfu_kvssd-0.1.11-x86_64-linux"
-socket = "/tmp/vfu_kvssd.sock"
+# kvssd disabled locally: host qemu lacks the vfio-user-pci device model
+# [kvssd]
+# binary = "{{ local.env.HOME }}/guests/vfu_kvssd"
+# url = "https://github.com/safl/vfio-user-kvssd/releases/download/v0.1.11/vfu_kvssd-0.1.11-x86_64-linux"
+# socket = "/tmp/vfu_kvssd.sock"
```
karlowich: I don't think you meant to push this change. At least, we don't need it.

cijoe/workflows/test-debian-trixie-iommu_enabled-pcie.yaml
```diff
@@ -39,7 +39,7 @@ steps:
 - name: test_os
   uses: core.testrunner
   with:
-    args: '-k "(not ramdisk) and (not pcie) and (not upcie) and (not fabrics)" tests'
+    args: '-k "iopoll or zoned" tests'
```
karlowich: I also don't think you meant to push this change. At least, we don't need it.

karlowich (changes requested):

cijoe/configs/debian-trixie-iommu_enabled.toml
yonggilsong: Right — local scaffolding for my QEMU guest (its host qemu lacks vfio-user-pci) that slipped into an amend. Removed in the re-push.

cijoe/workflows/test-debian-trixie-iommu_enabled-pcie.yaml
yonggilsong: Same story — the narrowed selection I used to validate the driver-reload locally. Removed in the re-push.

cijoe/tests/logic/test_xnvme_tests_ioworker.py
yonggilsong: Right on all three counts; the re-push addresses each.

The restore, and the reload before it, now wait for every device-node the configuration declares — filtered on presence before the unload, since a configuration may declare devices the environment does not carry (my local guest has no KV controller, so a wait on the full declared set would never return). Waiting on existence alone turned out to race too: udev creates the node before the controller is done setting up, and I/O submitted in that window fails. The wait therefore flushes the udev queue and requires every controller under `/sys/class/nvme` to report 'live' before checking for the nodes.

The restore's result is no longer discarded: it is captured and asserted once the context exits, so a failed restore fails the test-case without masking an exception from the body.

The parameter loss is fixed by snapshotting all of `/sys/module/nvme/parameters` before the unload and carrying the full set on both modprobes — `poll_queues` overridden on the way in, everything verbatim on the way out. nvme_core is unaffected; it stays loaded across a `modprobe -r nvme` (held by nvme_fabrics).

Validated in a QEMU guest (debian-trixie, 10 NVMe controllers): test_verify_iopoll passes through the reload (about 5s including both waits), `poll_queues` reads 0 afterwards, and the follow-on zoned/lblk-class cases on the other namespaces — the ones run 33464365426 tripped on — pass after the reload. Not exercised: a config declaring devices on a controller that is slow to enumerate beyond the 30s budget.

karlowich: We are seeing that the CI is taking more than 30s bringing up the devices causing the IOPOLL test to fail.
Additionally, it looks like only /dev/nvme0n1 comes back correctly. The other devices are left in a state that cause following tests to fail.
I would like to take back my request to add the support for poll_queues / IOPOLL to the CI.
I see it is causing unforeseen problems in an otherwise clean PR.
What you are providing is valuable without IOPOLL support, so I propose dropping the failing tests and going back to the original scope of the PR.

yonggilsong: Agreed, and done in the re-push. test_verify_iopoll and the driver-reload machinery are out; what remains is the original scope plus the SQPOLL expansion and the flush-rejection case with its control run — the rejection is submission-time, so it needs no polled queues. The reload experiment stays in this thread's history, and functional IOPOLL coverage goes back to #757 item 3 until CI has an environment for it.

karlowich (approved):
Great!

safl (approved):
LGTM!

PR #759, opened 2026-08-31
Accept --shm_id to join a multi-process group

fmma:
qublk currently opens its device standalone. With no `shm_id` the upcie backend never joins an existing runtime, so qublk cannot forward to a controller a homi server owns. xnvme_cli_to_opts already copies the value. The option table entry was missing.


safl (approved):
lgtm

karlowich (approved):
lgtm

PR #760, opened 2026-08-31
HOMI: preparation (1/3)

safl:
The HOMI implementation has been through a lot of churn and is starting to
settle. These four commits are the part that no longer depends on any of it,
so they are ready to review now rather than waiting for the rest to hold
still.

They open a new chain, xnvme/xnvme#760 here, then xnvme/xnvme#761 and
xnvme/xnvme#762, which replaces xnvme/xnvme#750, xnvme/xnvme#751 and
xnvme/xnvme#752.

CI covers everything here; none of it needs hardware CI does not have.


tests/ctrlr.c
```diff
@@ -151,7 +151,83 @@ test_ctrlr_nslist(struct xnvme_cli *cli)
...
+	ctx = xnvme_cmd_ctx_from_dev(dev);
+	err = xnvme_adm_gfeat_nqueues(&ctx, &nsq, &ncq);
+	if (err) {
+		xnvme_cli_perr("xnvme_adm_gfeat_nqueues()", err);
+		goto exit;
+	}
+
+	/* The helper exists to add the one the spec leaves out, so the check
+	 * that matters is against the raw feature rather than against a range:
+	 * a helper returning NSQA unchanged would still look plausible. */
```
karlowich, line 191: I don't understand this comment. Do we need it?

karlowich: Ah, it is about `+1`

karlowich (changes requested):
This looks good in general, just one inline comment.
Besides this, I'm missing some justification for this commit: "feat(cli): add a --cplane-id option"
It doesn't look like `cplane-id` adds anything other than confusion in the current state.

naddinadja (dismissed):
this looks good to me! :)

tests/ctrlr.c
safl: It is a preparation commit like the other commits on this PR, thus it can be piled onto https://github.com/xnvme/xnvme/pull/761 but the intent is to reduce the size of #761 and thus providing the things that stand a lone here.

safl: > I don't understand this comment. Do we need it?

~~It describes something that deviates from regular parsing logic, thus I appreciate the inline comment here.~~

It describes something that might otherwise appear as an error, thus I think it makes sense. to have a comment for it.

karlowich: The comment is okay as is

naddinadja (dismissed):
lgtm :)

include/libxnvme_cli.h
```diff
@@ -9,8 +9,9 @@
 #define XNVME_CLI_CORE_OPTS                                                                  \
 	{XNVME_CLI_OPT_ORCH_TITLE, XNVME_CLI_SKIP}, {XNVME_CLI_OPT_SUBNQN, XNVME_CLI_LOPT},  \
 		{XNVME_CLI_OPT_HOSTNQN, XNVME_CLI_LOPT}, {XNVME_CLI_OPT_BE, XNVME_CLI_LOPT}, \
+		{XNVME_CLI_OPT_SHM_ID, XNVME_CLI_LOPT},                                      \
 	{                                                                                    \
-		XNVME_CLI_OPT_SHM_ID, XNVME_CLI_LOPT                                         \
+		XNVME_CLI_OPT_HOMI_ID, XNVME_CLI_LOPT                                        \
```
karlowich, line 14: How about completely dropping the CLI version of `SHM_ID`?
I understand that it is worth keeping the `xnvme_opts` field to avoid changing a long standing option, but perhaps the CLI doesn't have to live up to the same level of stability?

docs/tools/homi/index.md
karlowich: FYI: It also says "Multi-process" here

karlowich (commented):
Looks good. Just one suggestion and a note inline.

include/libxnvme_cli.h
safl: It's use has been deprecated for xnvmeperf, homi, and qublk, however I prefer not o have to address any further blast-radius that will come along from entirely removing the cli-option now. If anything it can be put into an issue and addressed in a follow up.

karlowich: All good then

karlowich (dismissed):

PR #761, opened 2026-08-31
HOMI: serve controllers over a socket, not shared memory (2/3)

safl:
Stacked on xnvme/xnvme#760.

Worth a careful look: clients are separate processes that do not synchronise
with each other, so the server serves each connection on its own thread and
reads by blocking. Two locks say what is shared, one over the DMA heap and the
table of queues handed out, and one per controller over its admin queue, whose
completions are reaped in order and carry no record of who asked. Queue creation
needs both and takes them in that order.

Deliberate rather than missed: there is no isolation between clients of one
controller. A client maps BAR0, so it can ring any doorbell and write `CC`, and
all clients of a controller share a DMA domain.

The vendored uPCIe drop comes from safl/upcie#65, which is not released, hence
the `#wip` marker on that commit.

Verification: CI covers the guest suite under both IOMMU configurations, and the
homi selection runs 19 passed and 4 skipped on hardware. The gap is a server
holding several controllers with a client walking the status range; the lab host
declares one controller, so the multi-controller shapes and the enumeration loop
past the first index are unexercised anywhere.


lib/upcie/xnvme_be_upcie_cplane_serve.c
```diff
@@ -0,0 +1,1130 @@
...
+ * Queues one connection may hold, across every controller it uses
+ *
+ * Per client rather than per client and controller, which is what a connection
+ * now is. A process driving several controllers wants a queue on each, often
+ * one per thread, so this has to cover the whole spread and not a single
+ * controller's share of it.
+ */
+#define SERVE_IOQPAIRS_PER_CLIENT 64
+
+#define SERVE_ALLOCS_PER_CLIENT 32
```
fmma: `SERVE_ALLOCS_PER_CLIENT` needs to be raised or redesigned. qublk allocates a buffer per tag at startup. In total: nqueues * qdepth. With default qdepth of 64 it will fail.

PR #764, opened 2026-09-02
HOMI: controller-sharing with and without an iommu; HOMI-PART(2/3)

safl:
A controller behind an IOMMU can't be shared by naming something in the filesystem: the device file is exclusive, and what needs to change hands is the descriptor itself, which only a unix socket can carry. One process (`homi`) opens the controllers and serves them, passing the DMA heap and device and answering requests for queue pairs, buffers and admin commands; a client maps what it's given and submits on its own queues, so the data path never touches the socket. Five commits: the vendored uPCIe carrying the control plane, the client side of it in the backend, the `homi` tool, a systemd unit, and docs. Replaces #761/#762 (the same work split in two, neither reviewed); sits on #769; GPU clients follow in #770 on top of this.

Worth a careful look: the server serves each client connection on its own thread, since clients never synchronize with each other; two locks (the DMA heap and queue table; one per controller over its admin queue) guard what's shared. There is deliberately no isolation between clients of one controller — a client maps the whole of BAR0, so it can ring another client's doorbell. A process that names a server and finds none does not silently fall back to opening the controller itself (`ENOENT`); verified on hardware in both directions. The request-pool alignment fix in `_qpair_from_allocation()` fixed a real 4-8% served-vs-direct regression found via `perf c2c` (a cache-line collision between the pool and the next queue's header). The vendored uPCIe drop is safl/upcie#78 at 4676581, pre-release (`v0.10.0` not yet tagged, hence `#wip`); review the xNVMe side here, the uPCIe side on its own PRs.

CI covers the guest suite under both IOMMU configurations; the systemd unit is checked for build/install substitution only, not for actually starting (done by hand on the lab hosts). Sharing costs nothing measurable across three hosts and every shape tried, within 1-2% of direct in every case but one: a served client whose controllers sit on the other NUMA node from the server's heap loses up to 5%, a known, not-yet-fixed placement cost (per-node heaps are left for after v0.10.0). Two clients contending for one controller isn't covered by the suite.


lib/upcie/xnvme_be_upcie_admin.c
```diff
@@ -23,17 +23,50 @@ xnvme_be_upcie_sync_cmd_admin(struct xnvme_cmd_ctx *ctx, void *dbuf, size_t dbuf
...
+		/* There is one admin queue and it belongs to whoever opened the
+		 * controller. The payload does not travel with the request: the
+		 * command names an address this process registered or was
+		 * given, so the answer lands here without a copy. */
+		struct nvme_cplane_msg msg = {0};
+
+		if (dbuf) {
+			/* Describing the payload is all this needs the scratch
+			 * for; the admin queue it is submitted on belongs to the
+			 * server. */
```
naddinadja: Does this need an "and"? :)
"the admin queue it is submitted on _and_ belongs to the"

PR #769, opened 2026-09-06
uPCIe libconf entries and CUDA example checks; HOMI-PART(1/3)

safl:
Six commits pulled out of #764 (HOMI) that don't depend on it, so they land and review on their own: libconf entries for the uPCIe backends; a CUDA architecture-flag fix for the examples; a fix so the examples fail instead of silently printing `OK` when a kernel never launched; a mutex fixing a process-wide-heap corruption in the uPCIe backends' controller open/close and queue init/term; the IOVA range helper rename agreed in #756 (three user-visible env vars change: `XNVME_UPCIE_GPU_IOVA_{BASE,SIZE,SLICE}` -> `XNVME_UPCIE_IOVA_{BASE,SIZE,SLICE}`, no warning if an old name is still set); and the cijoe perf task (prepare/run/restore) that measured everything quoted in #764 and #770.

Worth a careful look: the heap lock. It fixes a real, reproduced abort (about one run in fifteen on a sixteen-controller host) from parallel controller closes corrupting an unlocked free list — reproduced before the fix, thirty clean runs at unchanged throughput after it. CUDA and HIP queue init/term in #770 take the same lock. The examples fix changes observable behavior: a host whose driver predates the CUDA toolkit now fails `test_cuda_mem_map` instead of silently passing; verified on an RTX A6000, `-arch=native` restores the pass.

Not covered: AMD hardware beyond the libconf entry. CI has no GPU, so the CUDA commits are only compile-checked there; the rename is compile-checked locally, the CUDA/HIP files it touches are covered by CI's CUDA job and the hardware runs under #770.


naddinadja (dismissed):
lgtm :)

karlowich (changes requested):
"fix(be/upcie): serialise the calls that reach the process-wide heaps" is breaking our convention. No other memory interfaces are thread-safe, I propose we just fix xnvmeperf instead.

"feat(cijoe): measure xnvmeperf on a prepared host, shape by shape" This seems very specific to some use case. The commit message doesn't make it clear why this needs to be part of the general xNVMe offerings.

naddinadja (changes requested):
This needs to be rebased on top of #788 - the commits have been changed a bit based on feedback. Otherwise it looks good, I just don't want to approve it and accidentally get the old version of the commits in :)

Issue #766, opened 2026-09-04
test_verify_flush on /dev/ng0n1 with io_uring_cmd is flaky on the trixie-iommu_enabled guest


safl:
`test_verify_flush[be=io_uring_cmd,async=io_uring_cmd]` on `/dev/ng0n1` fails
intermittently on the `trixie-iommu_enabled` guest. It is flaky rather than a
regression: the same commit passes on a re-run with nothing changed.

## Where it has been seen

- `main` @ `f45fdf87`, run [33857213948](https://github.com/xnvme/xnvme/actions/runs/33857213948):
  attempt 1 failed, attempt 2 of the same commit succeeded.
- xnvme/xnvme#756 @ `c5c6df7b`, run
  [33857783667](https://github.com/xnvme/xnvme/actions/runs/33857783667):
  failed the same way, having rebased onto that `main`.

Both in the `verify (debian, trixie-iommu_enabled, pcie, test_os)` job, both
`1 failed, 808 passed, 490 skipped`, and in both the only failing case is the
cdev one:

```
FAILED tests/logic/test_xnvme_tests_ioworker.py::test_verify_flush[uri=/dev/ng0n1,
  be=io_uring_cmd,mem=posix,admin=nvme,sync=nvme,async=io_uring_cmd,label=['cdev']] - assert not 1
```

## What it is not

The first failure landed on the merge of xnvme/xnvme#758, and `main` was green
on the commit before it, so that PR looked responsible. It is not:

- The failing case is `io_uring_cmd`, which submits through
  `xnvme_be_linux_ucmd_iov()` in `lib/xnvme_be_linux_async_ucmd.c`. #758
  changed `cmd_iov` in `lib/xnvme_be_linux_async_liburing.c`, which this path
  never enters.
- The options #758 declared on the verify subcommands in `tests/ioworker.c`
  are read as `cli->given[XNVME_CLI_OPT_QDEPTH] ? ... : QDEPTH_DEF`, and
  `xnvme_cli_to_opts()` gates `poll_io`/`poll_sq` the same way, so declaring
  them changes nothing for an invocation that passes none of them.
- The two testcases #758 added skip unless `async == io_uring`, and sit after
  this one in the file with `random_order: false`.
- On bare metal (Samsung 9100 PRO, kernel-attached) the same invocation passes
  5/5 at `f45fdf87` and 3/3 at `b539f6fc`, and still passes after running both
  sequences #758 added, so neither wedges a later cdev flush.

[...]

PR #771, opened 2026-09-07
tests(cijoe): report the tool output when an ioworker case fails

yonggilsong:
Two fixes for the reporting gap found in #766: a failing ioworker case in CI
surfaced as a bare `assert not 1`, with no way to tell whether the FLUSH was
rejected, timed out, or completed with an error status.

## What

- **ci(cijoe): survive quoted paths in the result-log-dump** — the dump step
  piped `find` into `xargs cat`. `xargs` strips the single-quotes that
  pytest-parametrized directory-names carry (e.g. `label=['cdev']`), so `cat`
  was handed paths that do not exist and the failing command's output never
  made it into the log — the `cat: '.../cmd_02.output': No such file or
  directory` seen in #766. The file existed; the dump could not read it.
  Loop over the paths instead, and print each file's name before its content
  so a dump of many files can be attributed.
- **tests(cijoe): report the tool output when an ioworker assertion fails** —
  hand the command output to every assertion in
  `tests/logic/test_xnvme_tests_ioworker.py`, so the pytest failure itself
  carries what the tool printed, independent of the artifact steps. Combined
  with 13f582fa (print the completion when the FLUSH fails), the status code
  is in the failure message.

Together with 13f582fa and d707eb3c (both on `main`), this covers the
"failure records nothing" item of #766.

## Verification

The three failure modes produce distinct output, verified by provoking each:

| Failure mode | Output in the assertion message |
| --- | --- |
| FLUSH rejected at submission (IOPOLL) | `# ERR: 'iowork_flush()': {errno: -38}` — no completion print |
| FLUSH completed with error status (incl. a timed-out command aborted by the driver) | `# ERR: 'on_flush_completion()'` + `xnvme_cmd_ctx: {cdw0: .., sc: .., sct: ..}` |
| FUA-write completed with error status | `# ERR: 'on_completion()'` + `xnvme_cmd_ctx: {..}` + non-zero `ecount` in the stats |

Systems and runs:

- **cijoe, ramdisk** (Linux 6.8, x86_64, no root): the test-ramdisk selection
  (`-k "ramdisk and not hugepage"`) over the modified file — 18 passed,
  4 skipped (the io_uring-only cases), 0 failed.
- **Real NVMe** (Samsung 970 EVO 1TB, kernel-attached, `--be io_uring` /
  `--async io_uring --direct 1`): `verify-flush` control passes; the four
  SQPOLL `verify`/`verify-flush` variants pass; `verify-flush --poll_io 1`
  fails at submission-time with the ENOSYS line above.
- **dm-flakey over a loop device** (fault injection): a flakey window during
  the write phase produces the FUA-write signature; flipping the table to a
  down-interval right before `iowork_flush()` (paused under gdb) fails the
  FLUSH alone and produces the `on_flush_completion()` signature, with all
  32768 writes completed clean.
- The `xargs` quoting failure reproduces standalone: `find | xargs cat` on a
  path containing `label=['cdev']` reports No-such-file; the loop prints it.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -808,7 +808,11 @@ jobs:
 
     - name: CIJOE, result-log-dump on error
       if: failure()
-      run: find cijoe -name "*.output" | xargs cat
+      run: |
+        find cijoe -name "*.output" | while read -r fpath; do
+          echo "=== ${fpath}"
+          cat "${fpath}"
+        done
 
     - name: CIJOE, upload report.html
       if: always()
@@ -912,7 +916,11 @@ jobs:
 
     - name: CIJOE, result-log-dump on error
       if: failure()
-      run: find cijoe -name "*.output" | xargs cat
+      run: |
+        find cijoe -name "*.output" | while read -r fpath; do
+          echo "=== ${fpath}"
+          cat "${fpath}"
+        done
 
     - name: CIJOE, upload report.html
       if: always()
@@ -999,7 +1007,11 @@ jobs:
 
     - name: CIJOE, result-log-dump on error
       if: failure()
-      run: find cijoe -name "*.output" | xargs cat
+      run: |
+        find cijoe -name "*.output" | while read -r fpath; do
+          echo "=== ${fpath}"
+          cat "${fpath}"
+        done
 
     - name: CIJOE, upload report.html
       if: always()
--- a/cijoe/tests/logic/test_xnvme_tests_ioworker.py
+++ b/cijoe/tests/logic/test_xnvme_tests_ioworker.py
@@ -5,30 +5,30 @@ from ..conftest import get_osname, xnvme_parametrize
 
 @xnvme_parametrize(labels=["dev"], opts=["be", "sync", "async", "admin"])
 def test_verify(cijoe, device, be_opts, cli_args):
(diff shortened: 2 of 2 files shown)


safl (dismissed):
lgtm

.github/workflows/verify.yml
```diff
@@ -806,9 +806,16 @@ jobs:
         echo "::endgroup::"
         cp /tmp/vfu_kvssd.log "${dst}-vfu_kvssd.log" 2>/dev/null || true
 
+    # Piping to xargs would strip the single-quotes that pytest-parametrized
+    # directory-names carry (e.g. "label=['cdev']"), so cat is handed paths
+    # that do not exist; loop instead, and name each file before its content
```
naddinadja: I would delete these comments (there are three of them) - the info is already in the commit message, and the comments don't make sense on their own without the history of these steps using `xargs` before this commit :)

naddinadja (changes requested):
This is great, just a small change request on the inline code comments :)

.github/workflows/verify.yml
yonggilsong: Agreed — they were talking to the reviewer, not the next reader. Dropped
all three; the rationale lives in the commit message.

naddinadja (approved):
great! :)

safl (approved):
lgtm

PR #772, opened 2026-09-07
ci(cijoe): fail on a bad QEMU feature probe, wait longer for the benchmark guests

safl:
Fixes the two jobs red on `main` since d707eb3c, neither of which is caused
by that commit.

The test_python job on trixie-iommu_enabled got a guest with four
controllers instead of eleven, because the qemu feature probe in
`xnvme_guest_start_nvme.py` failed and the script quietly built the guest
without ZNS, FDP and PI. That shifts every BDF after the first, so the suite
opened the wrong devices and hung in `xnvme-driver reset` on a vfio group
an spdk test had left open. The sibling jobs of the same run got eleven
controllers from the same image, so the probe is flaky rather than broken.
The first commit makes a failed or incredible probe fail the step.

The FreeBSD benchmark guest boots completely but reaches "login:" a little
past the 180 s that cijoe's `qemu.guest_start` hard-codes, and only when
another guest boots alongside it on the host. The second commit adds
`xnvme_guest_start`, the same step with the wait as an argument, and gives
the benchmark guests 600 s.

Worth a careful look: the credibility check in `detect_qemu_nvme_features()`
requires zoned and pi, not fdp, so the fork still passes; and the
600 s is a ceiling, not a delay, since `is_up()` returns as soon as the
prompt appears.

## Verification

The three probe outcomes were exercised locally against a missing binary,
a non-qemu binary and the system qemu, and take the intended paths. The
guest start itself is exercised only by this CI run: the trixie jobs cover
the probe on the real image, and benchmark-latency freebsd covers the longer
wait. The flaky probe was not reproduced, so the run proves the happy path,
not the failure path, and the FreeBSD fix rests on the timing read from the
logs of three runs rather than on a measured boot.

### DIFF
--- a/cijoe/scripts/xnvme_guest_start.py
+++ b/cijoe/scripts/xnvme_guest_start.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+"""
+Start a qemu guest and wait for its login prompt
+================================================
+
+The same as cijoe's own qemu.guest_start, except that the time to wait for the
+guest is an argument rather than a fixed 180 seconds. A first boot of the
+FreeBSD image generates its host keys and lands within seconds of that limit
+on an idle host, and past it when another guest is booting alongside.
+
+Retargetable: False
+-------------------
+"""
+import errno
+import logging as log
+from argparse import ArgumentParser
+
+from cijoe.qemu.wrapper import Guest
+
+
+def add_args(parser: ArgumentParser):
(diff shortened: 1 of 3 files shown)


naddinadja (dismissed):
Looks good to me :)

naddinadja (approved):
lgtm :)

karlowich (approved):
lgtm

PR #774, opened 2026-09-08
qublk: correct the FLUSH scope and follow-ups from the #755 review

yonggilsong:
This picks up the qublk items deferred to #757 out of the #755 review: items 9 through 13. The headline change reverses the cross-queue FLUSH fan-out; as discussed offline with @karlowich, going with the simplification is the agreed direction.

Four commits, each building and passing on its own:

1. refactor(tools/qublk): flush only the local queue (item 9, and item 10 with it)

The fan-out assumed a FLUSH covers only writes submitted on the same NVMe submission queue. The NVMe Base Specification scopes it wider: a flush applies to all commands completed by the controller prior to the submission of the flush, with no submission-queue qualifier (Base Spec 2.3, section 7.2 "Flush command"). The block layer, in turn, issues a flush only after the writes it covers have completed. A single FLUSH on the queue that received the ublk request therefore gives the same guarantee the fan-out did. This removes the remote-flush pool, the MSG_RING encoding and handlers, and the barrier counters (-355 lines); FLUSH and FUA writes dispatch like any other command. With at most one COMMIT_AND_FETCH per local tag between submits, the ring drops the depth * (2 * nqueues - 1) worst-case sizing, which also dissolves item 10: the io_uring_setup() entry-limit overflow that --nqueues 16 --qdepth 2048 used to hit can no longer be constructed.

2. fix(tools/qublk): stop the device when a commit SQE cannot be queued (item 11)

on_xnvme_complete() discarded the return of submit_commit_and_fetch(). Unreachable with the ring sized at one SQE per tag, but if the invariant is ever broken, fail loudly instead of leaving a ublk request uncommitted and the block layer waiting forever.

3. fix(tools/qublk): advertise FUA only when the backend honours it (item 12)

UBLK_ATTR_FUA was gated on the namespace reporting a volatile write cache, but whether the fua bit reaches the device depends on the backend: io_uring and libaio map it to RWF_DSYNC, the passthru backends carry the command verbatim, while psync- and block-backed paths issue a plain pwrite() and drop it. The attribute now derives from the resolved async/sync implementations; UBLK_ATTR_VOLATILE_CACHE stays gated on the write cache alone.

4. feat(tools/qublk): add a del sub-command (item 13)

qublk del --dev-id N recovers a device left behind by a killed server: STOP_DEV so the kernel aborts requests pending against the dead daemon, then DEL_DEV. Documented in docs/tools/qublk.

## Verification

QEMU guest (debian trixie, kernel 6.12.94, /dev/nvme3n1), since the cijoe qublk suite still needs a ublk-capable CI environment (items 14/15, not in scope here):

- io_uring backend: /sys/block/ublkbN/queue/fua reads 1; 4 MiB O_DIRECT write/read round-trip compares equal; clean teardown on SIGTERM.
- FUA gating: thrpool_bdev (sync=block) advertises fua 0, thrpool (sync=nvme) advertises fua 1 — matching what each path delivers.
- Multi-queue FLUSH path: with --nqueues 4, mkfs.ext4, mount, 8 MiB write with fsync, sync, umount, remount and read back all pass.
- del: after kill -9 the control-side device remains (the kernel tears down the block node when the daemon dies); qublk del --dev-id N removes it, and a nonexistent id fails with No such device.
- All four commits build individually; clang-format clean.

After this lands, the #757 items still open are 3 (the live-IOPOLL half), 5, 6, 7, 8, 14 and 15.

### DIFF
--- a/tools/qublk/io.c
+++ b/tools/qublk/io.c
@@ -130,7 +130,14 @@ on_xnvme_complete(struct xnvme_cmd_ctx *ctx, void *opaque)
 	}
 
 	xnvme_queue_put_cmd_ctx(q->xq, ctx);
-	submit_commit_and_fetch(q, io, result);
+	// The ring holds one SQE per tag and each tag has at most one commit
+	// pending, so this cannot fail by sizing; if it does anyway, stop the
+	// device rather than leave the request uncommitted and the block
+	// layer waiting on it forever
+	if (submit_commit_and_fetch(q, io, result) < 0) {
+		fprintf(stderr, "qublk: q%d tag %u: no SQE for commit\n", q->q_id, io->tag);
+		q->dev->stop = 1;
+	}
 }
 
 static int
--- a/tools/qublk/io.c
+++ b/tools/qublk/io.c
@@ -25,36 +25,6 @@
 #define IORING_SETUP_DEFER_TASKRUN (1U << 13)
 #endif
 
-/*
- * io_uring CQE user_data convention:
- *   bit 63 clear -> ublk tag CQE; user_data is the tag (< depth)
- *   bit 63 set   -> cross-queue MSG_RING; encoded as:
- *                   bits 62-60: type (QUBLK_MSG_*)
- *                   bits 31-16: originating q_id
- *                   bits 15-0 : originating tag
- * The MSG_RING payload uses sqe->off (delivered as target cqe->user_data) for
- * the encoded id and sqe->len (delivered as target cqe->res) for status.
- */
-#define QUBLK_UD_MSG_BIT (1ULL << 63)
-#define QUBLK_MSG_TYPE_SHIFT 60
-#define QUBLK_MSG_TYPE_MASK 0x7ULL
-#define QUBLK_MSG_QID_SHIFT 16
-#define QUBLK_MSG_QID_MASK 0xFFFFULL
-#define QUBLK_MSG_TAG_MASK 0xFFFFULL
-
-enum {
-	QUBLK_MSG_SOURCE = 0,     /* source-side CQE for our outgoing MSG_RING */
-	QUBLK_MSG_DO_FLUSH = 1,   /* recipient: issue NVMe FLUSH for (orig_q, orig_tag) */
-	QUBLK_MSG_FLUSH_DONE = 2, /* originator: a remote FLUSH completed (res = status) */
-};
-
-static inline uint64_t
-ud_msg(unsigned type, uint16_t qid, uint16_t tag)
-{
-	return QUBLK_UD_MSG_BIT | ((uint64_t)type << QUBLK_MSG_TYPE_SHIFT) |
-	       ((uint64_t)qid << QUBLK_MSG_QID_SHIFT) | (uint64_t)tag;
-}
-
 static size_t
 page_round_up(size_t v)
 {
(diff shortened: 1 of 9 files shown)


tools/qublk/main.c
```diff
@@ -25,6 +25,33 @@
...
+/*
+ * Whether the resolved backend delivers the NVMe 'fua' bit to the device:
+ * io_uring and libaio map it to RWF_DSYNC, the passthru backends carry the
+ * command verbatim, and emu/thrpool delegate to the sync layer, which does so
+ * only on the NVMe-ioctl path -- psync and block issue a plain pwrite().
+ */
+static uint8_t
+backend_honours_fua(const struct xnvme_dev *xdev)
+{
+	static const char *honours[] = {"io_uring", "libaio",  "io_uring_cmd",
```
safl: clang-format rejects this initialiser; this is the `source-format-check` failure.

tools/qublk/main.c
```diff
@@ -25,6 +25,33 @@
...
+	if (!opts->async) {
+		return 0;
+	}
+	for (size_t i = 0; i < sizeof(honours) / sizeof(*honours); i++) {
+		if (!strcmp(opts->async, honours[i])) {
+			return 1;
+		}
+	}
+	if (!strcmp(opts->async, "emu") || !strcmp(opts->async, "thrpool")) {
+		return opts->sync && !strcmp(opts->sync, "nvme");
```
safl: For `emu` and `thrpool` only `sync == "nvme"` counts, but the `io_uring_cmd`, `spdk`, `libvfn` and `upcie` sync mixins carry the command verbatim as well. Check `opts->sync` against that set, otherwise `--async thrpool --sync io_uring_cmd` advertises no FUA while delivering it.

tools/qublk/io.c
```diff
@@ -738,9 +431,9 @@ io_loop(struct qublk_queue *q)
 		 * NVMe completions can only arrive while commands are in flight,
 		 * and the upcie backend has no completion fd to wait on -- so we
 		 * must busy-poll the xnvme queue whenever it has outstanding work.
-		 * When it is empty, the only events that can wake us are a new
-		 * ublk request (FETCH) or a peer queue's barrier MSG_RING, both of
-		 * which post to this ring's fd; block on it instead of spinning so
+		 * When it is empty, the only event that can wake us is a new
+		 * ublk request (FETCH), which posts to this ring's fd; block on
+		 * it instead of spinning so
```
safl: Cut mid-sentence when the MSG_RING clause was removed: "block on / it instead of spinning so / an idle queue". Reflow the paragraph.

tools/qublk/main.c
```diff
@@ -191,6 +219,40 @@ sub_run(struct xnvme_cli *cli)
 	return -EIO;
 }
 
+static int
```
safl, line 237: No test for `del`. The manual check in the description is the test: start `run`, SIGKILL it, `del --dev-id N`, assert the device is gone and that a nonexistent id fails. `cijoe/tests/tools/test_qublk.py` has the session helper for it; add it now so it is in place when CI can run the suite.

safl (commented):
One thing blocks merging: `source-format-check` is red on the `honours[]` initialiser in `tools/qublk/main.c`; `make format-all` clears it.

Three inline comments besides that, one of them a missing test.

Reviewed with a coding agent. Read the four commits, the resulting `io.c`, `main.c`, `ctrl.c` and `qublk.h`, the mixin identifiers the FUA table is compared against, and the io_loop wait. Nothing built or run; no ublk device was exercised. Push back where it is wrong.

tools/qublk/main.c
yonggilsong: Reshaped the initialisers to the brace-per-line form, which formats identically across clang-format versions; source-format-check should be green now.

tools/qublk/main.c
yonggilsong: Right — extended the check to the passthru sync set. One correction to the list: there is no io_uring_cmd sync implementation (the sync interfaces in the tree are psync, nvme, block, spdk, libvfn, upcie and ramdisk), so the set is nvme, spdk, libvfn and upcie. Verified in a guest: --be emu (sync=nvme) and --be thrpool now advertise fua 1, --be thrpool_bdev (sync=block) stays at 0.

tools/qublk/io.c
yonggilsong: Reflowed.

tools/qublk/main.c
yonggilsong: Added test_del_leftover to cijoe/tests/tools/test_qublk.py: serve, SIGKILL the server, qublk del --dev-id 0, assert the control char-device is gone, and assert that a nonexistent id fails. It follows the session-script conventions from qublk_session.py (newline joins, mktemp log, preexisting-device guard). The same sequence is verified by hand in a guest; the case will run for real once the suite has a ublk-capable CI environment (items 14/15).

safl (approved):
lgtm

karlowich (approved):
lgtm

PR #776, opened 2026-09-09
fix(be/linux): correct the debug message in the zone-capacity fallback

yonggilsong:
Follow-up to the #773 review (https://github.com/xnvme/xnvme/pull/773#discussion_r3955226906): since the guard became a meson probe for the capacity member of struct blk_zone, the fallback branch is taken when the build headers lack the member, not when an ioctl is missing. The message now says so, with the wording @safl suggested.

Verified by building with the guard inverted so the fallback branch actually compiles, then building normally; clang-format clean.


safl (approved):
lgtm

naddinadja (approved):
lgtm :)

karlowich (approved):
lgtm

PR #777, opened 2026-09-09
ci(cijoe): install the test-suite dependencies when provisioning a guest

yonggilsong:
Item 15 of #757 turns out to be further along than the issue text says: the block-level qublk cases already run and pass on the trixie-iommu_enabled guest (27 passed on main run 34320608759). What still skips is the filesystem level: all 45 test_qublk_fs.py cases, because the guest has no mkfs.xfs.

Reworked per the review: instead of adding xfsprogs to toolbox/pkgs (whose lists feed the toolchain documentation and describe build/run dependencies of the library and tools), a guarded test_deps step now follows xnvme_build_prep in provision-using-tgz.yaml and its MAAS twin, exactly as suggested. The apt-get guard covers the FreeBSD guests the same workflows provision.

Verified against a local guest booted from the same nosi image: with xfsprogs removed, running only the test_deps step of provision-using-tgz.yaml installs it (skipped steps confirmed inert -- the guest was not re-imaged), and mkfs.xfs is present afterwards, which is exactly the guard test_qublk_fs.py skips on. This PR's own verify jobs exercise the step for real.

### DIFF
--- a/cijoe/workflows/provision-maas-using-tgz.yaml
+++ b/cijoe/workflows/provision-maas-using-tgz.yaml
@@ -55,9 +55,11 @@ steps:
   run: |
     hostname
 
+# xfsprogs is a test-suite dependency, not a build/run one: the
+# filesystem-level qublk cases format with XFS
 - name: packages_extra
   run: |
-    apt-get install -qy dmidecode patchelf || true
+    apt-get install -qy dmidecode patchelf xfsprogs || true
     pkg install -y dmidecode patchelf || true
 
 - name: xnvme_source_sync
--- a/cijoe/workflows/provision-using-tgz.yaml
+++ b/cijoe/workflows/provision-using-tgz.yaml
@@ -55,6 +55,13 @@ steps:
   with:
     xnvme_source: '{{ config.xnvme.repository.sync.remote_path }}'
 
+# Dependencies of the test-suite, as opposed to the build/run dependencies
+# installed by xnvme_build_prep; the guard is for the FreeBSD guests this
+# workflow also provisions. The filesystem-level qublk cases format with XFS.
+- name: test_deps
+  run: |
+    if command -v apt-get > /dev/null; then apt-get -qy install xfsprogs; fi
+
 - name: xnvme_build
   uses: xnvme_build
   with:
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -904,7 +904,7 @@ jobs:
         --monitor \
         --config "configs/${{ matrix.hostname }}.toml" \
         --output "${{ github.job }}-${{ matrix.hostname }}-provision-xnvme" \
-        xnvme_source_sync xnvme_build_prep xnvme_build xnvme_install \
+        xnvme_source_sync xnvme_build_prep test_deps xnvme_build xnvme_install \
         ldconfig xnvme_bindings_py_install_tgz fio_prep
 
     - name: CIJOE, run tests
--- a/Makefile
+++ b/Makefile
@@ -246,6 +246,7 @@ guest-provision:
 		-l \
 		xnvme_source_sync \
 		xnvme_build_prep \
+		test_deps \
 		xnvme_build \
 		xnvme_install \
 		ldconfig \


toolbox/pkgs/emitter/deps.yaml
```diff
@@ -455,6 +455,9 @@ platforms:
   alt: []
   aux:
   - [sys, [autoconf, automake, build-essential, libtool]]
+  # The filesystem-level qublk test-cases format with XFS; without mkfs.xfs
+  # they skip on the CI guest provisioned with this script
+  - [sys, [xfsprogs]]
```
safl: This might not be the right place to put this: what is here ultimately produces the toolchain documentation (https://xnvme.io/toolchain/linux/index.html#debian-stable-trixie), the packages needed for building the xNVMe library and tools and running them, not for testing them.

For now, install it in the provisioning workflow instead, right after `xnvme_build_prep` in `provision-using-tgz.yaml` and its MAAS twin:

```yaml
- name: test_deps
  run: |
    if command -v apt-get > /dev/null; then apt-get -qy install xfsprogs; fi
```

The guard is for the FreeBSD guests the same workflow provisions.

safl (commented):
Not the right place, see the inline comment; a provisioning step covers the need for now.

The change asked for is small: drop the two toolchain edits and add one step to `cijoe/workflows/provision-using-tgz.yaml` and `provision-maas-using-tgz.yaml`, after `xnvme_build_prep`. Without `|| true`, since a failed install would leave the fs cases skipping silently again, which is what this fixes.

The tidy version is a `testing` class in `deps.yaml` rendered into a `-test.sh` sibling of the `-build.sh` scripts, which the step above would then call. If you are up for it, then transform this PR to do that instead.

Read against the diff; nothing was run.

toolbox/pkgs/emitter/deps.yaml
yonggilsong: Agreed — moved it out of toolbox/pkgs and into a guarded test_deps step after xnvme_build_prep in provision-using-tgz.yaml and the MAAS twin, as suggested. Verified on a local guest: with xfsprogs removed, running only the test_deps step installs it and mkfs.xfs is present afterwards; the skipped steps stay inert (the guest is not re-imaged). PR title/body updated to match.

safl (commented):
One placement point, inline. Otherwise this is the shape asked for; whether the fs cases now run on the trixie guest is what the test_os job of this run will show.

Read against the diff and the run's logs; nothing was run here.

cijoe/workflows/provision-maas-using-tgz.yaml
```diff
@@ -70,6 +70,13 @@ steps:
   with:
     xnvme_source: '{{ config.xnvme.repository.sync.remote_path }}'
 
+# Dependencies of the test-suite, as opposed to the build/run dependencies
+# installed by xnvme_build_prep; the guard is for the FreeBSD guests this
+# workflow also provisions. The filesystem-level qublk cases format with XFS.
+- name: test_deps
```
safl: This workflow already has a step for exactly this, `packages_extra` at line 57, which installs `dmidecode patchelf` through both `apt-get` and `pkg`. Add `xfsprogs` to its apt line instead of a second step ten lines below with a different guard; two of these next to each other is how a third one appears. Keep the comment on why it is a test dependency there.

```yaml
- name: packages_extra
  run: |
    apt-get install -qy dmidecode patchelf xfsprogs || true
    pkg install -y dmidecode patchelf || true
```

The non-MAAS workflow has no such step, so `test_deps` there stays as it is.

cijoe/workflows/provision-using-tgz.yaml
```diff
@@ -55,6 +55,13 @@ steps:
   with:
     xnvme_source: '{{ config.xnvme.repository.sync.remote_path }}'
 
+# Dependencies of the test-suite, as opposed to the build/run dependencies
+# installed by xnvme_build_prep; the guard is for the FreeBSD guests this
+# workflow also provisions. The filesystem-level qublk cases format with XFS.
+- name: test_deps
```
safl, line 61: This step is skipped in CI. The verify job runs this workflow with an explicit step list, `.github/workflows/verify.yml` line 924:

```
xnvme_source_sync xnvme_build_prep xnvme_build xnvme_install \
ldconfig xnvme_bindings_py_install_tgz fio_prep
```

and cijoe skips every step not named there (`cli.py:434`, `if args.step and step["name"] not in args.step`), which is why the provisioning log says `step(test_deps) : skipped`, the same as it does for `guest_check`. On this PR's own trixie test_os job all `test_qublk_fs.py` cases are still `s`.

Add `test_deps` to that list after `xnvme_build_prep`. The two MAAS invocations at lines 1013 and 1097 pass no list, so the step runs there as it is; the Windows one at 1191 has its own list and rightly leaves it out.

safl (changes requested):
Not there yet: the step does not run in the guest CI job, so the fs cases still skip. See the inline comment for the one-word fix. Once the trixie test_os job of a run shows the 45 `test_qublk_fs.py` cases passing rather than skipping, this is good to merge.

Read against the diff and the logs of run 34416450802, where the trixie test_os job reports every `test_qublk_fs.py` case as `s` and `test_qublk.py` as passing; its three failures are the ZNS fio cases on the bdev path, the known flake, not this change.

cijoe/workflows/provision-maas-using-tgz.yaml
yonggilsong: Folded xfsprogs into the existing packages_extra apt line and dropped the second step, with the test-dependency rationale kept as a comment there.

cijoe/workflows/provision-using-tgz.yaml
yonggilsong: Right — my local check ran the step by name, which is exactly what hid this. Added test_deps after xnvme_build_prep in both explicit step lists: the one in 'make guest-provision' (which is what the trixie guest jobs run via make verify-guest) and the verify-physical provisioning invocation in verify.yml; the guard makes it inert where apt-get is absent. Re-verified locally through the -l step-list path this time: with xfsprogs removed, the named test_deps step installs it. The trixie test_os job of this run should now show the 45 fs cases running.

safl (commented):
lgtm

safl (approved):
lgtm

karlowich (approved):
lgtm

PR #778, opened 2026-09-09
fix(be): keep the nil stub from winning ramdisk auto-selection, and repair the meson test-suite

yonggilsong:
This picks up item 6 of #757: the default meson test-suite fails on every run, on unmodified main, and no CI job notices.

Root cause: every platform lists g_xnvme_be_ramdisk_nil first in its ramdisk group, so opening a ramdisk URI with no explicit backend options auto-selects the nil async interface -- a no-op stub that completes nothing. The async_intf and ioworker entries open "1GB" exactly that way, so 11 of the 14 standing failures are this one selection. The same arrays already handle this correctly for the NVMe char-device group, where emu probes first and the nil config sits last; commit 1 mirrors that order (emu, thrpool, nil) for the ramdisk group on all four platforms. Explicit selection is unaffected -- the cijoe combinations address ramdisk_emu and ramdisk_thrpool by name.

Chasing the remaining non-root failure surfaced two more small defects, fixed separately:

- fix(cli): --pract is declared as taking a number, but the handler set pract to true on mere presence, so "--pract 0" enabled PRACT instead of disabling it.
- tests(tools): the "prchk not set" lblk entry passed "--pract 0x0", which never parsed (the option is base-10) -- and pract 0 cannot run on the meson ramdisk at all, since the host-side PI path is rejected on a namespace without a PI format. The entry now passes "--pract 1" and exercises the same plain write-read path as its prchk siblings.

## Verification

meson test on main: 43 pass / 14 fail. With these three commits: 55 pass / 2 fail; the two remaining are the enum entries, which need root for device enumeration. xnvme library-info confirms auto-selection now resolves ramdisk_emu, and the ioworker verify entries do real data round-trips (write, read-back, compare) instead of EIO. Each commit builds individually; clang-format clean.

## Open question

Since no workflow runs meson test, this suite could regress again without anyone noticing. Do you want a CI job for it (it needs no privileges beyond what the failing enum entries want, which could stay excluded), or is the cijoe matrix considered the authoritative gate? Happy to follow up either way.

### DIFF
--- a/lib/xnvme_cli.c
+++ b/lib/xnvme_cli.c
@@ -1783,7 +1783,7 @@ xnvme_cli_assign_arg(struct xnvme_cli *cli, struct xnvme_cli_opt_attr *opt_attr,
 		args->pid = num;
 		break;
 	case XNVME_CLI_OPT_PRACT:
-		args->pract = true;
+		args->pract = num != 0;
 		break;
 	case XNVME_CLI_OPT_PRCHK:
 		args->prchk = num;
--- a/lib/xnvme_platform_freebsd.c
+++ b/lib/xnvme_platform_freebsd.c
@@ -273,9 +273,9 @@ struct xnvme_platform g_xnvme_platform_freebsd = {
 			&g_xnvme_be_freebsd_emu_nvme,
 			&g_xnvme_be_freebsd_nil_nvme,
 #ifdef XNVME_BE_RAMDISK_ENABLED
-			&g_xnvme_be_ramdisk_nil,
-			&g_xnvme_be_ramdisk_thrpool,
 			&g_xnvme_be_ramdisk_emu,
+			&g_xnvme_be_ramdisk_thrpool,
+			&g_xnvme_be_ramdisk_nil,
 #endif
 			NULL,
 		},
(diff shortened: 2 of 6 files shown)


safl (approved):
Good to merge with one addition. The three diagnoses hold against the code: numeric options parse base 10, so `--pract 0x0` never parsed; the handler ignored the value; and every platform array probed `g_xnvme_be_ramdisk_nil` first for the ramdisk group while the char-device group already had emu first and nil last.

Nothing to anchor, since the file is outside the diff. On the docs: the array excerpt in `docs/background/architecture/index.md` at line 1009 still lists `g_xnvme_be_ramdisk_nil` first; reorder it in the first commit, which is what makes the excerpt wrong. The config table above it names configs rather than the probe order and can stay.

On the open question: yes to a `meson test` step in the build-linux job, with the two enum entries excluded. The cijoe matrix never opens a ramdisk without naming its backend, which is why eleven of the fourteen failures went unseen; this is the only gate that would have caught it. Separate PR.

Read against origin/main at 210640cb, with the parsing and handler code in `xnvme_cli.c`; nothing was built or run.

naddinadja (approved):
lgtm :)

karlowich (approved):
lgtm.

To me, this is ready to merge.
I agree that having `meson test` run in the CI would be good, but I think we can have this in a separate PR

PR #781, opened 2026-09-11
tests(cijoe): reset a zoned block device through the block layer before writing it

safl:
One commit, for the `test_fio_engine_zns` failures that hit test_os on
trixie-iommu_enabled depending on test order, most recently three
parametrizations in one run on #770 (run 34536972409) with `fio: io_u error
on file /dev/nvme1n1: Input/output error: write offset=0` and its neighbours,
while the same job passed on main and on #777 the same hour.

The block layer since Linux 6.10 keeps a write pointer per zone in a zone
write plug and `blk_zone_wplug_prepare_bio()` fails a write that does not
start at it before the device sees the command. The suite resets and writes
zones of the same namespace through `/dev/ng1n1` and through user-space
drivers, which the block layer does not see, so after such a reset the
device reports the zone empty while the kernel still holds the pointer from
the last write it issued itself, and fio in zbd mode, which reads the write
pointers from the device and finds nothing to reset, has its first write
refused with EIO. What to look at is where the remedy sits:
`zoned_bdev_reset()` resets a zoned block device through the block device
itself, setting the kernel's pointers to zero along with the device's, and
the fio ZNS test and the two zoned write examples call it before writing.
A first cut put it in the `device` fixture ahead of every test on such a
device; that made the zoned examples' read tests fail, since they read what
the write tests before them left behind, so the call moved into the tests
that write.

Verification: reproduced by hand in the trixie CI guest (kernel 6.12.94): a
16 KiB write through `/dev/nvme1n1`, then `zoned mgmt-reset /dev/ng1n1
--slba 0`, after which `blkzone report` shows the zone empty and a 4 KiB
write at its start through the block device fails, and after
`blkzone reset /dev/nvme1n1` the same write succeeds. At the test level,
with stale pointers armed the same way on zones 0 and 1,
`test_fio_engine_zns` for `be=io_uring,mem=posix` on `/dev/nvme1n1` fails
without this change and passes with it; the whole zoned selection of the
suite, 261 cases in random order, then passed against the same guest with
the change. FreeBSD is excluded by the OS check and is only read;
there is no zone write plugging there.

### DIFF
--- a/cijoe/tests/apps/test_fio.py
+++ b/cijoe/tests/apps/test_fio.py
@@ -2,7 +2,7 @@ from pathlib import Path
 
 import pytest
 
-from ..conftest import get_osname, get_shm_id, xnvme_parametrize
+from ..conftest import get_osname, get_shm_id, xnvme_parametrize, zoned_bdev_reset
 from .fio_fancy import fio_fancy
 
 FIO_OUTPUT_FPATH = (
@@ -95,6 +95,10 @@ def test_fio_engine_zns(cijoe, device, be_opts, cli_args):
 
     size = "64M"
     # size = "1G"
+
+    # fio resets the zones it writes through the NVMe admin path, which the
+    # block layer does not see; see zoned_bdev_reset()
+    zoned_bdev_reset(cijoe, device)
 
     err, _ = fio_fancy(
         cijoe,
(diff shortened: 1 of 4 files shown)


yonggilsong (approved):
lgtm

karlowich (approved):
lgtm

PR #782, opened 2026-09-14
ci(build-linux): execute the meson test-suite against the ramdisk

yonggilsong:
Follow-up to #778, requested by @safl in the approving review: a meson test step in the build-linux job, with the two enum cases excluded.

The suite runs against the ramdisk backend, so it needs neither hardware nor privileges. No CI job ran it before, which is why the nil-probe defect repaired in #778 went unseen. The enum cases enumerate the devices of the host and are the only ones needing hardware; they are filtered out by name.

The second commit settles the remaining ask from the same review: the backend array excerpt in the architecture docs still listed the nil stub first; it now matches lib/xnvme_platform_linux.c.

Verified locally as an unprivileged user with the exact step command: 55 cases listed, 55 pass, 0 fail.


karlowich (dismissed):
lgtm

safl (commented):
Reframed as a change request below: the silent-swallow behaviour makes this a blocker rather than optional.

safl (changes requested):
This wants changing before merge: the bash filtering can drop tests silently.
The exit status of `meson test --list` is lost inside the `< <(...)` process
substitution, so a failed or truncated listing leaves `mapfile`'s `tests` array
empty or partial, and the following `meson test` then either runs the whole
suite (enum included) or quietly skips cases while still reporting green. A CI
step that can pass without having run the tests is the blocking part.

Meson's `suite` feature does this cleanly and we are not using it yet. Tag each
test in the `test()` call in `tests/meson.build` with `suite : fs.stem(test_source)`
so `enum.c` lands in an `enum` suite, then the step is one portable line with
nothing to swallow:

```yaml
run: meson test -C builddir --print-errorlogs --no-suite enum
```

No `mapfile`, no process substitution, no `^(xnvme:)?` prefix handling, and
`meson test` sets its own exit status. As a bonus it lets anyone run a single
file's tests with `--suite enum`. The cost is a `tests/meson.build` change
rather than a pure CI patch, which is the right place for it.

Not run locally; from reading the test registration in `tests/meson.build`.

Assisted-by: Claude Code:claude-opus-4-8

yonggilsong:
Adopted as suggested: each test in tests/meson.build now carries the stem of its source as its suite, and the step is the single --no-suite enum line. Verified locally: --no-suite enum runs 55 cases, all pass; --suite enum selects exactly the two enum cases. Amended into the first commit.

safl (approved):
Excellent, thanks!

karlowich (approved):
Great!

PR #783, opened 2026-09-16
docs(transports): say which target script takes which flag

safl:
On top of #739, which it follows up: the branch carries that PR's two commits and rebases to this one commit once #739 merges. Look at the last commit only.

What to look at: the reworded paragraph above the per-script tables (`probe` is not switched via `--nvme-provider`; `--transport-name` is the one flag all three share) and the `--nvme-trsvcid` description, which said TCP port while the flag now accepts `rdma`.

Verification: read against `add_args()` of the three scripts on the #739 head; the docs build was not run.


safl:
Folded into #739 as a cherry-pick instead of a stacked PR.

PR #784, opened 2026-09-18
chore: maintenance pass (README, Runners, and Workflows)

safl:
A maintenance pass on the README and the self-hosted Hetzner runner
provisioning, bundled into one PR to avoid separate CI runs for each.
Commit messages carry the why for each change; the Hetzner fixes were
found and verified live while actually bringing
xnvme-ci-hetzner-bookworm-amd64-02 up from a fresh install.

Assisted-by: Claude Code:claude-sonnet-5

PR #785, opened 2026-09-18
fix(ci): fix Hetzner runner provisioning (stale install image, qemu fork)

safl:
## Summary

Two independent fixes to the self-hosted Hetzner runner provisioning,
both found and reproduced while reinstalling
`xnvme-ci-hetzner-bookworm-amd64-02`.

**Stale install image.** `install.conf.tmpl` named
`Debian-1211-bookworm-amd64-base.tar.gz` as the install image. That file no
longer exists on Hetzner's rescue systems: the current point release is
`Debian-1215-bookworm-amd64-base.tar.zst` (renamed, and now
zstd-compressed instead of gzip). `installimage` failed immediately at
"Loading image file variables" trying to read the missing file's sidecar
metadata. Switches to the `Debian-bookworm-latest-amd64-base.tar.zst`
symlink Hetzner keeps current, instead of a hardcoded version, so this
doesn't go stale again the next time the image is rotated.

**QEMU built from a fork nothing needs anymore.**
`400-qemu-from-source.sh` built `SamsungDS/qemu#for-xnvme` into
`/opt/qemu`. That fork predates `vfio-user-kvssd`: the `ax42-*.toml` cijoe
configs that actually consume `/opt/qemu` pass real NVMe devices through
via `vfio-pci` and never reference `vfu_kvssd`, and the containerized
`verify` jobs (`debian-trixie-iommu_*.toml`, `freebsd-14.toml`) already
build plain upstream `qemu/qemu.git` at `v9.0.0` and get their KV device
from a separately-downloaded `vfu_kvssd` release binary. Points the runner
provisioning task at the same upstream repository and tag those
containerized jobs already use, and fixes the devs tutorial's qemu build
instructions (`docs/tutorial/devs/index.rst`) to match, since they told a
new contributor to check out the same fork.

## Test plan

- [x] Reproduced the image failure on `xnvme-ci-hetzner-bookworm-amd64-02`;
      confirmed via `/root/debug.txt` and `ls /root/images/` on the rescue
      system that the referenced `.tar.gz` is gone and only `.tar.zst`
      images remain
- [x] Re-ran the fixed `install.conf.tmpl` end to end on that host:
      installation completed and the host is back up as Debian 12.15
- [x] Grepped every `cijoe/configs/*.toml` for `/opt/qemu`,
      `qemu.repository`, and `vfu_kvssd` to confirm nothing self-hosted
      still needs the fork's behavior
- [ ] The upstream-QEMU build was still running on -02 as of this update;
      haven't yet run a `benchmark-latency` / `verify-maas` job against it
      post-change


safl:
Bundled into #784 to avoid separate CI runs for each. Closing this one.

PR #786, opened 2026-09-18
ci(runners): build upstream QEMU instead of the SamsungDS fork

safl:
## Summary

`400-qemu-from-source.sh`, part of the self-hosted Hetzner runner
provisioning, built `SamsungDS/qemu#for-xnvme` into `/opt/qemu`. That fork
predates `vfio-user-kvssd`, and checking what actually consumes
`/opt/qemu` today shows nothing needs it anymore:

- The `ax42-*.toml` cijoe configs (the ones these runners actually use)
  pass real NVMe devices through to the guest via `vfio-pci` and don't
  reference `vfu_kvssd` anywhere.
- The containerized `verify` jobs (`debian-trixie-iommu_*.toml`,
  `freebsd-14.toml`) already build plain upstream `qemu/qemu.git` at
  `v9.0.0`, and get their KV device from a separately-downloaded
  `vfu_kvssd` release binary, not from the qemu build at all.

Points the runner provisioning task at the same upstream repository and
tag the containerized jobs already use, instead of the fork, and fixes
the devs tutorial's qemu build instructions (`docs/tutorial/devs/index.rst`)
to match, since they told a new contributor to check out the same fork.

## Test plan

- [x] Grepped every `cijoe/configs/*.toml` for `/opt/qemu`,
      `qemu.repository`, and `vfu_kvssd` to confirm nothing self-hosted
      still needs the fork's behavior
- [x] Re-ran the fixed provisioning task end to end on
      `xnvme-ci-hetzner-bookworm-amd64-02` during its reinstall; upstream
      QEMU v9.0.0 builds and installs to `/opt/qemu` the same as before
- [ ] Haven't yet run a full `benchmark-latency` / `verify-maas` job
      against this runner post-change; will confirm once one lands


safl:
Bundled into #784 to avoid separate CI runs for each. Closing this one.

PR #789, opened 2026-09-21
refactor(xnvmeperf): add helpers in xnvme_util.h and xnvme_cli.h

karlowich:
This moves common operations like pinning threads or opening multiple devices into helpers instead of each tool having to reimplement it.
This is in preparation for a small qublk refactor #790 .
Pinning threads and calculating uPCIe heap size moves to a new internal header `xnvme_util.h`.
Opening/closing a list of devices moves to `xnvme_cli.h` 

NOTE: this changes xnvmeperf behaviour to not parallelise `xnvme_dev_close()`. I have measured that xnvmeperf closing 16 devices takes only 1.7s which I think is acceptable. Calling `xnvme_dev_close()` in parallel is not safe with `--be upcie` and thus the change fixes a bug.


lib/xnvme_cli.c
```diff
@@ -2247,6 +2247,47 @@ xnvme_cli_run(struct xnvme_cli *cli, int argc, char **argv, int opts)
...
+		xnvme_cli_perr("calloc()", err);
+		return err;
+	}
+
+	for (int i = 0; i < count; i++) {
+		opened[i] = xnvme_dev_open(uris[i], opts);
+		if (!opened[i]) {
+			char msg[XNVME_IDENT_URI_LEN + 32];
+
+			err = -errno;
```
safl: `err = -errno` here can be 0 if `xnvme_dev_open()` fails through a path that doesn't set errno. `xnvme_cli_devs_open()` then returns 0 without `*devs` ever being written, and every caller (`homi`'s `sub_start()`, `xnvmeperf_open_devs()`) dereferences an uninitialised pointer. Fall back to a fixed negative errno (e.g. `-EIO`) when `errno` reads 0, or set `*devs = NULL` on entry so a caller can't be handed garbage.

include/xnvme_util.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Pin the calling thread to 'cpu'
+ *
+ * @return 0 on success, an errno value otherwise
+ */
+static inline int
+xnvme_util_pin_to_cpu(int cpu)
+{
+#ifdef XNVME_PTHREAD_SETAFFINITY_NP_ENABLED
+	cpu_set_t cpuset;
```
safl, line 44: `cpu_set_t`/`CPU_ZERO`/`CPU_SET` are only visible when `_GNU_SOURCE` is defined before the first libc header the including translation unit pulls in. Today's two consumers (xnvmeperf.c, qublk/io.c) happen to define it, but nothing here enforces that on the next includer. Worth a line in the docstring, or move the body into lib/ where the project's own build flags apply.

include/libxnvme_cli.h
```diff
@@ -562,3 +562,27 @@ xnvme_cli_run(struct xnvme_cli *cli, int argc, char **argv, int opts);
...
+ * @param uris The device URIs to open
+ * @param count The number of URIs
+ * @param opts The device-options to open them with
+ * @param devs Set to an array of 'count' devices, release it with ::xnvme_cli_devs_close
+ *
+ * @return On success, 0 is returned. On error, negative errno is returned and no device
+ * is left open.
+ */
+int
+xnvme_cli_devs_open(const char **uris, int count, struct xnvme_opts *opts,
```
safl: We haven't really used a plural "devs" in the codebase before, but we have used "multi" elsewhere (tests/xnvme_tests_multi_ctrlr), so how about aligning this closer to the existing vocab, e.g. `xnvme_cli_dev_open_multi()`/`xnvme_cli_dev_close_multi()`? Every other open/close pair in the tree keeps the noun singular (`xnvme_dev_open()`, `xnvme_be_linux_dev_open()`, ...).

lib/libxnvme.map
```diff
@@ -70,6 +70,8 @@
 		xnvme_cli_perr;
 		xnvme_cli_run;
 		xnvme_cli_to_opts;
+		xnvme_cli_devs_open;
```
safl: The rest of this block follows header order; these two land after `xnvme_cli_to_opts` in the order they were added rather than where the header declares them.

tools/xnvmeperf/xnvmeperf.c
```diff
@@ -332,7 +310,7 @@ thread_fn(void *arg)
 	uint64_t runtime_ns = (uint64_t)args->time * 1000000000ULL;
 	int err;
 
-	err = pin_to_cpu(thread->cpu);
+	err = xnvme_util_pin_to_cpu(thread->cpu);
 	if (err) {
 		fprintf(stderr, "Warning: failed to pin thread to CPU %" PRIu16 "\n", thread->cpu);
```
safl, line 317: On platforms without `pthread_setaffinity_np`, `xnvme_util_pin_to_cpu()` now returns `ENOSYS`, and this still prints "failed to pin thread", which reads as a failure rather than an unsupported platform.

lib/xnvme_cli.c
```diff
@@ -2247,6 +2247,47 @@ xnvme_cli_run(struct xnvme_cli *cli, int argc, char **argv, int opts)
...
+
+	*devs = opened;
+
+	return 0;
+}
+
+void
+xnvme_cli_devs_close(struct xnvme_dev **devs, int count)
+{
+	for (int i = 0; devs && i < count; i++) {
```
safl: Minor: the NULL guard inside the loop condition (`devs && i < count`) would read clearer as an early `if (!devs) return;`.

safl (commented):
In good shape overall, no correctness bug on the documented happy path, but xnvme_cli_devs_open() can return 0 without ever setting *devs when errno happens to read 0 on a failed open, which is worth fixing before merging since this is now a public, exported function rather than a tool-private helper.

The inline comments are small and simple: the errno fallback, xnvme_util_pin_to_cpu()'s undeclared _GNU_SOURCE requirement, a naming mismatch against the rest of the tree's dev_open()/dev_close() convention, and the libxnvme.map entries appended out of header order.

Could be addressed in a follow-up rather than blocking this: docs/api/c/cli/xnvme_cli.md is generated by `make docs` and nothing in CI runs that target, so it needs a manual regen and commit alongside the new public functions; tests/xnvme_cli.c has no coverage yet for the new multi-device open/close path (opening N URIs, or failing partway through and checking the already-opened ones get closed); and the "failed to pin thread" warning in thread_fn() now fires on ENOSYS too, which reads as a failure rather than an unsupported platform.

Read the diff and the surrounding files on main; nothing built or run.

Assisted-by: Claude Code:claude-fable-5-1

lib/libxnvme.map
karlowich: This is wrong. The order in `libxnvme.map` matches the order in the header.

karlowich:
@safl 
>> docs/api/c/cli/xnvme_cli.md is generated by make docs and nothing in CI runs that target

This is not true, the `docgen` job runs `make docs` on every PR. Additionally, the markdown files aren't tracked so running `make docs` locally does nothing.



I will address the rest of the feedback

lib/xnvme_cli.c
karlowich: Will fix

tools/xnvmeperf/xnvmeperf.c
karlowich: Will fix

include/libxnvme_cli.h
karlowich: Good point, I agree that "devs" is not clear enough.
I don't really see the precedent for "multi". The only test I can find is `xnvme_tests_enum multi` which opens each device multiple times, not exactly the same.
I'm wondering if perhaps `xnvme_cli_dev_open_list()` is clearer.

include/xnvme_util.h
karlowich: I will introduce a xnvme_util.c

lib/xnvme_cli.c
karlowich: will fix

include/libxnvme_cli.h
safl: Yes, the tests multi is closest thing. The thing about "_list" is that in my mind then I expect to pass linked-list of sorts, not an array of pointers. Lists indicate a **type** of sorts, where multi is just stating the "ability" not the representation.

karlowich: Okay, we go with multi

safl (approved):
lgtm

naddinadja (approved):
looks great! :)

